PR #12, opened 2015-06-03
Added global "before" function which is ran before every route call

kacpersaw:


src/view.js
```diff
@@ -47,6 +47,21 @@ module.exports = function (Vue) {
         return this.invalidate()
       }
 
+      // for every route run global before function
```
yyx990803: I think the global before should not be triggered here - this function is called for very nested view on a route change. The global before should be called once only for each route change, so it should actually be in the `_match` method of the router class.

yyx990803:
Hi @kacpersaw , thanks a lot for the PR, but as I was reviewing the global hook design I find this implementation leaving a lot to be desired. Unfortunately I can't merge this as is, and I do need to move forward, so I have implemented the functionality myself. Again, thanks for taking a stab at this, really appreciated.

PR #15, opened 2015-06-10
Follow v-link with left button only to allow opening in a new tab by middle clicking

jobedom:
Links using `v-link` don't react properly to middle clicking (i.e.: open in a new tab). Links should only navigate to destination if left-clicked.


yyx990803:
Good catch, thanks!

PR #98, opened 2015-08-27
Use environment variable for sauce username setting instead of 'vuejs'

greennyreborn:
I was just trying to run test via saucelab.com and ran into the problem that username and access_key doesn't match.
Turns out in the `build/nightwatch.sauce.json`, `username` is set to `vuejs` manually.
So I change it to `${SAUCE_USERNAME}` just like the variable in the circle.yml.
And everything works!


greennyreborn:
Please ignore this request, I put some other irrelevant commit by mistake.

PR #145, opened 2015-09-17
Typo serivce - service

juukie:
Shouldn't $loadingRouteData be $loading (as mentioned a few lines below)?


juukie:
According to [this page](http://vuejs.github.io/vue-router/en/pipeline/data.html) the $loadingRouteData is correct so no worries about that.

Issue #140, opened 2015-09-16
Add delegate functionality to v-link


lepture:
An example for delegating `v-link-delegate`:

```html
<div v-html="aVar" v-link-delegate></div>
```

`aVar` is a HTML variable that contains inner links, e.g.

```html
<!-- aVar -->
<a href="/foo">go to foo</a>
```

We can't add `v-link` directive to this link since it is in a variable. My solution is adding a `v-link-delegate` directive to listen for link redirection.

page.js can handle the links very well since it captures every link redirection.

I am not sure whether it should be added to vue-router.

yyx990803:
Sounds like a decent solution. I'd take a pull request.

lepture:
@yyx990803 It seems we can merge `v-link` and `v-link-delegate` together.

```html
<div v-link></div>
```

`v-link` on `div` is a delegate obviously.

yyx990803:
@lepture hmm, good idea!

lepture:
I'm going to implement it.

yyx990803:
:+1:

lepture:
@yyx990803 Not sure how to implement the test case. The `click` function seems not working properly.

yyx990803:
You can manually dispatch a fake `click` event:

``` js
function trigger (el, event) {
  var e = document.createEvent('HTMLEvents')
  e.initEvent(event, true, true)
  el.dispatchEvent(e)
}
```

PR #147, opened 2015-09-20
Link delegate

lepture:
https://github.com/vuejs/vue-router/issues/140

`v-link` has delegate feature now.

I can't implement the test case correctly.

### DIFF
--- a/src/directives/link.js
+++ b/src/directives/link.js
@@ -26,10 +26,25 @@ export default function (Vue) {
         // don't redirect when preventDefault called
         if (e.defaultPrevented) return
 
-        if (e.button === 0) {
+        if (e.button !== 0) return
+
+        if (this.el.tagName === 'A') {
+          // v-link on <a v-link="'path'">
           e.preventDefault()
           if (this.destination != null) {
             router.go(this.destination)
+          }
+        } else {
+          // v-link delegate on <div v-link>
+          var el = e.target
+          while (el && el.tagName !== 'A' && el !== this.el) {
+            el = el.parentNode
+          }
+          if (!el || el.tagName !== 'A' || !el.href) return
+
+          if (sameOrigin(el)) {
+            e.preventDefault()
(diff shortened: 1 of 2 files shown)


BrainBuzzer:
gr8

test/unit/specs/core.js
```diff
@@ -292,6 +292,45 @@ describe('Core', function () {
...
+      template: '<router-view></router-view>'
+    })
+    router.start(App, el)
+    el = router.app.$el
+    router.go('/foo')
+    nextTick(function () {
+      var wrap = el.querySelector('#wrap')
+      var e = document.createEvent('HTMLEvents')
+      // target is read-only
+      e.target = wrap.querySelector('a')
```
lepture: can't figure out how to implement a delegate click

yyx990803: just dispatch it on `wrap.firstChild`?

lepture: @yyx990803 No, it can't be done this way. `v-link` is listening click event on wrap, it needs information of target.

src/directives/link.js
```diff
@@ -26,11 +26,26 @@ export default function (Vue) {
...
+
+        if (this.el.tagName === 'A') {
+          // v-link on <a v-link="'path'">
           e.preventDefault()
           if (this.destination != null) {
             router.go(this.destination)
           }
+        } else {
+          // v-link delegate on <div v-link>
+          var el = e.target
```
lepture: @yyx990803 can't set target on unit test.

test/unit/specs/core.js
yyx990803: `event.target` will always point to the element that dispatched it. The event will then propagate up and be captured by the listener on the parent.

lepture: @yyx990803 actually, it won't. I've tried it already.

Issue #103, opened 2015-08-30
Move babel-runtime to devDependencies?


rpkilby:
Hi. Can the babel-runtime dependency be moved to devDependencies? A quick repository search seems to indicate that babel-runtime isn't necessary, and the runtime causes issues when loaded in a System.js environment.

yyx990803:
It is indeed a runtime dependency, because the files distributed on npm is the babel-compiled version, which depends on `babel-runtime`. If you are using System.js, try importing `src/index.js` instead (the default main in package.json is `lib/index.js`)

rpkilby:
Ah - that makes sense.

I can't really give you a great explanation since the material is a bit beyond my understanding, but a bit of looking shows there's a [package override](https://github.com/jspm/registry/blob/master/package-overrides/npm/babel-runtime%405.0.0.json) in the jspm registry that specifically removes the dependencies from babel-runtime. I don't understand why, but my guess is that is has something to do with how [es6-module-loader](https://github.com/ModuleLoader/es6-module-loader/wiki/Production-Workflows) uses the runtime. Either way this is causing issues where, surprise, core-js isn't found.

In order to import `src/index.js`, should `"files": ["src"]` be added to the package.json?

PR #163, opened 2015-10-06
Distribute source with package files

rpkilby:
From the comments on #103 
> In order to import src/index.js, should "files": ["src"] be added to the package.json?


yyx990803:
This should be done by removing `src` from `.npmignore`.

PR #184, opened 2015-10-15
fix wrong characters in zh-cn doc.

Gerhut:
Here is the Chinese translation of [hook](https://zh.wiktionary.org/wiki/hook)


yyx990803:
多谢！

PR #232, opened 2015-11-12
Update Babel to 6

lepture:
The lib files are conflict with current `vue-loader`.


lepture:
I have tried a lot to fix the tests, but failed. https://circleci.com/gh/vuejs/vue-router

yyx990803:
Thanks for the PR - looks like it was a Babel 6 bug. It seems to be fixed now but there is now a different error.

Issue #260, opened 2015-12-01
Event does not propagate through router on dispatch


perkola:
I have a Vue instance that listens for an event and a router bound to it like so:
```
<div id="app">
    <router-view></router-view>
</div>
```
and on a route i have a child component that dispatches an event via `this.$dispatch('message', 'Test message');`. Without the router in between the event propagates properly from the child component to the parent Vue instance but with the router the message is never received at the parent. Why?

azamat-sharapov:
Reproduction?

perkola:
[http://jsfiddle.net/0brxezmj/15/](http://jsfiddle.net/0brxezmj/15/) is a reproduction of the problem. Note that the component directly under the parent Vue can dispatch events while the same component under the route-view can't.

azamat-sharapov:
It's working for me: http://jsfiddle.net/0brxezmj/16/

_Make sure you provide clean and **working** fiddle next time, so one does not have to spend time to clean your code to actually make it run_.

rpkilby:
Just for clarification - the underlying issue was that a Vue instance was passed to the router instead of a component constructor.

amirrustam:
@perkola looks like this is resolved, can you please close this out.

mknxdev:
> It's working for me: http://jsfiddle.net/0brxezmj/16/
> 
> _Make sure you provide clean and **working** fiddle next time, so one does not have to spend time to clean your code to actually make it run_.

I know this is an old post but I wanted to answer to @azamat-sharapov that told @perkola to "provide a clean and working fiddle".
I'm always surprised to see developers which can't fully read and understand the initial question, or they don't want to, or they're bad faith. I don't know...
The fact is, @perkola asked for help **because** its code was not working, so how can he provide you a **working** fiddle if he has a problem with that? Not everyone is an hardcore-developer like you, man. Not everyone is able to provide clean, optimized and working code at the first time. But you don't seem to understand that.

PR #274, opened 2015-12-08
Throw exception when router started w/ Vue instance

rpkilby:
It seems like passing a Vue instance to the router instead of a component constructor is a somewhat common and easy mistake to make. See #260 and vuejs/vue#1981.


yyx990803:
Thanks!

Issue #112, opened 2015-09-02
Accessing route VM within router hooks for updating document.title


holic:
I'd like to change the page title on route changes for better history management. I thought to do this in the `afterEach` router hook, using a data property on the VM (computed or otherwise). However, I can't find a good way to access the `<router-view>` VM from within this hook.

Any alternatives you would suggest?

holic:
I thought I'd try something like:

```html
<router-view v-ref="view"></router-view>
```

And then watch for a property on the VM reference:

```js
// root Vue app
new Vue({
  watch: {
    '$.view.title': function (title) {
      document.title = title
    }
  }
})
```

But this doesn't work. I also tried a computed property, with no luck:

```js
// root Vue app
new Vue({
  computed: {
    title: function () {
      return this.$.view && this.$.view.title
    }
  },
  watch: {
    'title': function (title) {
      document.title = title
    }
  }
})
```

Ultimately, I found that if I watch the `$route` property from the root VM, then add a `title` watcher to the router VM via `this.$.view.$watch`, it works. But I have to manually manage watchers (removing the previous one before adding a new one) and I don't get the first title on page initialization (`$route` seems to be set before `this.$.view` is available, so `immediate: true` fires with an empty `this.$.view`).

JosephSilber:
I think document title (and history title) support should really be built in to the router.

holic:
Agreed @JosephSilber. Ideally it would allow for a computed property on the router view so that the router view's component can update the title dynamically (e.g. unread count on an inbox).

holic:
I think I found a better solution than the one I mentioned above, via a mixin for my route components:

```js
module.exports = {
	watch: {
		title: {
			immediate: true,
			handler: function (title) {
				if (title) {
					document.title = title
				}
			}
		}
	}
}
```

But I do think that this should be built in to the router in some way.

yyx990803:
What's wrong with setting `document.title` in `activate` or `data` hook?

JosephSilber:
That wouldn't affect the history entry.

holic:
@yyx990803 I didn't want to have to include it with every single route component (mixins previously did not extend to the route object) and I want to be able to get the title value from the current `<router-view>` VM.

Atinux:
Hi,

[...]

PR #279, opened 2015-12-10
Emit route-data-loaded event for <router-view>

lepture:
With this patch, we can listen routeDataLoaded event on `<router-view>`.

```html
<router-view @route-data-loaded="doSomething"></router-view>
```

In this doSomething:

1. we can send collect analytics data
2. we can change document.title
3. and many other things

https://github.com/vuejs/vue-router/issues/112

### DIFF
--- a/src/pipeline.js
+++ b/src/pipeline.js
@@ -289,10 +289,12 @@ function loadData (component, transition, hook, cb, cleanup) {
     }
     if (!promises.length) {
       component.$loadingRouteData = false
+      component.$emit('route-data-loaded', component)
       cb && cb()
     } else {
       promises[0].constructor.all(promises).then(_ => {
         component.$loadingRouteData = false
+        component.$emit('route-data-loaded', component)
         cb && cb()
       }, onError)
     }


hartmut-co-uk:
Hi, when exactly is the @route-data-loaded event supposed to be dispatched? 
I've tried with v0.7.13 without success. Seems it's never getting triggered.

rvrfox:
英文文档里有这个事件，为啥我实际使用不行，然后搜代码也搜不到。

lepture:
@hartmut-ltd @nigelvon have you called `next` in `route.data`?

hartmut-co-uk:
Hi @lepture ,

thanks for the hint, I noticed the event only is emitted when `route.data` is actually used...

Quick example, `sayHello` is only triggered on transition to `Foo` route.
http://jsfiddle.net/moominpapa/3wwadz4t/3/

I guess for the event named 'route-data-loaded' this behaviour could be OK - it may need a note in the documentation though.
Is there an event triggered on each route change no matter if `route.data` is used or not? I could certainly use one...

hartmut-co-uk:
Note: no need to call `next`, it's working with an empty function declaration even, but not without defining `route.data`.
http://jsfiddle.net/moominpapa/3wwadz4t/4/

PR #287, opened 2015-12-12
Switch true and false

juukie:
I think these should be switched according to the example further below. 
~~~
<div v-if="$loadingRouteData">Loading ...</div>
~~~

and at the bottom: "$loadingRouteData will be set to false when all Promises have been resolved."


yyx990803:
oops, good catch :+1:

PR #330, opened 2016-01-15
Fixed example

ChrisCates:
This was a gotcha when I first looked at it. Just fixing it up so it works properly. :+1:


yyx990803:
It's not needed in the basic example because `VueRouter` auto installs when included via `<script>`. As the doc states, `Vue.use(VueRouter)` is only necessary when in a module environment.

Issue #378, opened 2016-02-21
URL query parameters getting encoded twice


decademoon:
If I perform this:

```
this.$router.go({ name: 'a', query: { b: '/c' } })
```

then the URL in the address bar becomes `/a?b=%252Fc` instead of `/a?b=%2Fc`.

`this.$route.query.b` will be `'%2Fc'` instead of `'/c'`.

---

It seems that each query parameter is being `encodeURIComponent`-ed, combined into the full path, then the full path is `encodeURI`-ed. Shouldn't the URL be `encodeURI`-ed before the query string is appended?

GuillaumeLeclerc:
Do you have any idea when it will be integrated on the npm registry ?

PR #391, opened 2016-02-26
修复 url 参数二次 encode 的问题

byrdkm17:
修复 #378 中提到的问题，请求合并。


davidbgk:
Is that possible to have a release including that fix?

PR #392, opened 2016-02-26
update docs

kazupon:
reference: fdcb73bc6d7c47c733768676de02cb18e7526738


yyx990803:
It seems the current docs is correct.

PR #393, opened 2016-03-03
Change suppressTransitionError to handleTransitionError

lepture:
`handleTransitionError` is a function, in this way, we can send error logs to sentry or any other services for production application.

BTW. I've changed the default method for reporting errors. The original one which will throw an Error during transition errors will cause the router redirecting infinitely.


yyx990803:
Thanks, will look at this for 0.8.0

lepture:
@yyx990803 I have another feature request. A pull request is coming soon.

lepture:
ping

PR #407, opened 2016-03-11
Add a handleClick method for rewriting purpose.

lepture:
The `handleClick` accepts an extra `el` parameter, which is used for rewriting the method.

Now, I rewrite the whole `onClick` for `v-link` in my case.

1. A `<a>` has title information
2. Click on `<a>` link, get title information from `<a>` element
3. Set `document.title` with this `<a>` element title information

It is not limited with `document.title`. For a hybrid Application, it is quite useful
to send the title information of the next page to the App container.

### DIFF
--- a/src/directives/link.js
+++ b/src/directives/link.js
@@ -82,7 +82,7 @@ export default function (Vue) {
       if (target) {
         // v-link with expression, just go
         e.preventDefault()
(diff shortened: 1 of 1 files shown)


lepture:
weird. how could it fail?

PR #415, opened 2016-03-20
Minor tweaks for Japanese docs

tejitak:
Fix several Japanese translations.

@kazupon can you double-check please?


kazupon:
@tejitak  LGTM :+1:

PR #417, opened 2016-03-22
Fixed the query encoding order issue.

dragantl:
The original code would encode full path differently based on whether the path was supplied as `name` or as `path`. Fixed so that all branches conform to the same encoding order, which is encoding of URI components and then or URI itself. Also, route-recognizer would decode in inverse order (first decode URI components and then URI). That was fixed as well (encode and decode orders should be inverse).

This should address issue https://github.com/vuejs/vue-router/issues/378 and fix an issue introduced by https://github.com/vuejs/vue-router/pull/391 PR.

### DIFF
--- a/docs/en/pipeline/data.md
+++ b/docs/en/pipeline/data.md
@@ -108,7 +108,7 @@ Using `$loadingRouteData` in templates:
   <div v-if="$loadingRouteData">Loading ...</div>
   <div v-if="!$loadingRouteData">
     <user-profile user="{{user}}"></user-profile>
-    <user-post v-repeat="post in posts"></user-post>
+    <user-post v-for="post in posts"></user-post>
   </div>
 </div>
 ```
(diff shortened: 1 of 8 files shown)


yyx990803:
Thanks for the PR! Is it possible to add a test case for this fix?

dragantl:
Ok, I'll add test cases for the three code paths. I'll either do it tomorrow or Monday as I'm leaving on vacation.

dragantl:
@yyx990803 I've added the tests to check correctness or querystring encoding and decoding via object path, object named path, and string path. They are passing. Should this be sufficient?

yyx990803:
Tests look good - although it seems there's some merge conflicts. Mind rebasing against the latest dev?

yyx990803:
I think it probably has to do with this commit: https://github.com/vuejs/vue-router/commit/cf00b488771ce74de4b13479c464474e59bdd28c

dragantl:
@yyx990803 ok, changes are rebased against that commit, test pass.

yyx990803:
Hmm, seems the commit history has become a bit weird - I manually cherry-picked your commits: https://github.com/vuejs/vue-router/commits/dev I hope you are ok with that!

Thanks a lot for the great work!

dragantl:
Yeah, not a problem, I thought so too.

PR #418, opened 2016-03-23
Avoid an error when a root '/' for VueRouter is specified (fix #379)

### LINKED ISSUE #379: Setting root to '/' throws an "Invalid regular expression" error
I want to set the root to `/` so that the activeClassName is set properly for nested routes within the root(because currently if you have a route definition for `/`, and then other routes starting with `/`, ex: `/cats` and you navigate to `/cats`, both anchors to `/` and `/cats` will have the active class), but I get this exception thrown: 
```
Invalid regular expression: /^\/: \ at end of pattern
```
being thrown by this line: https://github.com/vuejs/vue-router/blob/dev/src/history/html5.js#L14

tejitak:
This is reported in https://github.com/vuejs/vue-router/issues/379
The reproduce step is just specifying '/' as root like below.

```js
  history: true,
  saveScrollPosition: true,
  root: '/'
})
```
You can see `"Invalid regular expression: /^\/: \ at end of pattern"`.

The root param '/' is not needed and we can just remove it to work as expected, but I think it's better to avoid a runtime error when it's explicitly specified.

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -4,7 +4,7 @@ const hashRE = /#.*$/
 export default class HTML5History {
 
   constructor ({ root, onChange }) {
-    if (root) {
+    if (root && root !== '/') {
       // make sure there's the starting slash
       if (root.charAt(0) !== '/') {
         root = '/' + root


blake-newman:
Code is good, however can you add some tests to cover this scenario.

PR #425, opened 2016-03-26
Update to go.md

billmurrin:
Corrected append option to reflect `append: false` when not using append.


yyx990803:
Thanks, note the first line is `without`, so I believe the original is correct.

billmurrin:
Oh yeah, sorry to waste our time. Thanks again.

Issue #296, opened 2015-12-18
[Feature] A reload method like location.reload()


fnlctrl:
I have a router-view that needs to be reloaded after sending some POST requests and receiving whether the operation was successful, but
```
this.$route.router.go(this.$route.path)
```
doesn't trigger a reload even if the component's `route.canReuse` is set to false.

Now I can only do 
```
created.apply(this) // `created` is the life-cycle hook function for the component
```
in the ajax callback to force the router-view to update itself, which is cumbersome.

Perhaps something like $route.router.reload() would be great, as it would be the counterpart of location.reload().

fnlctrl:
Lacking a reload method also prevents me from putting request promises in `component.route.data()` because it's only triggered when route path changes...

roblav96:
Having a reload route method would be nice.

ccqgithub:
i think so ~~ Having a reload route method would be nice.

roblav96:
having a for re-render on a specific v-for item would be even better

amirrustam:
@fnlctrl can you please elaborate in detail on your particular use case for this feature? I'm finding it difficult to reason why the entire router-view needs to b reloaded based on a status of a POST request.

yyx990803:
Here's what I think: after the POST request, what you really want to do is not reloading the component, but rather **re-fetch the data**. However you have placed the data-fetching logic in your component's lifecycle hooks, but that doesn't have to be the case. Consider:

``` js
export default {
  created () {
    this.fetchData()
  },
  methods: {
    fetchData () {
       // fetch data
    }
  }
}
```

After the POST request is resolved, you simply call `this.fetchData()` to re-fetch. Reloading the entire component is doing more than necessary, since it leads to a full re-render of the DOM as well.

fnlctrl:
@yyx990803  No that's not what I meant..
I just want to reload the route, and component is reused (route.canReuse is true), and nothing is changed except for it's route.data() called again. (I really want to use route.data because it's so elegant).
Consider this.
```javascript
route: {
    data: ({to: {params: {someParam}}}) => ({
        foo: $.get('/path/to/resource', {someParam}),
    })
}
```
Now, without a reload method, I will have to write`$.get('/path/to/resource', {someParam})` again in fetchData..

fnlctrl:
It's like transitioning from `/a/1` to `/a/2`, the component is reused, only the route.data() is fired again.
I just want the same thing to happen when transitioning from `/a/1` to `/a/1`.

[...]

PR #442, opened 2016-04-01
Call 'data' hook by router.go() force option and router.reload()

tejitak:
Having router reload function (just call `data` hook for same component) would be a nice feature from several reported issues #296 etc.

- Add a `force: true` option for `router.go()`
The `data` hook will be called again even when a current same URL is specified.
Example:
  - Call `this.$router.go({ path: '/a', force: true })` from components
  - Define v-link like `<a v-link="{ path: '/inbox/message/123', force: true }">inbox</a>`  will trigger `data` hook by every click

- Add a `router.reload()`
  - Reload with current path and call data hook again


lepture:
You need to rebase this branch

Makio64:
Will be great if this branch is rebase and this feature integrated.

Makio64:
I try this by patching my router manullay and actually the pipeline is activated and the transitionOut is playing but then there is something blocking in the canDeactivate and the pipeline is stop.. :(

Issue #390, opened 2016-02-26
Feature: use v-on:click with v-link 


weotch:
This doesn't appear to work:

```html
<a v-link='{name: "home"}' v-on:click='onClick'>Home</a>
```

I think it's [this `e.preventDefault()`](https://github.com/vuejs/vue-router/blob/dev/src/directives/link.js#L84).  I get why that's there, but can you think of some way we can also handle these clicks from the Component?  Maybe a callback could be specified in v-link, like:

```html
<a v-link='{name: "home"}, onClick'>Home</a>
<a v-link='{name: "home", click: onClick}'>Home</a>
```

weotch:
Another option might be to give `v-on` a higher priority.  I'm currently looking into whether I can adjust the priority of an internal directive...

weotch:
FYI, this approach to reducing the `v-link` priority is working for me so far:

```js
onPriority = Vue.directive('on').priority
Vue.directive('link').priority = onPriority - 1
Vue.directive('link-active').priority = onPriority - 2
```

Glad you included a way to access directives!

perkola:
An idea is that the object passed to `v-link` could have some kind of callback property where you could specify a function on the component to execute.

repsac-by:
I think it is better to use
[Transition Hooks](http://vuejs.github.io/vue-router/en/pipeline/hooks.html)
```javascript
route: {
  canDeactivate() {
    transition.abort()
  }
}
```

rpkilby:
> I think it is better to use Transition Hooks

That doesn't seem like the correct responsibility. The transition hook is there to determine whether or not you can deactivate the current route. I would not want to shove my event handling code inside that. eg,

```html
<a v-link='{name: "home"}'>Cancel</a>
<a v-link='{name: "home"}' v-on:click='saveTheThing'>Save</a>
```

In both cases I'm navigating back to the main page, but canceling and saving are two distinct actions.

perkola:
I agree with @rpkilby. A use case is a hamburger menu component with `v-link` to some route. When a link is clicked the menu should collapse (preferably with something like `v-on:click.prevent='closeMenu()'`) before transitioning into the new route.

This _can_ be achieved with transition hooks that collapses the menu on each `activate` but that results in code duplication since every route has to invoke the `closeMenu()` function on the hamburger menu component.

Gomah:
```In HTML5 history mode, v-link will intercept the click event so that the browser doesn't try to reload the page.```

[...]

PR #447, opened 2016-04-04
lower v-link priority #390

paulpflug:
use-case from @rpkilby:
```html
<a v-link='{name: "home"}' v-on:click='saveTheThing'>Save</a>
```

alternative solution to #390 would be to document this:
```html
<a v-link='{name: "home"}' v-on:click.capture='saveTheThing'>Save</a>
```
(but I can't think of a use case where `v-link` should be evaluated before `v-on`)

### DIFF
--- a/src/directives/link.js
+++ b/src/directives/link.js
@@ -15,14 +15,14 @@ export default function (Vue) {
   } = Vue.util
 
   Vue.directive('link-active', {
-    priority: 1001,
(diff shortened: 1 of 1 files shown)


yyx990803:
Oops, I was working on the very same thing just now and just saw this... it's fixed in df223ac, but thanks for the PR!

Issue #377, opened 2016-02-19
v-link-active only works on direct parent


BenRomberg:
I'm trying to use the new v-link-active attribute and it only seems to work on the direct parent. I'd need to use it nested in a `<li v-link-active><ul><li><a v-link="...">` scenario.

The documentation hints that this should work: "v-link will locate the closest parent element that has v-link-active and apply the active classes on that element instead."

Also, I have a use-case where multiple `v-link-active` would prove useful, e.g. `<li v-link-active><ul><li v-link-active><a v-link="...">`. Would be awesome if that would work as well.

repsac-by:
I have something similar. In my case, `v-link-active` class adds only  on static links and does not work on links generated `v-for`

svedova:
Same here. 

It would be useful to keep a class on parent's parents. When you have nested routes the elements which link to parent route should have a supplementary class (v-link-active-trail) for instance. Would be useful...

cnicodeme:
More over, it would be interesting, in certain case, to work on multiple `v-link-active` (for example for nested sub-menus, when we need to select the current active element + show the parents, and parents element.

azamat-sharapov:
It would probably have performance issues then, because it needs to go all the way up to find `v-link-active`. No?

cnicodeme:
In my opinion I would say it's quite fast since it knows where to go (any dom element has only one parent up to `body`). But it's my opinion, not the absolute truth ;)

I made a quick & easy fix, simply put : `v-bind:class="{ 'active': hasRoute('/my/url') }"` in my `<template />` and the `hasRoute` simply do an `indexOf` between it's argument and the current url (from `$router`)

Works like a charm. Here's the code : 
```javascript
hasRoute: function (partial) {
    return (this.$route.path.indexOf(partial) > -1)
}
```

This let me place the active class anywhere I want.

yyx990803:
Cannot reproduce this, can you provide a jsfiddle?

yyx990803:
Hmm let me guess - are you having something like this?

``` html
<li v-link-active>
  <ul>
    <li v-for="xxx" v-link="xxx">xxx</li>
  </ul>
</li>
```

tejitak:
@yyx990803 I guess so. Here is a working example.
https://jsfiddle.net/tejitak/o44z47ag/5/

Currently each link directive listens `route change` event and update the associated v-link-active element based on matching with the current route and the link value.
[...]

Issue #383, opened 2016-02-23
`v-link-active` does not work when `v-link` is on a component


simplesmiler:
Repro: http://jsfiddle.net/simplesmiler/g61g06gr/

I found this while playing around with the [idea of BEM directive](https://jsfiddle.net/simplesmiler/bgrh5chc/). I needed a way to look up the local DOM tree for the directive that defined a block, and I decided to follow the example of `v-link`/`v-link-active` of setting a custom flag on the element. But apparently this does not work when `v-link` is put on the component, because component node is compiled before the rest of the template.

Is there any other way for two vertically separated directives to talk?

simplesmiler:
I guess I could try to signal the intention to the parent vm (assuming one component equals one block), and then execute things in correct order in the `compiled` hook.

yyx990803:
Fixed in 3d12e95

PR #448, opened 2016-04-05
Support same v-link-active from multiple v-link directives

tejitak:
One possible solution on issue #377 

Here is the summary of this change.
- Listen `route change` event in `v-link-active` directive and check if one of `v-link` children is currently matched
- `v-link` directive without `v-link-active` just check if its link is matched with current path
- Handling nested `v-link-active` parents

### DIFF
--- a/src/directives/link.js
+++ b/src/directives/link.js
@@ -18,8 +18,40 @@ export default function (Vue) {
 
   Vue.directive('link-active', {
     priority: onPriority - 1,
+
     bind () {
+      const vm = this.vm
+      /* istanbul ignore if */
+      if (!vm.$route) {
+        warn('v-link-active can only be used inside a router-enabled app.')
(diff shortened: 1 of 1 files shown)


yyx990803:
Hmm, thanks for the update, but I am actually in the middle of fixing https://github.com/vuejs/vue-router/issues/383 and may have some conflicts with this PR. So just want to give you a heads up that this may not get merged (probably submit as a new PR later).

tejitak:
Okay. I'll add tests at that time if my change looks good.

yyx990803:
Apologies - as I expected, when trying to solve https://github.com/vuejs/vue-router/issues/383 I caused conflict with your PR... but the multi `v-link-active` is now also supported in 0.7.12. Thanks a lot for the effort though!

PR #500, opened 2016-05-17
Prefer using description-style components

simplesmiler:


docs/en/basic.md
```diff
@@ -24,19 +24,19 @@ Creating a Single-page Application with Vue.js + vue-router is dead simple. With
...
     template: '<p>This is bar!</p>'
-})
+}
 
 // The router needs a root component to render.
 // For demo purposes, we will just use an empty one
 // because we are using the HTML as the app template.
 // !! Note that the App is not a Vue instance.
-var App = Vue.extend({})
+var App = {}
```
blake-newman: This may need to retain the same (description explains)

simplesmiler: As stated in http://router.vuejs.org/en/api/start.html, `router.start` can accept `component options object`.
And it seems to be working just fine: http://jsfiddle.net/simplesmiler/xyu276sa/565/

blake-newman: You are right 👍

PR #530, opened 2016-06-16
ES6 extends usage

Eden-Sun:
Would this way usage be more clear or might suffer any problems I don't know ?


yyx990803:
Thanks for the PR, but I don't think this is necessary... it doesn't provide any practical benefits except for using a class.

PR #578, opened 2016-07-25
Testing docs

xbill82:
Pulled together some information we used to find in the Vuejs forum and wrote it down here. Hope it helps.


docs/en/testing.md
```diff
@@ -0,0 +1,54 @@
...
+        }
+      },
+      replace: false
+    })
+
+    /* The router is initialized with the abstract option */
+    router = new VueRouter({ abstract: true })
+    router.map(Routes)
+
+    router.start(App, 'body')
```
yyx990803, line 34: we probably want to create a temporary div and append it to the body instead of mounting directly to the body, especially if we have other unit tests going on on the same page.

PR #589, opened 2016-07-29
fix some spell error

yolo2013:
change normalizePosotion to normalizePosition


src/util/path.js
```diff
@@ -44,9 +44,9 @@ export function resolvePath (
 }
 
 export function parsePath (path: string): {
-  path: string;
```
yyx990803, line 47: Can you leave these changes out? This is valid flow syntax.

PR #602, opened 2016-08-10
Expose meta in Route objects

fnlctrl:
Expose route.meta to allow reading RouteConfig.meta in transition hooks.

Reasons: 
In 1.0, we used to pass Custom Fields( like `auth` in route configs, as the [document](http://router.vuejs.org/en/route.html#custom-fields) suggested), but in 2.0 these custom fields aren't exposed in `route` object.
I've noticed that in `RouteConfig`, there is an optional field `meta`, which i guess serves the functionality of the previous Custom Fields, judging from the fact that 1). It's type is any 2).there are no usage references of it in the source code. 
So I guess `meta` should be exposed in route objects, otherwise there is no need to define it in the first place.

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -65,6 +65,7 @@ declare type Route = {
   fullPath: string;
   matched: Array<RouteRecord>;
   redirectedFrom?: string;
+  meta?: any;
 }
 
 declare type Matcher = (location: RawLocation, current?: Route) => Route;
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -126,7 +126,7 @@ export function createRoute (
 ): Route {
   const route: Route = {
     name: location.name || (record && record.name),
-    meta: record.meta || {},
+    meta: (record && record.meta) || {},
     path: location.path || '/',
     hash: location.hash || '',
     query: location.query || {},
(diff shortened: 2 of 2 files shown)


LinusBorg:
You already can access `meta`:

 In the next branches examples, there's already an example using the meta property:

set meta: 
https://github.com/vuejs/vue-router/blob/next/examples/navigation-guards/app.js#L53
Read meta:
https://github.com/vuejs/vue-router/blob/next/examples/navigation-guards/app.js#L61

fnlctrl:
@LinusBorg meta's not accessible in $route, unfortunately..

LinusBorg:
it's accessible through the routeRecord objects in the `$route.matched` array, if I'm not mistaken

https://github.com/fnlctrl/vue-router/blob/4c5c0469f2d8a4cc41f39f8409b188e59d7137f1/src/create-matcher.js#L135

This has the added benefit of providing access to the parent routeRecord's meta, as well.

fnlctrl:
@LinusBorg Thanks! I wasn't looking inside `matched` because it was something completely different in 1.0, and I couldn't find relevant information in the release notes. I would have argued that it's less convenient/straightforward than in 1.0 and then just move on to use route.matched instead, but since the PR is merged there's no need and I'm very grateful for this change :).

yyx990803:
@fnlctrl note the `meta` exposed is only that of the deepest matched sub route. In a lot of cases you will still want to iterate through all matched routes.

fnlctrl:
@yyx990803 I see...Thanks for the note!

Issue #608, opened 2016-08-12
[2.0] default named child routes not being matched


marcvdm:
### Vue.js & vue-router.js version

2.0.0.rc3
### Reproduction Link

https://jsfiddle.net/u5jjwzft/9/
### Steps to reproduce

Click on the links and you will see that when using the router-link with `:to="{name: 'parent'}"` wil not load default
### What is Expected?

When either you use `to="/parent"` or `:to="{name:'parent'}"` you expect to see the default child route component.
### What is actually happening?

It is not showing the default child route component.
On first page load of `/parent` it works but when clicking on one link and then to the named route it doesn't work.

yyx990803:
I'd say this is expected behavior - named routes are precise targets, i.e. when you specify `{ name: 'parent' }`, you are telling the router that you explicitly want to navigate to that route with no child route. If you want to navigate to the default child route, then you **should** be using the name of that child route instead.

joemsak:
I know this issue is old but I completely disagree. I made the default child the default for a reason and now I can never refer to my route with `{ name: 'my-route-name' }` for this particular view in my app. That makes no sense. If I change the URL of that view I can't rely on the named route logic to save me from the work of updating the links throughout the code.

I found this issue specifically because I was confused by this so-called "expected" behavior. I actually expected my default child to be rendered any time I visit that parent route/component, regardless of how I link to it

zevdg:
@joemsak I also found this behavior intuitive and it was problematic for me when I attempted to implement an "up" button to navigate up the site hierarchy.  I could programatically find the name of the current route's parent, but finding the name of that parent's defaultChild was problematic.  

Fortunately, there is a workaround described at https://github.com/vuejs/vue-router/issues/822#issuecomment-255685008 to give the behavior that we expect.
```javascript
new Router({
	routes: [{
		path: "/parent/:someProp",
		component: Parent,
		name: "parent",
		// explicitly reroute to default child on named navigation
		// https://github.com/vuejs/vue-router/issues/822#issuecomment-255685008
		redirect: "/parent/:someProp",
		children: [
			{
				path: "",
				name: "defaultChild",
				component: DefaultChild
			},
		// ...
```

PR #609, opened 2016-08-13
Fix for #608: fixing named routes matching for default child

marcvdm:
This is a fix for #608 

I used an underscore as prefix to define it as a default.

### DIFF
--- a/examples/named-routes/app.js
+++ b/examples/named-routes/app.js
@@ -6,6 +6,8 @@ Vue.use(VueRouter)
 const Home = { template: '<div>This is Home</div>' }
 const Foo = { template: '<div>This is Foo</div>' }
 const Bar = { template: '<div>This is Bar</div>' }
+const Parent = { template: '<div>This is Parent<br><router-view class="child-view">This is Parent</router-view></div>' }
+const Child = { template: '<div>This is Child</div>' }
 
 const router = new VueRouter({
   mode: 'history',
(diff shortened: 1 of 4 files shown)


LinusBorg:
Hi @marcvdm 

Thank you for our work on this PR.

Unfortunately, I will have to close it, since we decided to go another route to resolve this.

A few words on the reasoning:
- In #608, Evan already commented that this is expected default behaviour: If you transition to a named route, you expect to arrive exactly there, and get the route object for this route.
- However, as again discussed in #629, For the enduser this could still be confusing as a named route can behave differently that the same URL typed in the browser. So this has to be solves somehow.
- But the solution that you propsed with this PR would mean that a dev that wants to navigate to `name: 'home'` would not get that route object in $route, but would implicitly(!) get the route object for the default child route. This is also undesirable as it would confuse the developer without proper knowledge of this mechanism.

Therefore, to use the solution is: 
1. Leave the current behaviour as it is, 
2. But warn the developer when he uses a named route where a default child is present. 

The goal: developers should always use the named route of the default child, and simply not use the parent named route option at all.

``` js
// instead of this:
{
  path: 'parent',
  name: 'home'
  children: [
    {
      path: '',
      name: 'home.index'
    },
   // ...further children
  ]
}

// do this:
{
  path: 'parent', // no more named route here
  children: [
    {
      path: '',
      name: 'home'
    },
   // ...further children
  ]
}
```

PR #621, opened 2016-08-25
Add explanation of router modes

iWader:
Added an explanation of the router modes.

Resolves #445

PR #623, opened 2016-08-27
Explain server config for history mode

iWader:
Following on from #621 I've added docs to explain server configurations when using the router's history mode.

Just looking for feedback at this point. If this is heading in the right direction I'll add more examples for other webservers.

### DIFF
--- a/docs/en/server.md
+++ b/docs/en/server.md
@@ -15,13 +15,13 @@ There are a few caveats to this, in that your server will no longer report 404 e
 
 ```apache
 <IfModule mod_rewrite.c>
-    RewriteEngine On
-    RewriteBase /
-    RewriteRule ^index\.html$ - [L]
-    RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteCond %{REQUEST_FILENAME} !-d
-    RewriteRule . /index.html [L]
- </IfModule>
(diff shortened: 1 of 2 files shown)


docs/en/server.md
```diff
@@ -0,0 +1,43 @@
...
+  try_files $uri $uri/ /index.html;
+}
+```
+
+## Caveats
+
+To get around the issue that the server will no longer serve 404 errors you should implement a catch-all route within your Vue app to show a 404 page.
+
+```javascript
+router.map({
```
posva: This is actually the docs for vue-router 2, you have to use the new syntax:

``` js
new VueRouter({
  routes: [...]
})
```

Can you use a 2 spaces indentation, please?

posva:
Looks good as a starting point 😄 
@fnlctrl What do you think?

LinusBorg:
This will definitely be helpful.

fnlctrl:
@posva @LinusBorg  Looking good to me!

iWader:
I'll add a few further config examples tonight and see where we go from
there

posva:
@fnlctrl Merging this. @iWader you'll be able to create a new PR. You'll also be able to use the latest version

PR #632, opened 2016-09-01
Allow trailing slashes in path comparison (fix #628)

### LINKED ISSUE #628: [next] active links: allow trailing slash in route matching (nested routes)
I would like to have nested routes with a default component in the nested-root like so:

``` js
routes: [
  { 
    path: '/home', 
    name: 'home',
    component: Home,
    children: [
      {
        path: '',
        name: 'home.index',
        component: HomeIndex
      },
      {
        path: 'foo',
        name: 'home.foo',
        component: HomeFoo
      }
    ]
  }
]
```

For the active link highlights i'd like to scope the children separately. Can we serve the same content on both `/home` and `/home/` and have the `/home` link active in both cases?
### Vue.js & vue-router.js version

2.0.0-rc.4, 2.0.0-rc.4
### Reproduction Link

https://jsfiddle.net/jwahdatehagh/44ed5xxf/1/

fnlctrl:


LinusBorg:
Great!

Issue #624, opened 2016-08-29
Feature request: dynamic redirect


Centaur:
This request is for 2.0.
The non-exact route matching strategy is good, so that we can define a `router-link` as the entrance of one module which contains several subroutes and the `router-link` remains active when we are navigating in this module.  It is common that this `router-link` is defined as a redirect to some default subroute, e.g.:

``` js
var router = new VueRouter({
  routes:[{
     path: '/entry', redirect: '/entry/subroute'
  }, {
     path: '/entry/subroute', component: SubRoute
  }, {
     path: '/entry/anothersubroute', component: AnotherSubRoute
  }]
})
```

``` html
<router-link to="/entry">Entry</router-link>
<router-link to="/anothermodule">AnotherModule</router-link>
```

Every time we click the `Entry` link , we go to `SubRoute`, even if last time when we left `Entry` module we were at `AnotherSubRoute`.

If we want the `router-link` to be smart enough to go to the last subroute we visit in `Entry` module, we must first save where we were in `Entry` module some place, such as a store.

``` js
var store = {
    last_entry_route: ''
}
```

and update it whenever we enter a subroute.

Currently we can use an extra component to achieve the dynamic behavior:

``` js
const DynamicRedirect = Vue.extend({
  render(){
    this.$router.replace(store.last_entry_route)
  }
})
```

and use it like this: 

``` js
{path: '/entry', component: DynamicRedirect}
```

But this is quite a bit of code, and doing just a route redirect in `render` function does not feel quite right. I wish we can have a function as the `redirect` target of a route, i.e.

``` js
{path:'/entry', redirect: () => store.last_entry_route}
```

fnlctrl:
@posva @yyx990803 I think this is a nice-to-have feature! Should we implement it?

yyx990803:
@fnlctrl sounds good - feel free to give it a shot!

fnlctrl:
@yyx990803 Will do! :smile:

posva:
@fnlctrl  I think it's a nice idea too!

fnlctrl:
@yyx990803 
Implemented: https://github.com/vuejs/vue-router/pull/634

fnlctrl:
Closing as it has landed in rc.5

PR #634, opened 2016-09-01
Implement dynamic redirect (#624)

fnlctrl:
With e2e tests

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -53,7 +53,9 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
   ): Route {
     const { query, hash, params } = location
     const { redirect: originalRedirect } = record
-    const redirect = typeof originalRedirect === 'function' ? originalRedirect() : originalRedirect
+    const redirect = typeof originalRedirect === 'function'
+        ? originalRedirect(createRoute(record, location))
+        : originalRedirect
     const name = redirect && typeof redirect === 'object' && redirect.name
     if (name) {
       // resolved named direct
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -65,9 +65,9 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
         hash,
         params
       }, undefined, location)
-    } else if (typeof redirect === 'string') {
+    } else if (typeof redirect === 'string' || typeof redirect === 'function') {
       // 1. resolve relative redirect
-      const rawPath = resolveRecordPath(redirect, record)
+      const rawPath = resolveRecordPath(typeof redirect === 'function' ? redirect() : redirect, record)
       // 2. resolve params
       const path = fillParams(rawPath, params, `redirect route with path "${rawPath}"`)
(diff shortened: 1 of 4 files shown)


src/create-matcher.js
```diff
@@ -65,9 +65,9 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
         hash,
         params
       }, undefined, location)
-    } else if (typeof redirect === 'string') {
+    } else if (typeof redirect === 'string' || typeof redirect === 'function') {
       // 1. resolve relative redirect
-      const rawPath = resolveRecordPath(redirect, record)
+      const rawPath = resolveRecordPath(typeof redirect === 'function' ? redirect() : redirect, record)
```
fnlctrl: Should we cache `type of redirect === 'function'`? 
I didn't because I thought it was too trivial.

posva: It doesn't have an impact on performance so I think it's ok this way.

However, I think it's better to call redirect if it's a function and let the rest of the code the same:

`redirect = typeof redirect === 'function' ? redirect() : redirect`

This way the return of redirect can also be an object with a name and it's more flexible

fnlctrl: Yep! You're right, fixing it now

yyx990803:
Shouldn't the redirect function have access to the target route to be actually useful?

src/create-matcher.js
```diff
@@ -52,7 +52,8 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
     location: Location
   ): Route {
     const { query, hash, params } = location
-    const { redirect } = record
+    const { redirect: originalRedirect } = record
+    const redirect = typeof originalRedirect === 'function' ? originalRedirect() : originalRedirect
```
fnlctrl: @yyx990803 Sure it would be useful, but the target route is `match()`-ed after consuming the redirect option... and `this` context is not available here, so I don't know what can be passed to the redirect function...

yyx990803: we can pass `createRoute(record, location)`

fnlctrl: Yeah, that makes sense! On it now.

yyx990803:
Ideally also need a test case for the `to` argument ;)

fnlctrl:
Right.. I'll get back to it after dinner :smile:

fnlctrl:
```
http://localhost:8080/__build__/shared.js
Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/__build__/auth-flow.js 
Failed to load resource: the server responded with a status of 404 (Not Found)
```

Strange that I can't seem to run e2e tests on the current PC I'm using (was running fine on another one), `npm run dev` has the same error too...
Tried `npm cache clear` `npm install` many times still no luck..

Using CI for debug now.. sigh

fnlctrl:
Note:
Stuck on removing hash and query after redirect

posva:
Apart from tests failing LGTM

yyx990803:
Merging as-is and fixing it later

PR #636, opened 2016-09-01
Fix hash comparison for isSameRoute (fix #635)

### LINKED ISSUE #635: [next] active router-link with query
When `:to="{ name: 'xyz', query: {} }"` form is used, active class is not set correctly. Debugging revealed that `:exact="true"` links also compare `hash` with strict equality, so I was able to get this stuff working by including `hash: ""`, which is obviously not what one would ever expect ;)

Example: https://jsfiddle.net/borisokunskiy/yjm90k1j/1/

To repro just click through various links and observe classes set correctly only on "Exact links with hash="":" group. 

That being said, I believe that non-exact links being active regardless of query value is not expected either, but I wonder if that's by design.

fnlctrl:
Fix hash comparison for isSameRoute (#635)
when one of a.hash and b.hash is undefined while the other is empty string

This bug causes active class not being applied when using named view + exact matching

### DIFF
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -6,13 +6,13 @@ export function isSameRoute (a: Route, b: ?Route): boolean {
   } else if (a.path && b.path) {
     return (
       a.path === b.path &&
-      a.hash === b.hash &&
+      (a.hash || '') === (b.hash || '') &&
       isObjectEqual(a.query, b.query)
     )
   } else if (a.name && b.name) {
     return (
       a.name === b.name &&
-      a.hash === b.hash &&
+      (a.hash || '') === (b.hash || '') &&
       isObjectEqual(a.query, b.query) &&
       isObjectEqual(a.params, b.params)
     )
--- a/examples/active-links/app.js
+++ b/examples/active-links/app.js
@@ -25,7 +25,7 @@ const router = new VueRouter({
     { path: '/about', component: About },
     { path: '/users', component: Users,
       children: [
-        { path: ':username', component: User }
+        { path: ':username', name: 'user', component: User }
       ]
     }
   ]
@@ -48,6 +48,11 @@ new Vue({
         <li>
           <router-link :to="{ path: '/users/evan', query: { foo: 'bar' }}">
             /users/evan?foo=bar
+          </router-link>
+        </li>
+        <li><!-- #635 -->
+          <router-link :to="{ name: 'user', params: { username: 'evan' }, query: { foo: 'bar' }}" exact>
+            /users/evan?foo=bar (named view + exact match)
           </router-link>
         </li>
         <li>
--- a/test/e2e/specs/active-links.js
+++ b/test/e2e/specs/active-links.js
@@ -3,7 +3,7 @@ module.exports = {
     browser
     .url('http://localhost:8080/active-links/')
       .waitForElementVisible('#app', 1000)
-      .assert.count('li a', 10)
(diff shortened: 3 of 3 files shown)


src/util/route.js
```diff
@@ -6,13 +6,13 @@ export function isSameRoute (a: Route, b: ?Route): boolean {
   } else if (a.path && b.path) {
     return (
       a.path === b.path &&
-      a.hash === b.hash &&
+      (a.hash || '') === (b.hash || '') &&
```
posva, line 9: Feels a bit hacky to be done here. Can't route get normalised somewhere else?

fnlctrl: Hmm.. Yes it's hacky.. I'll look deeper and see if I can normalize it elsewhere.

fnlctrl: By definition, isSameRoute's parameter `a` and `b` is type `Route`, whose `hash` is a required string field, but sometimes b.hash is undefined. So there may be some bug flow wasn't able to discover.

posva: Actually, it's **is** required so the check has to be done somewhere

fnlctrl: @posva I think I've found the cause: It's in router-link's render function, where it passed a `Location` object (whose `hash` is optional) directly into isSameRoute:
see https://github.com/vuejs/vue-router/blob/next/src/components/link.js#L31-L33

I'm surprised that flow didn't warn on implicit type conversion: passing a `Location` object to `isSameRoute`, while it asked for a `Route`

posva: That's weird, this line: https://github.com/vuejs/vue-router/blob/next/src/util/route.js#L3 is telling that `b` can be a route or `null` so giving it a location should indeed raise an error...

fnlctrl: To sum up, @yyx990803 A more elegant fix for this (instead of the `hash || ''` fix) involves modifying `Location` type, or a converter from `Location` to `Route`:
1. `compareTarget` can be a `Location`
   https://github.com/vuejs/vue-router/blob/next/src/components/link.js#L31
2. and it's passed into `isSameRoute` whose second param is type `?Route` (which means `Route | void | null` if I understood correctly), but flow didn't give a warning
   https://github.com/vuejs/vue-router/blob/next/src/components/link.js#L33

posva: Location can be an empty object because everything is optional in it.
Is there a way on flow to say a Location must have either a name or a path?
I think using multiple types and then saying location is one of them can work

fnlctrl: @posva A naive approach can be

```
declare type Location = {
  _normalized?: boolean;
  name?: string;
  path: string;
  hash?: string;
  query?: StringHash;
  params?: StringHash;
} | {
  _normalized?: boolean;
  name: string;
  path?: string;
  hash?: string;
  query?: StringHash;
  params?: StringHash;
}
```

But it seems clumsy... I'll dig up flow's documents

posva: I think we are looking for this https://flowtype.org/docs/disjoint-unions.html#_

edit: which is actually what you did xD

It'd be great to combine unions with intersections

fnlctrl: @posva Tried that, but flow still doesn't give warnings : (
and even if we make it give warnings, we still have to fix passing `Location` to a function that requires `Route`..

yyx990803:
The root cause was because `components/link` was not flow-enabled... I've fixed it and made some adjustments in 270683f

PR #641, opened 2016-09-04
Fix  #629 - add warning when named route has a default child route.

### LINKED ISSUE #629: [next] Different behaviour for default subRoute when using path vs. named route
I have a nested routes with a default component:

``` js
routes: [
  { 
    path: '/home', 
    name: 'home',
    component: Home,
    children: [
      {
        path: '',
        name: 'home.index',
        component: HomeIndex
      }
    ]
  }
]
```

I'd expect for `HomeIndex` to be rendered into `Home`'s `<router-view>` when i enter `Home`, but when i navigate to it using `:to="{ name: 'home' }"`, only `Home` is rendered.

Using `to="/home"` directly works as expected.
### Vue.js & vue-router.js version

2.0.0-rc.4, 2.0.0-rc.4
### Reproduction Link

https://jsfiddle.net/jwahdatehagh/rdtgfrfb/

LinusBorg:
### What does this PR do?

This change will log a warning to the console when the developer has defined a named route that also has a default child route.  The warning reads like this: 

> [vue-router] Named Route 'bar' has a default child route.
> When navigating to this named route (:to="{name: 'bar'}"), the default child route will not be rendered.
> Remove the name from this route and use the name of the default child route for named links instead.

It will be logged to console once, during creation of the route map. I chose this to warn the user early and make him remove the name of the parent and instead only name the child.

Alternatively, we could log this warning each time there's a transition to such a route, but I would prefer to trigger upfront.
### Content of the Message

I'm sure the warning message can be improved, just leave a suggestion here.
### Tests
- I tested the warning with a modified version of the "named routes" example. 
- I did not add a test case to the e2e tests, 
  - because I did not want to put a "bad practice" into the example, 
  - to put that into the examples I would have to mix nested and names routes, which are currently not mixed in their examples.
- I did not add a unit test, because 
  - `create-route-map` has no unit tests so far.
  - I'm struggling to get the "hasBeenWarned()" Test-helper form vue working here, which would be required to test weither the warning was logged.


posva:
I don't think it's ok to show warnings like this on production and that we should add a variable to check the environment as Vue does

LinusBorg:
uhm - right, of course. I'll add it tonight hopefully. busy day.

LinusBorg:
Didn't have any time unfortunately. Will push the change tomorrow around this time.

LinusBorg:
added the check for production mode. also added a unit test.

fnlctrl:
@LinusBorg Great!
That's a lot of commits though, care to squash them?

LinusBorg:
Well when they are committed to the base branch they cant be squashed no? Still a little unfamiliar with certain git stuff. ^^

Also no idea why those e2e tests fail, they work locally.

fnlctrl:
Hmm... I'll clone and try testing.

As for squashing already pushed commits, the only option would be force pushing..
Here's a helpful SO post http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed

LinusBorg:
I'm referring to this option (the second one):

image

So we can squash once we merge this PR.

fnlctrl:
Wow that's a really useful github feature! Didn't know one before.

posva:
No need to manually squash. We can do that when merging with github 🙂

**edit**: Sent this on the metro 😁

yyx990803:
Can we remove the `npm-debug.log` from the PR?

LinusBorg:
Oh, of course. That shouldn't be in there, I merged a test branch into this one which I shared with the others tracking down a problem with Jasmine. 

Will remove it later, I'm at the office now.

fnlctrl:
@LinusBorg Tests are passing here too... Strange indeed it didn't pass on ci.

test/unit/specs/create-map.spec.js
```diff
@@ -0,0 +1,45 @@
...
+        component: Baz,
+        name: 'bar.baz'
+      }
+    ]
+  }
+]
+
+describe('Creating Route Map', function () {
+  beforeAll(function () {
+    console.warn = console.warn
```
LinusBorg: reminder: this can be removed as well.

src/create-route-map.js
```diff
@@ -43,6 +43,17 @@ function addRouteRecord (
   }
 
   if (route.children) {
+    // Warn if route is named and has a default child route.
+    // If users navigate to this route by name, the default child will
+    // not be rendered (GH Issue #629)
+    if (process.env.NODE_ENV !== 'production') {
```
posva, line 49: Shouldn't we use the replace plugin for rollup to replace this?

LinusBorg: Oh Jesus, the router prod build config does not have `process.env.NODE_ENV` set? 

Damn. time to learn rollup, haven't used it yet.

Seems indeed to be a job for the replace plugin, to simulate what the DefinePlugin for Webpack did.

LinusBorg: @posva: or maybe https://github.com/calvinmetcalf/rollup-plugin-node-globals ?

posva: I was thinking about this: https://github.com/rollup/rollup-plugin-replace
It has the same behaviour that is already used in other repositories like the wepback template

LinusBorg: done.

LinusBorg:
Soooo:
- removed the npm debug log
- added `rollup-plugin-replace`
- with that plugin, enabled rollup config to run in `production` and `development` modes
- added --environment option to npm build script.
- improved test case to test against both environments.

also built the final files and tested them with the examples - works, and no console warnings. didn't commit any of that of course. But should I commit `/dist`?

...but one e2e transition test fails me ...
- and I removed a debugger statement in abstract.js that shouldn't be in the repo i think.

can anyone take a second look? This PR got a bit bigger than expected ...

test/unit/specs/create-map.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+  it('has a nameMap object for default subroute at \'bar.baz\'', function () {
+    expect(this.maps.nameMap['bar.baz']).not.toBeUndefined()
+  })
+
+  it('in development, has logged a warning concering named route of parent and default subroute', function () {
+    process.env.NODE_ENV = 'development'
+    this.maps = createRouteMap(routes)
+    expect(console.warn).toHaveBeenCalled()
+    expect(console.warn.calls.argsFor(0)[0]).toMatch('vue-router] Named Route \'bar\'')
```
fnlctrl, line 48: Maybe we can just use

```
expect('blah blab blah').toHaveBeenWarned()
```

here, like in [vue's repo](https://github.com/vuejs/vue/blob/next/test/ssr/ssr-env.spec.js#L42)?
capture

And sorry for my lack of knowledge... but what is `console.warn.calls` ? Can't seem to be able to google for docs about it... tried "console.warn.calls" and "function.prototype.calls" but no luck :joy: Is it a non-standard V8 engine api?

src/create-route-map.js
```diff
@@ -43,6 +43,17 @@ function addRouteRecord (
...
+      if (route.name && route.children.some(child => /^\/?$/.test(child.path))) {
```
fnlctrl, line 50: We should also be checking for `route.children` here, I guess that's what caused the tests to fail on ci, but it's a mystery that they pass locally...

And in https://github.com/vuejs/vue-router/blob/next/examples/nested-routes/app.js#L29, it says

> an empty path will be treated as the default, e.g.

So maybe we can just use `child.path === ''` here, as `'/'` won't be recognized as a default?

LinusBorg: > We should also be checking for route.children here

That's already checked for on line 45.

>  I guess that's what caused the tests to fail on ci, 

The failed tests on CircleCI are e2e tests on transitions - the unit test (which include this test) run without error on CircleCI so I don't see the connection.

> So maybe we can just use child.path === '' here, as '/' won't be recognized as a default?

for default child routes both `'/'` and `''` work, that's why I test for both with the regex

fnlctrl: Oops...missed that line, my bad..:joy:

I'll continue investigating why e2e tests failed then.

test/unit/specs/create-map.spec.js
LinusBorg: `.toHaveBeenWarned()` is a custom jasmine test helper by Evan and I was not motivated to learn how to add it to vue-router as well, so I just used a[ jasmine spy](https://jasmine.github.io/2.4/introduction.html#section-Spies) on console.warn.

That spy is also the source of the[ `.calls`](https://jasmine.github.io/2.4/introduction.html#section-31) property that confused you ;)

fnlctrl: Oh... I see :laughing:

yyx990803:
The tests are passing after re-running without cached dependencies.

PR #647, opened 2016-09-08
modify the boostrap to bootstrap,in the file example->index.js

sunopar:
the word in vue-router->example->index.js line19
the origin word is 'boostrap', and i change it to 'bootstrap'


posva:
Thanks!

PR #665, opened 2016-09-17
Trouble shooting for vue-router 2.0

virus-warnning:


yyx990803:
It's necessary only if you want to use the `template` option. This is not specific to `vue-router` and has been documented on the new doc site: http://rc.vuejs.org/guide/installation.html#Note-on-NPM-Builds

PR #668, opened 2016-09-20
Warn on duplicate path inside `routes` array

fnlctrl:
Warn on duplicate path inside `routes` array, so 

``` js
{
  routes: [
    { path: '/foo', component: Bar },
    { path: '/foo', component: Foo }
  ]
}
```

will get a warning: `There is already a route config for '/foo', it will be overridden.`

### DIFF
--- a/test/helpers/to-have-been-warned.js
+++ b/test/helpers/to-have-been-warned.js
@@ -1,0 +1,44 @@
+if (typeof console === 'undefined') {
+  window.console = {
+    warn: function () {},
+    error: function () {}
+  }
+}
+
+function hasWarned (msg) {
+  var count = console.warn.calls.count()
+  var args
+  while (count--) {
+    args = console.warn.calls.argsFor(count)
+    if (args.some(containsMsg)) {
+      return true
+    }
+  }
+
+  function containsMsg (arg) {
+    if (arg instanceof Error) throw arg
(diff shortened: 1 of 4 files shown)


fnlctrl:
Oops...looks like e2e-tests-random-failing is happening again...They are passing locally.

posva:
LGTM
About tests, can't you relaunch them? Maybe 1s is not enough

fnlctrl:
@posva I have no experience with CI servers :joy:... @posva  @yyx990803 How do I relaunch them?

test/helpers/to-have-been-warned.js
```diff
@@ -0,0 +1,44 @@
...
+            pass: warned,
+            message: warned
+              ? 'Expected message "' + msg + '" not to have been warned'
+              : 'Expected message "' + msg + '" to have been warned'
+          }
+        }
+      }
+    }
+  })
+})
```
fnlctrl, line 44: Side note: This is copied and modified (console.error => console.warn) from the core repo: https://github.com/vuejs/vue/blob/next/test/helpers/to-have-been-warned.js

posva: I saw it haha.
It may be worth to create a package for jasmine in the future

PR #681, opened 2016-09-23
Add TypeScript declarations

ktsn:
vue-router 2.0's typings :)

I wrote them with referring the flow declarations and the doc. Interface names are followed to flow's.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "selenium-server": "^2.53.1",
     "typescript": "^2.0.3",
     "uglify-js": "^2.7.0",
-    "vue": "vuejs/vue#eef05ac9e93a2b841caf576e5bc3e85d28472594",
+    "vue": "^2.0.0-rc.7",
     "vue-loader": "^9.4.0",
     "webpack": "^1.13.1",
     "webpack-dev-middleware": "^1.6.1"
--- a/package.json
+++ b/package.json
@@ -5,10 +5,14 @@
   "author": "Evan You",
   "license": "MIT",
   "main": "dist/vue-router.js",
+  "typings": "types/index.d.ts",
   "files": [
     "dist/vue-router.js",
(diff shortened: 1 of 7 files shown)


types/vue.d.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import Vue = require("vue");
+import VueRouter = require("./index");
+import { Route, RawLocation } from "./index";
+
+declare module "vue/types/vue" {
+  interface Vue {
+    $router: VueRouter;
+    $route: Route;
```
ktsn, line 12: Strictly speaking, these properties are nullable but I wrote them as non-nullable intentionally for convenience. 
Because we would always have these properties practically if we use vue-router.

package.json
```diff
@@ -60,8 +65,9 @@
     "rollup-plugin-replace": "^1.1.1",
     "rollup-watch": "^2.4.0",
     "selenium-server": "^2.53.1",
+    "typescript": "^2.0.3",
     "uglify-js": "^2.7.0",
-    "vue": "^2.0.0-rc.6",
+    "vue": "vuejs/vue#eef05ac9e93a2b841caf576e5bc3e85d28472594",
```
posva: I guess this is temporary 😆

ktsn: Yeah, exactly. But current published typings on npm is outdated because `ComponentOptions` is now generic type...
Should we wait until next version of Vue is published?

yyx990803: yeah, I'm publishing rc7 today.

ktsn: OK, so I will rewrite this after published 😄

ktsn: Oops, I fixed vue dependency but it causes inconsistency vue and vue-template-compiler on CI...

yyx990803: I'll re-run it without cache

PR #682, opened 2016-09-23
Improve flow declarations

ktsn:
- update `Function` types to stricter format.
- define `Dictionary<T>` and replace all `{ [name: string}: T }` to it.

I think composing existing types with generics is more understandable than defining a new type alias.

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -5,6 +5,8 @@ declare module 'path-to-regexp' {
   }
 }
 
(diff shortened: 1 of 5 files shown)


posva:
That looks awesome! I don't know that much about flow yet and haven't tested locally for some other ocurrences but sof far LGTM!

yyx990803:
👍

PR #690, opened 2016-09-27
prevent trigger parent link

lifeforfun:
My problem encountered is that:

```
  <router-link tag="li" :to="item.path" data-toggle="dropdown-menu" data-trigger="hover">
      <a>parent menu</a>
      <ul v-if="item.children">
          <router-link v-for="child of item.children"  tag="li" :to="child.path">
               <a>child menu</a>
          </router-link>
      </ul>
  </router-link>
```

When i click a **child menu** , it first trigger the **child menu** and then trigger the **parent menu**, cause the route become **parent menu**.

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -41,7 +41,7 @@ export default {
     const on = {
       click: (e) => {
         e.preventDefault()
-        //prevent trigger parent link
+        // prevent trigger parent link
         e.stopPropagation()
         if (this.replace) {
           router.replace(to)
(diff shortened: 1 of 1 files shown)


src/components/link.js
```diff
@@ -41,6 +41,8 @@ export default {
     const on = {
       click: (e) => {
         e.preventDefault()
+        //prevent trigger parent link
```
posva: Here, a comment seems unnecessary to me

LinusBorg:
This problem was essentially fixed by [this commit](https://github.com/vuejs/vue-router/commit/e5f0964ccc8388153a2ae61633deef448f1faa17).

Generally I think stopPropagation is almost always a bad idea.

PR #693, opened 2016-09-27
Change router.go to router.push in router.replace

OriginalEXE:
router.replace documentation references router.go, when it should really reference router.push


fnlctrl:
Oops.. Thanks!

PR #739, opened 2016-10-08
Changed dynamic segment to userId

dharik:
:to="" and push() were using param name "userId" but the route path was using "/user/:id". This is confusing because how would params {userId:123} translate to :id?


posva:
Thanks! LGTM. Ping @fnlctrl

fnlctrl:
👍

PR #786, opened 2016-10-15
Add event property to router-link

### LINKED ISSUE #785: Navigate link on custom events
Hi there,

I'd love to have my `router-link` components be activated on a `mousedown` event, but it looks like the component is hard-coded to only navigate on a `click` event. Adding a property to activate the link on a custom event would be awesome.

AdamNiederer:
Resolves #785
The event property allows the link to be activated on a custom event.
It defaults to click, so this shouldn't be a breaking change.

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -21,7 +21,11 @@ export default {
     exact: Boolean,
     append: Boolean,
     replace: Boolean,
-    activeClass: String
+    activeClass: String,
+    event: {
+      type: String,
+      default: 'click'
+    }
   },
   render (h: Function) {
     const router = this.$router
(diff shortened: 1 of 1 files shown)


yyx990803:
Wondering what's the use case for this?

AdamNiederer:
Binding `mousedown` to big navigation links makes your website feel a lot more responsive - that was the exigence for writing the patch. If your app can spend the 50ms between mousedown and mouseup loading components, that's a huge boost to felt responsiveness.

I'm sure there's a use for the other mouse events (wheel and doubleclick come to mind), and possibly the HTML5 drag events.

I'm interested in having it as a native property simply because we use the mousedown trick everywhere.

fnlctrl:
`mousedown` wouldn't work in mobile browsers, so some people may need to bind `touchstart` as well, then we may end up having to support binding multiple events...
I think we can just keep it simple by binding events explicitly on `<router-link>`s.

src/components/link.js
```diff
@@ -21,7 +21,11 @@ export default {
     exact: Boolean,
     append: Boolean,
     replace: Boolean,
-    activeClass: String
+    activeClass: String,
+    event: {
+      type: String,
```
yyx990803, line 26: I think it should also accept `Array` type, e.g. `<router-link :event="['mousedown', 'touchstart']"></router-link>`

yyx990803 (changes requested):

Issue #761, opened 2016-10-12
just not render , no error shows


joe-lz:
main.js

```
// The Vue build version to load with the `import` command
// (runtime-only or standalone) has been set in webpack.base.conf with an alias.
import Vue from 'vue'

// 路由VueRouter
import VueRouter from 'vue-router'
Vue.use(VueRouter)
// 1. 定义（路由）组件。
// 可以从其他文件 import 进来
import indexComponent from './pages/App'
import projectsComponent from './pages/projects'

const index = indexComponent
const projects = projectsComponent
// 2. 定义路由
// 每个路由应该映射一个组件。 其中"component" 可以是
// 通过 Vue.extend() 创建的组件构造器，
// 或者，只是一个组件配置对象。
// 我们晚点在讨论嵌套路由。
const routes = [
  { path: '/', component: index },
  { path: '/projects', component: projects }
]
// 3. 创建 router 实例，然后传 `routes` 配置
// 你还可以传别的配置参数, 不过先这么简单着吧。
const router = new VueRouter({
  routes // （缩写）相当于 routes: routes
})
// 4. 创建和挂载根实例。
// 记得要通过 router 配置参数注入路由，
// 从而让整个应用都有路由功能
/* eslint-disable no-new */
new Vue({
  router
}).$mount('#app')



```

LinusBorg:
Hi  

Thanks for filing this issue! Could you please provide a small reporduction on jsfiddle.net so we can debug this without trying to replicate this ourselves?`

That would be very helpful, thanks!

ragnarlotus:
Hello, it happened to me too. In the documentation just says to add the Vue.use(VueRouter) if using module system, but it is not enough. You will also have to do the following.

Replace:
`new Vue({
  router
}).$mount('#app')`
With:
`new Vue(Vue.util.extend({ router }, App)).$mount('#app');`

posva:
@deulos Why would you extend the App instance?
It's much easier to do

``` js
new Vue({
  el: '#app',
  router,
  render: h => h(App)
})
```

ragnarlotus:
Because it was the way I found to solve it. Yours works too and I like it more.

Would be great to include it in router docs, will happen to more people for sure.

Thank you for the correction!

posva:
@deulos I'll check it tomorrow

posva:
@jeodiong any news?

joe-lz:
[...]

PR #791, opened 2016-10-16
Clarify example by removing Vue.util.extend

posva:
See https://github.com/vuejs/vue-router/issues/761#issuecomment-253646153


posva:
@fnlctrl Is this ok to you until we find a better way with @chrisvfritz ?

fnlctrl:
Yeah, LGTM!

PR #797, opened 2016-10-17
Allow calling instance methods before vue instantiation. (fix #795)

### LINKED ISSUE #795: history is undefined when the router.replace is called before injecting the router into Vue instance
I need to call `router.replace` before injecting the `router` into the `Vue` instance. But can't figure out why `history` property is undefined in that situation:

``` js
VueRouter.prototype.replace = function replace (location) {
  this.history.replace(location) // TypeError: Cannot read property 'replace' of undefined
}
```

Here is the [online example](http://jsfiddle.net/xgrjzsup/186/) (check the console) and the code snippet:

``` js
// 0. If using a module system, call Vue.use(VueRouter)

// 1. Define route components.
// These can be imported from other files
const Foo = { template: '<div>foo</div>' }
const Bar = { template: '<div>bar</div>' }

// 2. Define some routes
// Each route should map to a component. The "component" can
// either be an actual component constructor created via
// Vue.extend(), or just a component options object.
// We'll talk about nested routes later.
const routes = [
  { path: '/foo', component: Foo },
  { path: '/bar', component: Bar }
]

// 3. Create the router instance and pass the `routes` option
// You can pass in additional options here, but let's
// keep it simple for now.
  routes
})

try {
    router.replace('/bar');
} catch(e) {
    console.log(e); // TypeError: Cannot read property 'replace' of undefined
}

// 4. Create and mount the root instance.
// Make sure to inject the router with the router option to make the
// whole app router-aware.
const app = new Vue({
  router
}).$mount('#app')

// Now the app has started!
```

fnlctrl:
Previously it would throw errors.

Allowed methods: push, replace, go, back, forward

Reason:
These methods only try to manipulate browser history, and should work even when root vue instance has not been created.

Common use case:
Call `router.replace()` to set browser url state. This action doesn't require vue's existence, so it shouldn't throw an error when called before vue's instantiation.
 Otherwise, users may need to check router's actual mode after fallback (history or hash) and then call `history.replaceState` or `location.replace` accordingly.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -107,6 +107,14 @@ export class History {
       hook && hook(route, prev)
     })
   }
+
+  /**
+   * Dummy method to make flow happy...
+   * otherwise it won't let me call `this.history.onHashChange`
+   * inside a `switch case`...
+   * Advices on how to remove this method are much appreciated.
+   */
+  onHashChange () {}
 }
 
 function normalizeBase (base: ?string): string {
--- a/src/index.js
+++ b/src/index.js
@@ -2,8 +2,8 @@
 
 import { install } from './install'
 import { createMatcher } from './create-matcher'
-import { HashHistory } from './history/hash'
-import { HTML5History } from './history/html5'
+import { HashHistory, getHash } from './history/hash'
+import { HTML5History, getLocation } from './history/html5'
 import { AbstractHistory } from './history/abstract'
 import { inBrowser, supportsHistory } from './util/dom'
 import { assert } from './util/warn'
@@ -36,6 +36,20 @@ export default class VueRouter {
       mode = 'abstract'
     }
     this.mode = mode
+
+    switch (mode) {
+      case 'history':
+        this.history = new HTML5History(this, options.base)
+        break
+      case 'hash':
+        this.history = new HashHistory(this, options.base, this.fallback)
+        break
+      case 'abstract':
+        this.history = new AbstractHistory(this)
+        break
+      default:
+        assert(false, `invalid mode: ${mode}`)
+    }
   }
 
   get currentRoute (): ?Route {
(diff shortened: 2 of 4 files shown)


src/history/base.js
```diff
@@ -107,6 +107,14 @@ export class History {
...
     })
   }
+
+  /**
+   * Dummy method to make flow happy...
+   * otherwise it won't let me call `this.history.onHashChange`
+   * inside a `switch case`...
+   * Advices on how to remove this method are much appreciated.
+   */
+  onHashChange () {}
```
fnlctrl, line 117: ^ help needed.

src/index.js
```diff
@@ -51,19 +65,17 @@ export default class VueRouter {
...
+    switch (this.mode) {
       case 'history':
-        this.history = new HTML5History(this, options.base)
+        this.history.transitionTo(getLocation(this.history.base))
         break
       case 'hash':
-        this.history = new HashHistory(this, options.base, fallback)
+        this.history.transitionTo(getHash(), () => {
+          window.addEventListener('hashchange', () => {
+            this.history.onHashChange()
```
fnlctrl, line 75: This is what made flow unhappy..
Error log (when you remove the dummy method in the comment above):

```
src/index.js:75
 75:             this.history.onHashChange()
                              ^^^^^^^^^^^^ property `onHashChange`. Property not found in
 75:             this.history.onHashChange()
                 ^^^^^^^^^^^^ AbstractHistory

src/index.js:75
 75:             this.history.onHashChange()
                              ^^^^^^^^^^^^ property `onHashChange`. Property not found in
 75:             this.history.onHashChange()
                 ^^^^^^^^^^^^ HTML5History
```

Even when `this.history` is guaranteed to have `onHashChange` inside `case 'hash'`

fnlctrl:
CI bug again... sigh..

posva:
I don't know yet how to handle that dummy function, but this function is worth some extra tests

yyx990803:
merged as 83418bce4b66b0320047a0f7cbd50a3a2d2b4d1f

PR #806, opened 2016-10-19
Allow alias option to be an empty string (Fix #805).

### LINKED ISSUE #805: can't add  alias: '' to s child route (to make it behave like a default route.
Repro: https://jsfiddle.net/p7uk1pbv/1/

The other way around works: https://jsfiddle.net/p7uk1pbv/2/

Cause: https://github.com/vuejs/vue-router/blob/dev/src/create-route-map.js#L62

LinusBorg:
Fixes #805 
Previoulsly, alias paths were only added to the map when `if (route.alias)`. empty strings failed this test.


posva (approved):

posva:
@fnlctrl You can merge this once you approve it too 🙂

fnlctrl:
@posva Will do! I'll take a look tonight.

fnlctrl (approved):

PR #807, opened 2016-10-20
Reports about the build needed for the demo work

vinicius73:


posva:
I don't think this is necessary since there's an example at the bottom of the page

vinicius73:
This PR is based on feedback from my local community. More than one time they had doubts with this situation.

posva:
I'd rather add global note at the beginning of the guide: http://router.vuejs.org/en/essentials/getting-started.html

> All examples will be using the standalone version of vue to make template parsing possible

@fnlctrl What are your thoughts on this?

fnlctrl:
Yeah, that should be enough.

posva:
@vinicius73 Do you find that statement clear enough?

vinicius73:
yes @posva is much better so I will update.

posva (approved):
Ping @fnlctrl

fnlctrl (approved):

PR #809, opened 2016-10-20
router link now sets a hashed path to href when using hash-mode. (fix #780)

### LINKED ISSUE #780: Hash mode router-link generated anchor href not a hash url
For "hash" being the default mode, shouldn't the generated anchors generate hashed URLs?  (For cases where one would want to copy URL or open link in a new tab/window)

The following `router-link`:

``` html
<router-link to="/example">Example</router-link>
```

Generates:

``` html
<a href="/example">Example</a>
```

Expected:

``` html
<a href="#/example">Example</a>
```

Clicking on the generated (non-hashed) anchor seems to work, it changes the URL to the hashed version of the URL.

LinusBorg:
#### Problem

When suing hash mode, router-link would still set a path without a hash to the `href`attribute (see #780)
#### Solution
- added function that constructs href path depeding on router mode. (link.js)
- added `/examples` for hash mode (there were none)
- added `/e2e` tests for new example.

### DIFF
--- a/examples/hash-mode/app.js
+++ b/examples/hash-mode/app.js
@@ -1,0 +1,42 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+// 1. Use plugin.
+// This installs <router-view> and <router-link>,
+// and injects $router and $route to all router-enabled child components
+Vue.use(VueRouter)
+
+// 2. Define route components
+const Home = { template: '<div>home</div>' }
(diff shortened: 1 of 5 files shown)


fnlctrl (approved):

fnlctrl:
ping @posva

posva (approved):
@fnlctrl Looks good!

PR #816, opened 2016-10-21
Fixed typo

vuchl:


fnlctrl:
It actually should be `and`, mind fixing that?

posva (approved):

PR #819, opened 2016-10-22
Correct replaceHash function

aurelienlhx:
Replace slice 0 to 0 when there is no hash with slice 0 to url length. 
Exemple if the application is on a suburl as http://www.exemple/app/ a replace is done from http://www.exemple.com/app/ to http://www.exemple.com/#/ instead of http://www.exemple.com/app#/. This changement allow to correct that and keep the root.


yyx990803:
This is not necessary - `location.replace('#abc')` preserves current path.

PR #830, opened 2016-10-25
Support `target="_blank"` for `<router-link>`

fnlctrl:


src/components/link.js
```diff
@@ -49,6 +49,11 @@ export default {
         // don't redirect on right click
         /* istanbul ignore if */
         if (e.button !== 0) return
+        // don't redirect if `target="_blank"`
+        /* istanbul ignore if */
+        const target = this.$el && this.$el.attributes.getNamedItem('target')
+        if (target && target.value === '_blank') return
```
phanan: Two things:
- Why not `const target = this.$el && this.$el.getAttribute('target')`?
- Won't this fail for `target="_blank "` (with an extra space)?

phanan (commented):
I don't really know how to use this GitHub review. Should have made it a comment instead :P

src/components/link.js
fnlctrl: Fixed, I guess I'm just too unfamiliar with native dom api :joy:

src/components/link.js
```diff
@@ -49,6 +49,11 @@ export default {
         // don't redirect on right click
         /* istanbul ignore if */
         if (e.button !== 0) return
+        // don't redirect if `target="_blank"`
+        /* istanbul ignore if */
+        const target = this.$el && this.$el.getAttribute('target')
+        if (target && target.toLowerCase().split(' ').indexOf('_blank') > -1) return
```
hogart: Since code is written in ES Net, shouldn't it be `target.toLowerCase().split(' ').includes('_blank')`?

fnlctrl: That cannot be transpiled, and would require a polyfill.

posva (approved):

posva:
@fnlctrl I absolutely forgot about this, sorry

fnlctrl:
@posva Totally forgot this too 😂😂😂
I think I'll just wait for Evan to get back to this.

src/components/link.js
```diff
@@ -49,6 +49,11 @@ export default {
         // don't redirect on right click
         /* istanbul ignore if */
         if (e.button !== 0) return
+        // don't redirect if `target="_blank"`
+        /* istanbul ignore if */
+        const target = this.$el && this.$el.getAttribute('target')
```
yyx990803: We should use `e.target` here. `this.$el` may be a wrapping `<li>` if `tag` prop is used.

src/components/link.js
yyx990803: how about just `/\b_blank\b/i.test(target)`

yyx990803 (changes requested):

bedlate:
@yyx990803 , hello you da, but how can I do it in the instance, like this:

`this.$router.push({name: 'users.index', target: '_blank'})`

Mouvedia:
If it's supported you should update [this line](https://github.com/vuejs/vuejs.org/commit/489145a05fc875660f1d5e0e57946629eeea8a8a#diff-2ff218682023cc310e3acd06a7271e63R270).

PR #834, opened 2016-10-25
Correction and consistency improvement

zaccharles:


fnlctrl:
Thanks!

docs/en/api/route-object.md
```diff
@@ -10,15 +10,15 @@ The route object can be found in multiple places:
 
 - As the return value of calling `router.match(location)`;
 
-- Inside navigation guards as the first argument:
+- Inside navigation guards as the first and second argument:
```
phanan, line 13: Shouldn't it be `arguments` (plural)?

zaccharles: I think you're right. I'll raise another PR to correct this.

zaccharles (commented):
lol wrong button...

docs/en/api/route-object.md
zaccharles: #842

fnlctrl: Well I guess it should be `as the first two arguments`.
https://www.quora.com/Which-one-is-grammatically-correct-The-first-and-the-second-chapter-is-very-difficult-or-The-first-and-second-chapters-are-very-difficult

zaccharles: I think that is more a matter of preference, as I think both are correct. Since you're a vuejs member, I'll change it to your preference.

PR #862, opened 2016-11-02
Fix lazy-loading routes with Webpack 2

atinux:
Hi,

I'm using Webpack 2 and there is a bug on the resolve method inside `resolveAsyncComponents`.
It seems by using Webpack 2, resolve is called twice, first with the resolved Component, and second with undefined.

You can reproduce this error by the fork I made:
```bash
git clone https://github.com/Atinux/vue-router-bug.git
cd vue-router-bug
npm install
npm run dev
``` 

Then, go to [http://localhost:8080/lazy-loading/](http://localhost:8080/lazy-loading/) and navigate between the links.

Please see it in action:

My fix if to only check if resolvedDef is defined before updating the matched route.components.

Thank you :fire:

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -205,7 +205,9 @@ function resolveAsyncComponents (matched: Array<RouteRecord>): Array<?Function> 
     if (typeof def === 'function' && !def.options) {
       return (to, from, next) => {
         const resolve = resolvedDef => {
-          match.components[key] = resolvedDef
+          if (resolvedDef) {
+            match.components[key] = resolvedDef
+          }
           next()
         }
 


fnlctrl:
Hi, thanks for creating this PR.
However, it's actually not caused by vue-router, but rather the inconsistent behavior of webpack's `require`. So I don't think we should add a check here and pretend everything is working fine. You may create an issue at webpack's repo, thanks!

atinux:
Actually it's because your resolve method is called twice, I think also it's a bad behavior to depend on webpack and not checking the value before assigning to the router components. Don't you think?

fnlctrl:
> Actually it's because your resolve method is called twice

The resolve method is called by webpack, sorry but there's nothing we can do about that.

> I think also it's a bad behavior to depend on webpack and not checking the value before assigning to the router components

Vue and vue-router consistently receives and registers components as-is, and never checks for their validity. I think if they were to be checked, throwing an error is better than doing nothing with an `if`. You may open an issue at Vue's repo about Vue's not rejecting invalid component objects (undefined, null) if you'd like, but it won't solve the issue here since it's caused by webpack.

atinux:
I see, so I guess there should be a warning on the vue-router README to say that vue-router does not support lazy-loading with webpack 2 actually?

fnlctrl:
It actually does, as long as you use webpack 2's recommended `System.import`

atinux:
Good, what about this:

``` js
(resolve) => require.ensure([], () => resolve(require('Foo.vue')), 'Foo')
```

How can you use System.import for grouping the components?

atinux:
Ok just found it, it has to be:

``` js
const Bar = r => require.ensure([], () => System.import('./Bar.vue'), '/bar')
```

Thanks for your help @fnlctrl

johannes-z:
@Atinux I stumbled upon this problem as well. I think the "fix" you provided in your list time is applicable if you are using chunks, right? How would I need to transform the following line of code where I'm not using chunks? I couldn't find out how to use `System.import` in that case.

``` js
const Bar = (resolve) => require(['./Bar.vue'], resolve)
```

atinux:
Hi @johannes-z, what do you mean by using chunks? Is it in your webpack config?

johannes-z:
@Atinux 
In my webpack config I only specify the `publicPath`. The code "fix" you provided is for using chunks I guess, refering to the [official documentation](https://router.vuejs.org/en/advanced/lazy-loading.html):

Lazy Loading Routes:

``` js
const Foo = resolve => require(['./Foo.vue'], resolve)
```

Grouping Components in the Same Chunk:

``` js
const Bar = r => require.ensure([], () => r(require('./Bar.vue')), 'group-foo')
```

So instead of `r(require('./Bar.vue'))` you used `System.import('./Bar.vue')`. I don't want to group my components in chunks, though.

atinux:
I see @johannes-z 

You just have to do:

``` js
const Foo = () => System.import('./Foo.vue')
```

And that's it 👍

fnlctrl:
For future reference: 
`System.import` has been deprecated by yesterday's release of webpack 2.0: https://github.com/webpack/webpack/releases
From that version on, please use `import()` instead.

cobrAJ:
Did this problem has other solutions ?
I haven't use webpack or nodeJs or ES6 , I just use the RequireJS+ VueJs 2.0 .
Can I fixed this problem or not ?

PR #869, opened 2016-11-03
updated file based on vue-cli

ShukriChiu:
vue-cli project is standalone version of vue and should modify some codes to pass this example


fnlctrl:
同https://github.com/vuejs/vue-router/pull/867#issuecomment-258068160

Issue #831, opened 2016-10-25
Reverse URL resolving should be public method


hogart:
"Reverse URL resolving" means "converting name and params to URL string".

Right now there's only `VueRouter#match` method, which is great, but it's not documented, doesn't take `base` and `mode` into account, and doesn't normalize relative urls, so it's not really an equivalent of what's going inside `<router-link>` `render` method. 

Rationale for making reverse URL resolving public is following:
- In isomorphic apps we may need to send some link via email, so we should be able to get this link as  text
- Several kind of apps, (e.g. google drive, bit.ly) require to provide link which can be copied by user
- Issues like #827 where `<router-link>` doesn't support some attribute will have simple work-arounds

yyx990803:
resolved via merging #918

PR #877, opened 2016-11-04
This PR addresses issue #831, adding public Router#resolve method 

hogart:
Along with necessary typings and docs.

From technical perspective this change adds protected `_resolve` and public `resolve` methods, latter being wrapper around first. `_resolve` is used in `<router-link/>`, since it requires a lot of additional info. `resolve` just returns relevant bits from that.

Function `createHref` was moved to `src/index.js`, but maybe belongs in `sc/util/path.js`.


posva:
Thanks for the pr!
I'd rather have a single resolve method. Let's wait for @fnlctrl and @yyx990803 input

src/components/link.js
```diff
@@ -26,14 +24,10 @@ export default {
   render (h: Function) {
     const router = this.$router
     const current = this.$route
-    const to = normalizeLocation(this.to, current, this.append)
-    const resolved = router.match(to)
-    const fullPath = resolved.redirectedFrom || resolved.fullPath
-    const base = router.history.base
-    const href = createHref(base, fullPath, router.mode)
+    const { normalizedTo, resolved, href } = this.$router._resolve(this.to, current, this.append)
```
posva: you can directly call `router._resolve`

hogart:
> I'd rather have a single resolve method.

Me too. We can rename that internal method, I think.

LinusBorg:
I like this PR. @hogart, Can you solve the conflicts?

hogart:
Thanks, @LinusBorg. I merged my branch with upstream.

LinusBorg:
@hogart sorry to bother again. This also needs a a unnit test for the new #resolve() function. Could you add one?

fnlctrl:
Sorry I forgot this... I agree with @posva that we should use a single resolve method, both internally and as public api. Let's wait for Evan's input on this.

hogart:
@LinusBorg I'd like to, but I didn't found place for testing Router class.

@fnlctrl, @posva how do you imagine it? We need several pieces of info in `link` component, most of which are useless outside of that context. On the other hand, resolving algorithm stays the same, so it's either this or duplicating code.

posva:
@hogart By one method I meant that I'd rather have `_getResolutionData` inside `resolve` and use resolve everywhere

hogart:
@posva I understand what you want, but, sorry, I don't see how we can use `resolve` in `link` component, because it requires additional data. 
Can you please elaborate how in your vision it should work?

posva:
@hogart Oh, I wanted to say that resolve would return the whole object not just an `href` 😆

hogart:
@posva that's definitely possible:) But in my opinion this pollutes API surface, are you guys sure about that?

posva:
no, I personally think it's ok. Why does it pollutes the api surface, is it because you may never use the rest of the returned object?

hogart:
@posva yes, because of that.
Ok, I implement appropriate changes. Anything to land this PR already:)

hogart:
I'm closing this PR because messed-up changelog, see #918 instead.

PR #894, opened 2016-11-10
use push instead of replace in ensureURL

### LINKED ISSUE #887: `beforeRouteLeave` NOT called in a second time of `history.back()`
when i call `history.back()` twice, the `beforeRouteLeave` hook will NOT be called in the 2nd time

Reproduction:
http://plato.crossjs.com/test.html#/

vvenv:
Fixes #887

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -17,7 +17,7 @@ export class History {
   go: (n: number) => void;
   push: (loc: RawLocation) => void;
   replace: (loc: RawLocation) => void;
-  ensureURL: () => void;
+  ensureURL: (push?: boolean) => void;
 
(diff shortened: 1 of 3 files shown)


fnlctrl:
Hmm, using `pushState`/`pushHash` is a valid fix for the scenario in #887, though it shouldn't be done in `ensureUrl`, instead they should only be triggered after `next(false)` is called, here: https://github.com/vuejs/vue-router/blob/dev/src/history/base.js#L70-L72. 
Would you mind fixing it? Should probably add a parameter to `ensureURL` to decide whether to call `replace` or `push`.

/ping @posva @LinusBorg

fnlctrl (approved):

PR #899, opened 2016-11-12
Router is able to recognize components by string

akoumjian:


fnlctrl:
Thanks! Maybe `a component name that has been registered with Vue.component()` can be clearer?

akoumjian:
We could expand the example to use one of each possible entries and demonstrate by example. Also open to alternative language.

docs/en/essentials/getting-started.md
```diff
@@ -40,7 +40,8 @@ const Bar = { template: '<div>bar</div>' }
 // 2. Define some routes
 // Each route should map to a component. The "component" can
 // either be an actual component constructor created via
-// Vue.extend(), or just a component options object.
+// Vue.extend(), a component name created with Vue.component(),
```
yyx990803: "a component id string registered via `Vue.component()`"

yyx990803 (changes requested):

PR #905, opened 2016-11-15
[1.0] Conditional saving of scroll position in HTML5 history backend

Hagith:
Scroll position is saved only when ```saveScrollPosition``` router option is set.

This fixes integration of [autotrack](https://github.com/googleanalytics/autotrack/) with vue-router.
Additional replaceState() call doesn't play well with [urlChangeTracker plugin](https://github.com/googleanalytics/autotrack/blob/master/docs/plugins/url-change-tracker.md) and causes that ```pageview``` isn't sent.

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -16,7 +16,7 @@ export default class HTML5History {
       this.root = null
     }
     this.onChange = onChange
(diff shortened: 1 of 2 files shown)


posva:
Closing as there isn't any ongoing development going on v1

PR #910, opened 2016-11-17
Remove extras from old params in named routes

### LINKED ISSUE #906: Using named route persists old route params
### Vue.js version
2.0.5

### Vue-Router version
2.0.2

### Reproduction Link

https://jsfiddle.net/erikgall/spz2j6o6/

### Steps to reproduce / What is happening

The problem occurs when navigating back to a parent route (or sibling route) with no parameters from a child route that has a parameter. The bug seems to only happen when using named routes as well.

```
// current path: /students/classrooms/1
<router-link :to="{ name: 'students.classrooms'  }">All Classrooms</router-link>
```

#### Component/Route Hierarchy

root component --> layout component (ex. student, instructor, guest) --> page component

-> ./app.vue
---> ./students.vue
------->./students/index.vue
-------> ./students/classrooms/index.vue (path: `students/classrooms`)
-------> ./students/classrooms/show.vue (path: `students/classrooms/:classroom`)

When navigating from the show.vue component to any other component the classroom param persists and is not removed.

`students/classrooms/:classroom` to `students/classrooms`

This is an issue if you depend on using a v-if directive to render a component that should be inserted if a route param exists like in the example below.

### What is Expected?

If a route contains a param it should be present in the `vm.$route.params` object and when navigating from that route to a new one the params object is updated to reflect the current url.

znck:
Fixes #906

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -14,6 +14,10 @@ const regexpCache: {
   }
 } = Object.create(null)
 
+const regexpParamsCache: {
+  [key: string]: Array<string>
+} = Object.create(null)
+
 const regexpCompileCache: {
   [key: string]: Function
 } = Object.create(null)
@@ -31,6 +35,7 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
 
     if (name) {
       const record = nameMap[name]
+      const paramNames = getParams(record.path)
 
       if (typeof location.params !== 'object') {
         location.params = {}
@@ -38,7 +43,7 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
 
       if (currentRoute && typeof currentRoute.params === 'object') {
         for (const key in currentRoute.params) {
-          if (!(key in location.params)) {
+          if (!(key in location.params) && paramNames.indexOf(key) > -1) {
             location.params[key] = currentRoute.params[key]
           }
         }
@@ -196,6 +201,26 @@ function fillParams (
     assert(false, `missing param for ${routeMsg}: ${e.message}`)
     return ''
   }
+}
+
+function getParams (path: string): Array<string> {
+  const hit = regexpParamsCache[path]
+
+  if (hit) {
+    return hit
+  }
+
+  const keys = []
+  const paramNames: Array<string> = []
+
+  Regexp(path, keys)
+
+  for (let i = 0, len = keys.length; i < len; ++i) {
+    const key = keys[i]
+    if (key && 'name' in key) paramNames.push(key.name)
+  }
+
+  return (regexpParamsCache[path] = paramNames)
 }
 
 function resolveRecordPath (path: string, record: RouteRecord): string {
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -155,13 +155,10 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
   return match
 }
 
-function matchRoute (
-  path: string,
-  params: Object,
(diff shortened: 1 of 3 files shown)


fnlctrl:
Thanks for the PR! I'll review it asap.

src/create-matcher.js
```diff
@@ -198,6 +203,26 @@ function fillParams (
...
+  const hit = regexpParamsCache[path]
+
+  if (hit) {
+    return hit
+  }
+
+  const keys = []
+  const paramNames: Array<string> = []
+
+  Regexp(path, keys)
```
fnlctrl: We don't have to call `Regexp(path)` here again (without cache, actually), there's already code for getting cached keys and regexp here (https://github.com/znck/vue-router/blob/ab0a779ba6f09f5b28138c9445ce84d7263daef6/src/create-matcher.js#L163-L172). 

We should probably create a separate function that uses the code above, take a `path` as only parameter, returns a `{keys, regexp}` object, maybe name it `getKeysAndParams`.

Then we can get keys with ES6 destructuring

``` js
const {keys} = getKeysAndParams(path)
```

src/create-matcher.js
```diff
@@ -198,6 +203,26 @@ function fillParams (
...
+
+  for (let i = 0, len = keys.length; i < len; ++i) {
+    const key = keys[i]
+    if (key && 'name' in key) paramNames.push(key.name)
+  }
```
fnlctrl: This whole for-loop can be simplified to `keys.map(k => k.name)` for brevity. Since we are already using cache here, the performance penalty of using a `Array.prototype.map` is negligible.

src/create-matcher.js
```diff
@@ -31,14 +35,15 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
 
     if (name) {
       const record = nameMap[name]
+      const paramNames = getParams(record.path)
```
fnlctrl: We can use [similar code here](https://github.com/znck/vue-router/blob/ab0a779ba6f09f5b28138c9445ce84d7263daef6/src/create-matcher.js#L196-L198) to avoid a separate  getParams function

fnlctrl (changes requested):
@znck  Also, can we add accompanying e2e tests for it (that addresses https://github.com/vuejs/vue-router/issues/906)?
The test provided in the original PR (https://github.com/vuejs/vue-router/pull/851/files) didn't actually cover the case of `using extra params`, we should probably fix it too so that this PR can cover both cases (bugfix and using extra params)

src/create-matcher.js
```diff
@@ -198,6 +203,26 @@ function fillParams (
...
+  const paramNames: Array<string> = []
+
+  Regexp(path, keys)
+
+  for (let i = 0, len = keys.length; i < len; ++i) {
+    const key = keys[i]
+    if (key && 'name' in key) paramNames.push(key.name)
+  }
+
+  return (regexpParamsCache[path] = paramNames)
```
fnlctrl: This can then be simplified to 

``` js
return regexpParamsCache[path] || 
   (regexpParamsCache[path] = keys.map(k => k.name))
```

like https://github.com/znck/vue-router/blob/ab0a779ba6f09f5b28138c9445ce84d7263daef6/src/create-matcher.js#L196-L198

fnlctrl (approved):
/ping @yyx990803

jeerbl:
+1

yyx990803:
👍 great work @znck and @fnlctrl

PR #917, opened 2016-11-20
Don't use old optional params in routes

### LINKED ISSUE #906: Using named route persists old route params

znck:
Fixes #906 
Related #910

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -212,7 +212,9 @@ function fillParams (
 
 function getParams (path: string): Array<string> {
   return regexpParamsCache[path] ||
-    (regexpParamsCache[path] = getRouteRegex(path).keys.map(key => key.name))
+    (regexpParamsCache[path] = getRouteRegex(path).keys
+      .filter(key => !key.optional)
+      .map(key => key.name))
 }
 
 function resolveRecordPath (path: string, record: RouteRecord): string {
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -35,7 +35,10 @@ export function createMatcher (routes: Array<RouteConfig>): Matcher {
 
     if (name) {
       const record = nameMap[name]
-      const paramNames = getParams(record.path)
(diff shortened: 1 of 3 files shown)


znck:
Ping @fnlctrl

src/create-matcher.js
```diff
@@ -212,7 +212,9 @@ function fillParams (
 
 function getParams (path: string): Array<string> {
```
fnlctrl: Probably should rename it to `getRequiredParams` to avoid ambiguity,
or maybe remove this function and use inline expression at https://github.com/znck/vue-router/blob/c3ec53191647d5a4dfb7481ea3c4d088f2c6c6d7/src/create-matcher.js#L38

fnlctrl (approved):

yyx990803:
Thanks again @znck !

PR #925, opened 2016-11-22
avoid rearrange query params

miccycn:
avoid rearrange query parameters, 
coz it will cost a new request 
and make some troubles like generating token which is based on query string


yyx990803:
There is one test failing 🙈

miccycn:
:( Should I modify the unit test file?....Umm, I might as well to create an ISSUE....

PR #946, opened 2016-11-28
Avoid rearrange query parameters

miccycn:
https://github.com/vuejs/vue-router/issues/926

### DIFF
--- a/src/util/query.js
+++ b/src/util/query.js
@@ -55,7 +55,7 @@ function parseQuery (query: string): Dictionary<string> {
 }
 
 export function stringifyQuery (obj: Dictionary<string>): string {
-  const res = obj ? Object.keys(obj).sort().map(key => {
+  const res = obj ? Object.keys(obj).map(key => {
     const val = obj[key]
 
     if (val === undefined) {
--- a/test/e2e/specs/active-links.js
+++ b/test/e2e/specs/active-links.js
@@ -14,7 +14,7 @@ module.exports = {
       .assert.attributeContains('li:nth-child(6) a', 'href', '/active-links/users/evan#foo')
       .assert.attributeContains('li:nth-child(7) a', 'href', '/active-links/users/evan?foo=bar')
       .assert.attributeContains('li:nth-child(8) a', 'href', '/active-links/users/evan?foo=bar')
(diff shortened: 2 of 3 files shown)


fnlctrl (approved):

LinusBorg:
The order of keys returned by `Object.keys()` is [not](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) [reliable](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Statements/for...in):

> The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop

> A for...in loop iterates over the properties of an object in an arbitrary order (see the delete operator for more on why one cannot depend on the seeming orderliness of iteration, at least in a cross-browser setting).

so I don't think that simply removing the `.sort()` could produce a reliable solution to the problem of #926 and will instead break "randomly".

fnlctrl:
@linusborg It's been discussed in #926, sorry it's in Chinese since OP switched to it in the middle of it. Basically the reasons are:
1. Developer should never rely on the order of query parameters. So removing `.sort()` shouldn't be a breaking change for those who don't. If they do rely on query order, it should serve as a warning to let them switch to dynamic segments.
2. OP agreed that this change is unreliable, but it works well on his target browsers so it's good enough for him. Since only he cares about it (I obviously don't), and as said in 1). people should never rely on query order, it's a good compromise to avoid complicating current code just to preserve query params order.
3. Evan replied him in weibo (Chinese version of twitter) and instructed him to remove the `sort`, so I think it 's fine.

miccycn:
Yes, it is not reliable. 

It is not the best way to solve it, but I think Vue-Router is not responsible for sorting the query string by its keys.
But now it cause some trouble.
If Vue-Router dose not do that, once there are problems in this case, it's browser`s responsibility, not Vue-Router.

LinusBorg:
Well, that is one way to see it, but another way would be that because the browser does not guarantee order, vue-router does, with `.sort` - but I can also understand your problem with #926 

Both scenarios are not optimal.

miccycn:
@LinusBorg
Actually, I am  more curious about what it is necessary to do so, coz other functions seems do not need an ordered query.

PR #973, opened 2016-12-02
New feature: Passing props to components from vue-router

bfanger:
Decouple component from the router by using props:

Documentation: https://github.com/bfanger/vue-router/blob/dev/docs/en/essentials/passing-props.md
Examples:  https://github.com/bfanger/vue-router/tree/dev/examples/route-props


fnlctrl:
Interesting idea!
/ping @posva @LinusBorg

src/util/props.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+      }
+      return props
+    default:
+      warn(false, `props in "${route.path}" is a ${typeof config}, expecting an object, function or boolean.`)
+  }
+}
+
+function hasProp (component, prop) {
+  if (!component.props) {
```
fnlctrl: Can probably skip this ` if (!component.props)` check since it's already done in https://github.com/bfanger/vue-router/blob/5ad9d480eeed6700a26597484d0c0454d30a13e3/src/util/props.js#L20-L22
Or maybe just inline this function and use the `typof` expression..

src/util/props.js
```diff
@@ -0,0 +1,46 @@
...
+        return
+      }
+      const props = {}
+      for (const prop in route.params) {
+        const value = route.params[prop]
+        if (hasProp(component, prop)) {
+          const propType = component.props[prop].type
+          if (propType === null || propType === String) {
+            props[prop] = value
+          } else if (propType === Number && value.match(/^-?([0-9]+|[0-9]*\.[0-9]+)$/)) {
```
fnlctrl: Should probably cache this regex literal `/^-?([0-9]+|[0-9]*\.[0-9]+)$/`

src/components/view.js
```diff
@@ -49,6 +51,7 @@ export default {
           matched.instances[name] = undefined
         }
       }
+      data.props = resolveProps(route, component, matched.props && matched.props[name])
```
fnlctrl: ~~Should merge with the original `data.props`, or it would be a breaking change as the current code disallows explicitly passing props to `<router-view>`, like `<router-view foo="bar"/>`~~

Edit: Sorry, I was wrong about that..

bfanger: The props of `<router-view>` are still passed to the rendered element. The `<div>` in Hello still gets the `class="view"`

src/util/props.js
```diff
@@ -0,0 +1,46 @@
...
+        return
+      }
+      if (!component.props) {
+        return
+      }
+      const props = {}
+      for (const prop in route.params) {
+        const value = route.params[prop]
+        if (hasProp(component, prop)) {
+          const propType = component.props[prop].type
```
fnlctrl: I'm not sure about reading prop types at runtime and converting them to `Number` if possible.. Maybe we should just leave them as string since url params are string by nature. And if `Number` was supported, should we also support `Boolean`?

Also, if this type conversion were to be supported, `type` can be an array, so we'll have to loop over that should it be the case.

bfanger: Not ideal, but the alternative was either not to support Number type, or pass it as an string and get a propType warnings and subtle bugs in `itemsById.find(item => item.id === id)`

I think routing to ID is a common use-case to allow it in the {props: true} scenario

fnlctrl: Maybe in that case, the user should provide a function and convert types by himself. This way we avoid the implicit behavior.

src/components/view.js
bfanger: Did some further testing and setting properties on `<router-view>` overwrites the values set in data.props, so they are merged somewhere

src/util/props.js
bfanger: Agreed, i've inlined the hasProp function

src/util/props.js
bfanger: Agreed, done

src/components/view.js
LinusBorg: Yep, that's normal Vue behaviour for any component.

src/util/props.js
bfanger: I've simplified the behavior of the {props: true} which now sets the props to route.params.
Vue complaining about type is much better than silently ignoring some properties and leaving users confused why it doesn't work.

posva:
I love the idea!
Haven't read all the code yet

LinusBorg:
I think this is a really great idea. 

/ping @yyx990803 you should check this out, seems very useful!

yyx990803:
Looks good! Will review in depth after new year and likely merge when we work on 2.2 for the router.

test/e2e/nightwatch.config.js
```diff
@@ -1,4 +1,7 @@
 // http://nightwatchjs.org/guide#settings-file
+var seleniumServer = require('selenium-server');
+var chromedriver = require('chromedriver');
```
posva: This line is not necessary

test/e2e/nightwatch.config.js
```diff
@@ -7,7 +10,7 @@ module.exports = {
 
   'selenium': {
     'start_process': true,
-    'server_path': 'node_modules/selenium-server/lib/runner/selenium-server-standalone-2.53.1.jar',
+    'server_path': seleniumServer.path,
```
posva: To be consistent we should do `server_path': require('selenium-server').path` instead

docs/en/essentials/passing-props.md
```diff
@@ -0,0 +1,71 @@
...
+
+This allows you to use the component anywhere, which makes the component easier to reuse and test.
+
+### Boolean mode
+
+When props is set to true, the route.params will be set as the component props.
+
+### Object mode
+
+When props is an object, this will be set as the component props as-is.
```
posva, line 42: It'd be even better to say when it's useful to pass an object as the props

posva (changes requested):

test/e2e/nightwatch.config.js
bfanger: done

test/e2e/nightwatch.config.js
bfanger: lines are removed

docs/en/essentials/passing-props.md
bfanger: I've improved the description and code snippet.

posva (approved):

yyx990803:
@bfanger apologies for taking so long to merge this - great work!

docs/en/api/options.md
```diff
@@ -13,6 +13,7 @@
     name?: string; // for named routes
     components?: { [name: string]: Component }; // for named views
     redirect?: string | Location | Function;
+    props?: boolean | string | Function;
```
baoti, line 16: It is `Boolean` | `Object` | `Function` in passing-props.md

PR #979, opened 2016-12-03
Added VueRouter.version

ytbryan:


posva:
files in the dist directory are generated files, nothing should be added there. Also, it shouldn't start with a `v`
@fnlctrl Adding the version looks legit but we should add it to vuex as well to be consistent. WDYT?
@yyx990803 Shouldn't this https://github.com/vuejs/vue/blob/dev/src/core/index.js#L11 be added with a replace plugin, using something like `__VERSION__` and then, in the rollup file, read the version from the package? This'd only work if the package is updated **before** bundling.

fnlctrl:
Yeah, maybe we should add versions to vuex/vue-router..

yyx990803:
@posva yes - both suggestions make sense. PR welcome.

posva:
@yyx990803 Ok, will do

posva:
@yyx990803 Actually, there's already a manual replacement done with a regex in vue to support another version for weex aswell.
I was thinking about using https://github.com/rollup/rollup-plugin-replace. I'll do so for vue-router and vuex.

posva:
Thanks for the suggestion @ytbryan 😉 
Closing this in favor of #991

PR #996, opened 2016-12-07
Add contributions templates

posva:
@fnlctrl I don't know what the Chinese text says 😛 I just took the vue issue template and replaced the jsfiddle links. I still think that the issue template is too long and I fear people will just ignore it and we end up not gaining any time.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,48 @@
+<!--
+中文用户请注意：
+
+1. issue 只接受带重现的 bug 报告，请不要用来提问题！不符合要求的 issue 会被直接关闭。
+2. 请尽量用英文描述你的 issue，这样能够让尽可能多的人帮到你。
+
+Got a question?
+===============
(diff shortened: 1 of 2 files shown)


fnlctrl:
Yeah, maybe we should make it shorter..

About the Chinese text (It's actually concise and straight to the point 😂 , though not very friendly):
```
中文用户请注意：
Attention Chinese users:

1. issue 只接受带重现的 bug 报告，请不要用来提问题！不符合要求的 issue 会被直接关闭。
The issue list only accepts bug reports with reproductions, please do not ask questions here! Issues that fails to comply will be immediately closed.

2. 请尽量用英文描述你的 issue，这样能够让尽可能多的人帮到你。
Please try to describe your issue in English, so that more people can help.
```

fnlctrl (approved):

posva:
I'll think of something this afternoon. All of the points are important imo but the real deal is to prevent useless issues to appear and in order to do that we have to find the right balance so people really take the time to read the template

posva:
@fnlctrl After thinking about it, I don't think it's necessary to shrink the message. It's fine like this. We can start with this anyway and improve progressively. Feel free to merge

PR #1020, opened 2016-12-14
 decode URI when aborting transition (Fix #760)

### LINKED ISSUE #760: URL parameters are re-encoded when aborting a transition
If any parameter contains a URL encoded value, the value is re-encoded when the transition is aborted.

If the current URL is http://website.com/#!/accounts?PARAM=%20, calling `transition#abort` redirects to http://website.com/#!/accounts?PARAM=%25%20 instead of the original URL.

Checking `transition#abort`, it seems the call to `#stringifyPath` performs the double encoding when a `String` parameter is passed.

LinusBorg:
(Fixes #760)

### Problem

When aborting a transition, and the URL contained escaped chars, the chars would be escaped again. I added a decode in between to make this work again.

### DIFF
--- a/src/transition.js
+++ b/src/transition.js
@@ -46,6 +46,8 @@ export default class RouteTransition {
       // on initial load, it gets caught in an infinite loop.
       const abortingOnLoad = !this.from.path && this.to.path === '/'
       if (!abortingOnLoad) {
+        // When aborting, we have to decode the Path, because router.replace()
+        // will encode it again. (https://github.com/vuejs/vue-router/issues/760)
         const path = this.from.path ? tryDecode(this.from.path, true) : '/'
         this.router.replace(path)
       }
--- a/src/transition.js
+++ b/src/transition.js
@@ -13,6 +13,7 @@ import {
   canReuse
 } from './pipeline'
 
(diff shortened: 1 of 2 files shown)


dfmelicor (approved):

BreakingDev:
Any idea when this will be merged? @dfmelicor

src/transition.js
```diff
@@ -45,7 +46,8 @@ export default class RouteTransition {
       // on initial load, it gets caught in an infinite loop.
       const abortingOnLoad = !this.from.path && this.to.path === '/'
       if (!abortingOnLoad) {
-        this.router.replace(this.from.path || '/')
+        const path = this.from.path ? tryDecode(this.from.path, true) : '/'
```
posva, line 51: Can you add a little comment over here to point out why we need to call `tryDecode`?

LinusBorg: sure!

LinusBorg: done.

posva (approved):

fnlctrl (approved):

znck (approved):

kazupon (approved):

Issue #938, opened 2016-11-25
Suggestion for guard: beforeRouteUpdate


wolthers:
Consider this setup:

    routes: [{
      path: 'profile/:profileId',
      component: ProfileComponent,
      beforeEnter (to, from, next) {
        //Fired the first time the profile route is matched
      },
    }]

The first time the route is matched, the `beforeEnter` callback is executed which allows me to implement any guards I want. This is however not the case if I navigate from `profile/user1` to `profile/user2`. This works as intended and makes sense to me because the route isn't "entered" again. However, I see a valid use case for implementing a guard in this scenario and by reading closed issues, the only suggestion given before closing is to create a watcher for $route. This feels wrong for several reasons:

1. There's no way to create a watcher for per-route beforeEnter configuration in the VueRouter constructor options.
2. If implemented on the component, the component now knows about application route structure. This might be ok in many scenarios but I don't think it should be forced.

I suggest creating a `beforeRouteUpdate` callback which works exactly the same way as `beforeRouteEnter`, but only fires when a route param like `:profileId` changes.

fnlctrl:
We've had some discussion in the team and there are two possible options, the first is what you proposed, the second is to let `beforeEnter` fire on params/query/hash change. I guess we'll wait for Evan to make the call, but it's good to start a discussion here.

smallpath:
Hi, I think the global guard `beforeEach` can help, because it will be triggerred even only `params` or `query` changes. And it's pretty simple to use `beforeEach` especially when you have `preFetch` hooks in SSR. 

I have completed a loading bar at [my blog](https://smallpath.me/) using `beforeEach`, you can see it by navigating to any other route. [This link](https://github.com/Smallpath/Blog/blob/develop/client/front/src/client-entry.js#L20) is how I implemented it using `beforeEach`. Hope it helps you

By the way, I found that `watch` is not great when the first time I upgrade to vue2.0, since `watch` can listen to `params` and `query` changes but can't prevent the route changes because it doesn't have `next` method which all before guards have. I think a `beforeRouteUpdate` hook makes since to replace `beforeEach` when your web application is very large

fnlctrl:
@wolthers @Smallpath Would you think it's rational if `beforeEnter`/`beforeRouteEnter` is called on params/query/hash change? This way we avoid adding a new hook.

smallpath:
@fnlctrl Yes, I think it's better than making a new hook

wolthers:
[...]

PR #1026, opened 2016-12-19
Fixes #1012 same route guard execution

### LINKED ISSUE #1012: Route push to same named route with different params will not execute beforeRoute guards
**NOTE:** I cannot create a fiddler right now. I will update the ticket later, when I can, with a fiddler. However, I did link to the source that was the problem. I verified this in my browsers debug tools (chrome). 

2.0.7 / 2.0.2

### Steps to reproduce
1. Define a route that has a beforeEnter guard (See below for my use case)
2. Call `router.push()` from that route to the same route but with different parameters (i.e. `/requests/1` to `/requests/2`)

### What is Expected?
The route transition is executed AND the beforeEnter guard is executed. 

Essentially I have a route that is defined as such:

```javascript
{
    path: '/request/:id',
    name: 'showRequest',
    component: ShowRequest,
    beforeEnter: (to, from, next) => {
        store.dispatch('fetchRequest', to.params.id)
        next()
    }
}
```

Inside my `Vue` I have a method that performs the following:

```javascript 
api.clone(this.request, function(response) {
     router.push({ name: 'showRequest', params: { id: response.id } })
})
```

The API call is returning a `cloned` object but with a unique ID. I wish to transition to the `showRequest` route but with the different parameter. What happens is that the route (url in my browser) is set, but the beforeEnter guard is not executed. Therefore, the URL is updated but the `fetchRequest` `Vuex` action is never called. 

I ended up attaching a debugger to understand what is going on. The first thing I verified was that the `isSameRoute` was returning `false` [(see isSameRoute)](https://github.com/vuejs/vue-router/blob/dev/src/util/route.js#L45). Since the [paths are not equal](https://github.com/vuejs/vue-router/blob/dev/src/util/route.js#L53) (different ID at the end), this function returns false. 

The next thing to happen is that the router is attempting to [resolveQueue](https://github.com/vuejs/vue-router/blob/dev/src/history/base.js#L139). This is where things break down. The two parameters passed in, `this.current.matched` and `route.matched` are equal. Therefore the `activated` and `deactivated` arrays are empty. Therefore the `beforeEnter` guard I have is never executed. 

The problem is the comparison [being done between the two routes ](https://github.com/vuejs/vue-router/blob/dev/src/history/base.js#L149). Slicing an array at its final index will undoubtedly return an empty array. This is precisely what is happening. Some sort of different comparison or override is needed.

predhme:
#1012 - Removing the `resolveQueue` function and simply using the `route.matched` and `this.current.matched` as the means for executing the guards resolves this issue.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -49,10 +49,8 @@ export class History {
       return abort()
     }
 
-    const {
-      deactivated,
-      activated
-    } = resolveQueue(this.current.matched, route.matched)
+    const activated = route.matched
+    const deactivated = this.current.matched
 
     const queue: Array<?NavigationGuard> = [].concat(
       // in-component leave guards
@@ -134,26 +132,6 @@ function normalizeBase (base: ?string): string {
   }
   // remove trailing slash
   return base.replace(/\/$/, '')
-}
-
-function resolveQueue (
-  current: Array<RouteRecord>,
-  next: Array<RouteRecord>
-): {
-  activated: Array<RouteRecord>,
-  deactivated: Array<RouteRecord>
-} {
-  let i
-  const max = Math.max(current.length, next.length)
-  for (i = 0; i < max; i++) {
-    if (current[i] !== next[i]) {
-      break
-    }
-  }
-  return {
-    activated: next.slice(i),
-    deactivated: current.slice(i)
-  }
 }
 
 function extractGuard (


yyx990803:
Thanks for the PR - however, the current behavior is by design and this PR would introduce a breaking change. We will probably go with something similar to #938 instead.

predhme:
@yyx990803 Completely understandable. Thanks!

PR #1043, opened 2016-12-22
location.replace and assign order

dolymood:
Should called location.replace when replace is true


posva (approved):
@yyx990803 @fnlctrl I guess this wasn't done on purpose 😆 . If it was, it deserves a comment

PR #1045, opened 2016-12-22
Add base in abstract mode (fix #983)

### LINKED ISSUE #983: Router base in abtract mode (SSR)
When setting a base in the router configuration, on the server-side rendering, `router.history.base` is an empty string.

It think it's in the abstract constructor that does have the base in argument: https://github.com/vuejs/vue-router/blob/dev/src/history/abstract.js#L10

atinux:
This PR fix the issue #983 to make it work with SSR.

Ping @LinusBorg

### DIFF
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -7,8 +7,8 @@ export class AbstractHistory extends History {
   index: number;
   stack: Array<Route>;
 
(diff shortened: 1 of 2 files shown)


atinux:
Any idea when the new version with this fix will be released? @yyx990803

PR #1047, opened 2016-12-23
Custom encodequery

### LINKED ISSUE #1027: Commas don't need to be encoded in query component
2.1.5 / 2.1.1

### Steps to reproduce
```
router.push({path: 'foo', query: {
  bar: '1,2'
}})
```

### What is happening and what is expected?
Currently the router will redirect to `foo?bar=1%2C2` which is fine and works. Using comma as a sub-delimiter is very common these days and it's much nicer to see `foo?bar=1,2`

According to http://stackoverflow.com/questions/2366260/whats-valid-and-whats-not-in-a-uri-query rfc3986 allows unencoded commas as sub-delimiters in the query component.

I think the default could/should stay as encoding the full query string but an option to allow unencoded commas (or whatever the sub-delimiter is) would be really useful.

posva:
Let me know if this implementation and the api are good enough so I can update the docs too

---
Closes #1027
Pass a function to use it instead of encodeURIComponent or pass false to
pass query arguments without encoding them. This would allow you to
encode only some characters doing something like
`encodeURIComponent(queryValue).replace('%2C', ',')`
would keep the commas on your query arguments.
⚠️ Setting it to false should only be used if you have other kinds of
sanitizations going on or if you're sure the query variables are URI
safe.

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -14,11 +14,13 @@ const Bar = { template: '<div>bar</div>' }
 // 3. Create the router
 const router = new VueRouter({
   mode: 'history',
+  encodeQuery: false,
   base: __dirname,
   routes: [
     { path: '/', component: Home },
     { path: '/foo', component: Foo },
-    { path: '/bar', component: Bar }
+    { path: '/bar', component: Bar },
+    { path: '/query', component: Bar }
   ]
 })
 
@@ -37,6 +39,7 @@ new Vue({
         <router-link tag="li" to="/bar" :event="['mousedown', 'touchstart']">
           <a>/bar</a>
         </router-link>
+        <li><router-link to="/query?opts=a,b,c">/query?opts=a,b,c</router-link></li>
       </ul>
       <router-view class="view"></router-view>
     </div>
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -57,6 +57,7 @@ declare type Location = {
   path?: string;
   hash?: string;
   query?: Dictionary<string>;
+  encodeQuery?: Function | boolean;
   params?: Dictionary<string>;
   append?: boolean;
   replace?: boolean;
--- a/src/index.js
+++ b/src/index.js
@@ -138,6 +138,9 @@ export default class VueRouter {
     href: string
   } {
     const normalizedTo = normalizeLocation(to, current || this.history.current, append)
+    normalizedTo.encodeQuery = normalizedTo.encodeQuery === undefined
+      ? this.options.encodeQuery
+      : normalizedTo.encodeQuery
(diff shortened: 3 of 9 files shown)


posva:
@yyx990803 @fnlctrl I forgot about this PR. The idea is to allow the user to bypass the aggressive `encodeURIComponent` replacement eg: not replacing `,` in `?foo=foo,bar` query. The work on the PR is not finished yet as the `encodeQuery` property is not read from the `Location` object yet. But I wanted to draft something before going on with a full implementation

posva:
I added the missing part and tested further:

- `/?query=e,b`
- `/?query=e%2Cb`
- `{ path: '/', query: { query: 'e,b' }, encodeQuery: false }`

Are all considered the same route -> if the user is on any of these routes trying to go to another, nothing won't happen.
I think this behaviour is ok (first 2 routes were already considered the same before) because if you want to encode a query, you do it wherever you the same route is used. Which leads me to another question, should we support a global `encodeQuery`? Personally, I think we shouldn't because it isn't safe to use it globally. It could be enabled globally with a custom router-link:

```js
{
  functional: true,
  render (h, context) {
    const to = context.data.attrs.to
    context.data.attrs.to = typeof to === 'string'
                            ? { path: to, encodeQuery: false }
                            : {...to, encodeQuery: false }
     return h('router-link', context.data, context.children)
  }
}
```

yyx990803:
Without a global option the usage potentially feels quite verbose. What is the safety concern that you have here? I mean if a user decide to use this option then it is up to the user to implement the custom encoding function correctly.

posva:
I was concerned about the user not implementing the encoding function correctly and accidently sending wrong requests.
It makes sense being able to overwrite this encodeQuery function globally, eg when instantiating the router. My concern was that you probably don't need it in every route and that leaving the default behaviour is safer

posva:
So, let's add it to the router constructor with a default `true` value and let it be overridden by Locations.
Also, should we prohibit setting this option to false in the constructor? The user could just use a custom function to bypass the error but if someone does intentionally, they'll get an error

edit: ~~It could also be defined at a route level, I may accidently added this since the normalize is called in createMatcher~~ It's on the match function, not the createMatcher

I'll do this properly tomorrow with a clearer mind

yyx990803:
After thinking about this, it seems the comma is the only special case here (encoded by `encodeURIComponent` but not actually reserved according to spec). In addition there's also a number of chars `*()!'` that are not encoded by `encodeURIComponent` but are actually reserved in the spec. I updated the encode function to take both into account, and I think that's better than introducing an extra option for this edge use case.

posva:
Ok. I really thought there were more characters. That spec looks so confusing to me I didn't understand it correctly...

PR #1053, opened 2016-12-27
Update named-views.md

anishkny:
Demo has superfluous <div id="app">


anishkny:
Opening issue instead of PR, because jsfiddle needs to be updated.

PR #1065, opened 2016-12-30
Add bower.json

dschulz:
I think this is required to be able to deploy to webjars.org as a Bower package.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "vue-router",
   "main": "dist/vue-router.js",
-  "description": "the official router for Vue.js",
+  "description": "official router for Vue.js",
   "authors": "Evan You",
   "license": "MIT",
   "ignore": [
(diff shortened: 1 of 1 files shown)


rpkilby:
Not sure if this would be merged. See https://github.com/vuejs/vue/issues/4040

dschulz:
Thanks for the feedback @rpkilby. Didn't know about this, I just assumed the file was missing in this particular repository because I saw there are other projects under vuejs/ that do have a bower.json file. For example [vuejs/vuex/blob/dev/bower.json](https://github.com/vuejs/vuex/blob/dev/bower.json).

posva:
ping @fnlctrl @yyx990803

fnlctrl (approved):

posva:
@rpkilby I think it's ok to have a simple `bower.json` file contributed that doesn't need any (or almost any) maintenance. We still recommend using `npm` (or yarn) 😁

kazupon (approved):

PR #1067, opened 2017-01-01
Clarify where to use In-Component guards

posva:
Feedback from
https://forum.vuejs.org/t/in-component-guard-beforerouteleave-not-working-solved/4658/4


kazupon (approved):
LGTM :)

PR #1084, opened 2017-01-09
🇩🇪 Reference dynamic route matching in navigation guards

PascalAOMS:


posva (approved):
I don't speak German, but Thanks a lot! 👍

PR #1101, opened 2017-01-17
Fix active route matching

### LINKED ISSUE #1091: Different route define, but the same suffix path, will all be actived!
`vue-router`'s version:  *3.10.9*
```
// routes.js
const routes = [
   {name: 'tag', path:'tag'},
   {name: 'tagtype', path:'tagtype'}
]

// pug's template
ul
   li(v-for="item in items")
      router-link(active-class="active" v-bind:to="item.to") 

// data
[
  {title: 'tagtype',  to: {name: 'tagtype'}},
  {title: 'tag', to: {name: 'tag'}},
]
```
The url path `/tag` '/tagype' will all be actived!
If change `tagtype` with a different suffix, like `atagtype`, this bug will be fixed!

posva:
Fix #1091
Fix the doc too

### DIFF
--- a/docs/en/api/router-link.md
+++ b/docs/en/api/router-link.md
@@ -94,7 +94,7 @@
 
   - default: `false`
 
-  The default active class matching behavior is **inclusive match**. For example, `<router-link to="/a">` will get this class applied as long as the current path starts with `/a`.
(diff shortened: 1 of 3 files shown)


posva:
Something I just realised, if we add this it'll no longer be possible to set as active a link to `something:foo` if the current route is `something`

yyx990803:
What do you mean?

posva:
Nevermind. It's better to have this behaviour as default anyway

PR #1105, opened 2017-01-19
Check for performance.now support

### LINKED ISSUE #1104: vue-router 2.1.2 has broken in IE 9
2.1.10 / 2.1.2

### Steps to reproduce

In IE 9 (IE 11 with document mode 9), there are two error in console. One is vue, another is vue-router.

About vue-router, console show `Object doesn't support this property or method (now)`.

对象不支持“now”属性或方法

I have checkout the vue-router source code, and find the problem.

https://github.com/vuejs/vue-router/commit/1961d3511dd286ae3266a111019ec77ed5b07b29#diff-6f3e0f40e05988a09d760ec1a86f3420

On line 17, `inBrowser` is true, but `window.performance` does not have `now` method in IE 9.

It is a mistake.

posva:
Fix #1104

IE 9 supports the performance API but doesn't have the now method. Since we're relying specifically on that one, better directly check for performance.now

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -15,7 +15,7 @@ import {
 
 // use User Timing api (if present) for more accurate key precision
 // IE 9 supports performance API but doesn't have performance.now
-const Time = inBrowser ? (window.performance.now ? window.performance : Date) : Date
+const Time = inBrowser ? (window.performance && window.performance.now ? window.performance : Date) : Date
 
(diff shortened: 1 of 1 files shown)


yyx990803:
Oops, didn't see this PR when I pushed the fix... sorry!

posva:
No worries 🙂

PR #1108, opened 2017-01-20
1102: $route returns consistent value across all Vue instances

jhartman86:
- Allow injecting router to discrete Vue instances and ensure the $route (getter) returns consisten value

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -15,6 +15,7 @@ export default class VueRouter {
   static install: () => void;
   static version: string;
 
+  apps: Array<?any>;
   app: any;
   options: RouterOptions;
   mode: string;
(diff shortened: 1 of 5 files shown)


jhartman86:
Maintainers - not expecting for this PR to get accepted right away as I'm guessing it violates the coding standards somewhere. If feedback can be provided on what needs to be fixed for it to be accepted it'd be great.

Note - the simplest fix was to keep track of all the app instances and just set the private _route property on all of them. That effectively accomplishes keeping the $route value in sync across all instances.

src/index.js
```diff
@@ -89,7 +92,11 @@ export default class VueRouter {
     }
 
     history.listen(route => {
-      this.app._route = route
+      this.apps.forEach((app) => {
+        if (app) {
```
posva: A check shouldn't be necessary, should it?

jhartman86: Thats what I thought as well, but the syntax check wouldn't let it pass. Don't remember exactly, but it was along the lines of "Cannot set property on a possibly null value".

PR #1128, opened 2017-01-30
typo: router.addRoutes(routes)

laurentpayot:
`router.addRoutes(routes)`, not `router.addRoute(routes)`


posva:
Nice catch! Thanks!

PR #1137, opened 2017-02-05
Fix route hash docs

### LINKED ISSUE #1135: $route.hash includes '#'
The documentation, https://router.vuejs.org/en/api/route-object.html,  says that $route.hash is "The hash of the current route (without #)". As of 2.2.0 it _does_ include the "#".  Does the documentation need to be updated or do we need to add the feature?

posva:
Closes #1135

I couldn't update the other languages. I'm confident about the hash being enforced because of https://github.com/vuejs/vue-router/blob/dev/src/util/location.js#L43-L45
But wanted to ping you first @fnlctrl before reaching out for translation help

### DIFF
--- a/docs/en/api/route-object.md
+++ b/docs/en/api/route-object.md
@@ -54,7 +54,7 @@ The route object can be found in multiple places:
 
   - type: `string`
 
-    The hash of the current route (without `#`), if it has one. If no hash is present the value will be an empty string.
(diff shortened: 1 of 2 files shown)


kazupon (approved):

posva:
Note to myself: merge #1138 when this is merged

fnlctrl (approved):

fnlctrl:
Fixed zh-cn docs

posva:
Waiting for CI to pass

PR #1143, opened 2017-02-07
[Feature] add param cast option

blake-newman:
#Add param casting options. It is quite often that you wish to cast the params to a number or for custom decoding options.

This is achievable currently through custom prop option as function, however this is limited to the scope of the prop feature. Use of `this.$route.params` or `this.$store.route.params` will always be of type string which means often casting around the application.

This feature allows casting at a single point (the definition)

```js
{
  path: '/number/:id',
  params: {
    id: Number, 

    // or custom cast function
    id: value => Number(value) + 1
  }
}
```

If this approved i can commit changes to the documentation, to correct current incorrect information (https://github.com/vuejs/vue-router/blob/master/docs/en/route.md#dynamic-segments), and to document new feature.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -15,6 +15,7 @@
     <li><a href="route-matching">Route Matching</a></li>
     <li><a href="active-links">Active Links</a></li>
     <li><a href="redirect">Redirect</a></li>
+    <li><a href="route-params-cast">Route Parameter Casting</a></li>
     <li><a href="route-props">Route Props</a></li>
     <li><a href="route-alias">Route Alias</a></li>
     <li><a href="transitions">Transitions</a></li>
--- a/examples/route-params-cast/Hello.vue
+++ b/examples/route-params-cast/Hello.vue
@@ -1,0 +1,20 @@
+<template>
+  <div>
+    <h2 class="hello">Hello {{ name }}</h2>
+    <p class="type">this.name is a: <b>{{ type }}<b></p>
+  </div>
+</template>
+
+<script>
+
+export default {
+  computed: {
+    name () {
+      return this.$route.params.name || 'Vue!'
+    },
+    type () {
+      return typeof this.name
+    }
+  }
+}
+</script>
--- a/examples/route-params-cast/app.js
+++ b/examples/route-params-cast/app.js
@@ -1,0 +1,32 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+import Hello from './Hello.vue'
+
+Vue.use(VueRouter)
+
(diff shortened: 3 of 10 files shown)


LinusBorg:
I like this functionality, but I'm not sure about the name of the key `params:`. itS pretty generic  and sounds like it *defines* the params while it really typecasts them, and you don't have to specify all of your params there, only the ones you want to typecast.

I think especially with the[ `props:` feature](http://router.vuejs.org/en/essentials/passing-props.html) introduced in 2.2, this can lead to confusion for someone just reading the code, because both deal with params in a way.

I don't have a better name from the top of my head, but maybe we can come up with something more descriptive?

posva:
I also think it's confusing.
It'd be better to use the function-mode for props: https://router.vuejs.org/en/essentials/passing-props.html#function-mode

I thought about using an dict of functions but that would prevent from passing prop functions to views (a breaking change)

LinusBorg:
@posva function mode for props would only typycast props, not `$route.params`

posva:
@LinusBorg Can't you simply cast whatever you want from the route?

blake-newman:
@posva @linusborg I too was questioning the naming. `cast` is probably better naming. I also realised that query parameters probably need to accept this feature. Thus another object option name is needed.

@posva if you type cast the route, you probably won't need the function type for prop option. Unless you are doing computation of a route param.  




@yyx990803 the master of API naming what's your thoughts?

posva:
But you can already cast using a function in the props: http://jsfiddle.net/posva/0937bqx9/

blake-newman:
@posva yes, cast to props only which doesn't affect params/query values in store or route object. 

With a SSR application the store is passed to the preFetch functions which are non casted values. So you have to recast, and alot anywhere else you use the store or $route object. 

If a cast option existed then all places would be precasted. Meaning the props object can just be of the value true in most places. So no overhead there. 

IMO the props option shouldn't be the truth of casting due to its flexibility. If the props option where to change what's stored in the store then it would reduce the versatility of the feature.

posva:
@blake-newman Ok, I see. I'm an SSR noob 😅

blake-newman:
@posva @LinusBorg See slack, i completely forgot about this PR.

Under further consideration, i think we should just provide props on the route object. This way you can invoke it manually when required. As per requirements above. This way we can do something like: 

```js
{
  async fetchData(store, to, from) {
    const props = to.props.default() // or to.props.default(to) for simple integration
    store.dispatch('a/action', props)
  },
}
```

Overall the introduction will remove repeated code

PR #1147, opened 2017-02-09
[Feature] resolve props for router.getMatchedComponents

blake-newman:
The new props feature allows to decouple components. However when using SSR you normally do preFetch the component data. This means we are once again coupled between the router and the component. 

Current scenario:
```js
  function preFetch(store) {
    const id = store.route.params.id;
    // Do something with id
  }

  export default {
    preFetch,
    beforeMount() {
      preFetch(this.$store);
    },
  };
```

By resolving the props to `$props` on components retrieved from `router.getMatchedComponents` we are able to use the props feature to keep components and the router decoupled with SSR.

Decoupled component:
```js
  function preFetch(store, props) {
    const id = props.id;
    // Do something with id
  }

  export default {
    name: 'IndexPage',

    props: {
      test: String,
    },

    preFetch,
    beforeMount() {
      preFetch(this.$store, this.$options.propsData);
    },
  };
```

preFetch: 
```js
  await Promise.all(
    entry.router.getMatchedComponents()
      .filter(component => component.preFetch && component.preFetch)
      .map(component => component.preFetch(entry.store, component.$props)),
  );
```

### DIFF
--- a/test/unit/specs/props.spec.js
+++ b/test/unit/specs/props.spec.js
@@ -1,0 +1,40 @@
+import { resolveProps } from '../../../src/util/props'
+
+describe('Props utils', () => {
+  describe('resolveProps', () => {
+    beforeAll(() => {
+      spyOn(console, 'warn')
+    })
+
+    it('undefined', () => {
+      const result = resolveProps()
+      expect(result).toBeUndefined()
+    })
+
+    it('object', () => {
+      const result = resolveProps(null, { test: 'yes' })
+      expect(result).toEqual({ test: 'yes' })
+    })
+
+    it('function', () => {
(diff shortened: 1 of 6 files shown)


test/unit/specs/props.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      const result = resolveProps()
+      expect(result).toBeUndefined()
+    })
+
+    it('object', () => {
+      const result = resolveProps(null, { test: 'yes' })
+      expect(result).toEqual({ test: 'yes' })
+    })
+
+    it('object', () => {
```
posva: you duplicated that test 😄

blake-newman: Haha, oops.

PR #1157, opened 2017-02-11
[DOCS] Improve german translation

LinusBorg:
I have revamped the german translation as I found that it had quite a few paragraphs that were translated too "verbatim" and didn't read well to german eyes.

Also corrected some mistakes.


posva:
Cool! Thanks a lot. Do we need to wait for some changes, or is it ready to merge?

LinusBorg:
Should be good to go. I will inevitably find some typos once it's live, but that's normal I guess :D

Will push fixes directly to `dev` for those.

PR #1159, opened 2017-02-12
Call next() only after user-defined callback

tjk:
I'm not familiar with the codebase so this could be completely wrong, but I'm seeing issues when calling `next` in a `beforeRouteEnter`. I expect the callback passed into it to execute before it proceeds. ~~I will work on a minimal repro to show off what I mean but hopefully that is clear.~~

Here is a minimal repro: http://jsfiddle.net/9ututkjp/2/. Let me know what you think!

(Depending on whether I'm making the right assumption and this approach is correct, I'm happy to *try* and add a test if this is the desired behavior).

Edit: updated repro link.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -246,7 +246,6 @@ function bindEnterGuard (
 ): NavigationGuard {
   return function routeEnterGuard (to, from, next) {
     return guard(to, from, cb => {
-      next(cb)
       if (typeof cb === 'function') {
         cbs.push(() => {
           // #750
@@ -254,8 +253,10 @@ function bindEnterGuard (
           // the instance may not have been registered at this time.
           // we will need to poll for registration until current route
           // is no longer valid.
-          poll(cb, match.instances, key, isValid)
+          poll(cb, match.instances, key, isValid, next)
         })
+      } else {
+        next(cb);
       }
     })
   }
@@ -265,13 +266,15 @@ function poll (
   cb: any, // somehow flow cannot infer this is a function
   instances: Object,
   key: string,
-  isValid: () => boolean
+  isValid: () => boolean,
+  next: (Vue) => void
 ) {
   if (instances[key]) {
     cb(instances[key])
+    next()
   } else if (isValid()) {
     setTimeout(() => {
-      poll(cb, instances, key, isValid)
+      poll(cb, instances, key, isValid, next)
     }, 16)
   }
 }
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -267,7 +267,7 @@ function poll (
   instances: Object,
   key: string,
   isValid: () => boolean,
-  next: (vm: Vue) => void
+  next: (vm: _Vue) => void
 ) {
   if (instances[key]) {
     cb(instances[key])
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -256,7 +256,7 @@ function bindEnterGuard (
           poll(cb, match.instances, key, isValid, next)
         })
       } else {
-        next(cb);
+        next(cb)
       }
     })
   }
@@ -267,7 +267,7 @@ function poll (
   instances: Object,
   key: string,
   isValid: () => boolean,
-  next: (Vue) => void
+  next: (vm: Vue) => void
 ) {
   if (instances[key]) {
     cb(instances[key])


tjk:
Hm, I'm running into some issues around this part of the code so I'm still trying to grok it... looking forward to some feedback, though!

src/history/base.js
```diff
@@ -265,13 +266,15 @@ function poll (
   cb: any, // somehow flow cannot infer this is a function
   instances: Object,
   key: string,
-  isValid: () => boolean
+  isValid: () => boolean,
+  next: (vm: _Vue) => void
 ) {
   if (instances[key]) {
     cb(instances[key])
+    next()
```
tjk, line 274: For parity, should I call `next(cb)` here?

simplesmiler:
A few notes:
1. In your repro the `*` router is placed before the `/repro`, so `/repro` is never reached.
2. The function passed to `next` won't be executed immediately. In fact, this function will be called *after* the navigation was confirmed, because `vm` will not be constructed until navigation is confirmed.

posva:
It's quite strange tests passed on CI, they didn't pass locally for the callback in a `next` in a `beforeRouteEnter`

tjk:
@simplesmiler thanks!

1. I updated the repro link (swapping the routes might have happened in a later revision... because I know I saw the console errors by clicking on the repro route at some point -- this is fixed now).
2. Are you saying this patch is the wrong approach then? As I've been trying to trace through vue-router code more, I notice `History.prototype.confirmTransition` calls `onComplete` after "checking" all the `enterGuards`, then should run the `postEnterCbs` (which would poll... for a reason I don't understand yet) and only then proceed from the guard after running the user-defined callback.

After some more digging in this pretty tricky code I feel like maybe my assumption that a user-defined callback should run before proceeding from the `beforeRouteEnter` may be wrong. Is that correct? In other words, could you tell me how I should alter my repro code?

tjk:
@posva I wasn't able to get the tests to run locally yet but when I saw them pass on the CI server I punted on getting my test set up locally. I'm betting I'm just making a poor assumption so I'm hoping someone can point me in the right direction (ie. how would you correct my "repro"). Cheers.

tjk:
I've unpatched my vue-router and added `null` checks in the Vue instance methods, but I'm curious if that is the proper approach? Any way to avoid a render without data (which is all I'm trying to accomplish, really)? (I'm likely doing something wrong)

posva:
@tjk What are you trying to achieve? Get the next callback called before the route component mounted hook?

tjk:
@posva Exactly. What am I doing wrong? 😄 

Edit: Or rather (to be clear), I want vue-router to call the function I pass to the `next` callback (the one being called with the vue model as first argument) to run before render (or calls to computed methods, etc.).

yyx990803:
Unfortunately, the current behavior is by design and the proposed change will not work (it's also a breaking change).

`next` needs to be called in order to confirm the navigation -> the component will only be instantiated after the navigation is confirmed -> only then can the `vm` instance be passed to the callback. So logically this simply won't work.

tjk:
@yyx990803 I'm fine with this being closed, but could you please tell me if simply `null` checking before accessing state objects that we instantiate in this manner is the right approach.

yyx990803:
I'm afraid I don't understand your question...

tjk:
@yyx990803 How would you make this jsfiddle (http://jsfiddle.net/9ututkjp/2/) work when you click the "bug?" link? Make the template `{{ person && person.name }}`?

LinusBorg:
For this concrete example: https://jsfiddle.net/Linusborg/9ututkjp/3/

Usually you would wather have the condition on an element in a `v-if` directive

tjk:
Hm that's unfortunate. Unless I'm misunderstanding, a loading view would be required so the UI doesn't flicker if you want to asynchronously load something in `beforeRouteEnter`.

Thanks for everyone's help.

yyx990803:
The UI won't change until `next` is called in `beforeRouteEnter`. This hooks was introduced precisely to avoid having to provide a loading view.

tjk:
The whole point of this is that if I want to attach something to the vue instance state, I need to do it via the function passed as first arg to `beforeRouteEnter`'s `next` (because it is yielded the vue instance `vm`, not accessible otherwise in `beforeRouteEnter`). But the UI changes before that function is called (see `next(cb)` in my bad patch)... before the state of the vue instance has been updated (hope that is clear).

yyx990803:
@tjk ah I get it now - for you original use case, a null check like `{{ person && person.name }}` would be necessary. Or you wrap the entire block that depends on person in a `if` block. Now I do understand why there may be a flicker when you do this - I think this is something we can try to improve.

tjk:
@yyx990803 Thanks! (Another option would be to attach the state to a global store via some Flux-like architecture like Vuex before calling `beforeRouteEnter`'s `next` but I think it would be good to aim to improve this to allow the state to stay "local" to the vue instance if possible).

PS: I absolutely love Vue and it's ecosystem. Thanks for all the great work you do!

Issue #1170, opened 2017-02-16
Vue-router can not work properly in the typescript2.0 environment


ulivz:
I use vue2.0 and typescript2.0, as well as vue-router2.0 as my development technology stack, which is my project template —— [template](https://github.com/toxichl/vue-typescript-template-ui)，You run this project (`npm run dev`), you will find the error as follows:

`vue-typescript-template-ui\node_modules\vue-router\types\router.d.ts`
`(2,10): error TS2305: Module '"vue"' has no exported member 'ComponentOptions'.`
`.......`

Does it have a problem with the type declaration file of vue-router2.0 ？

posva:
@ktsn @kazupon Can you chime in, please 😁?

ktsn:
I guess `@types/vue-resource` still use Vue v1.x typings, so it conflict with Vue v2.x typings. 😞

posva:
@ktsn Thanks for looking it!
In that case, @toxichl  you should file an issue on the vue-resource repo.
Since there's nothing to do on vue-router, I'll close this 🙂

ulivz:
No, I modify the vue-router's source code to solve this problem, not vue-resource types have problems. 😞@ktsn @posva

posva:
Can you submit a PR in that case?

ulivz:
I have submitted a PR in [click here](https://github.com/vuejs/vue-router/pull/1171)， you can reproduce the bug with the demo project I have given and fix my bug based on my PR. @posva

PR #1171, opened 2017-02-16
Modify the vue-router type declaration file and enable it to support TypeScript 2.0 complier

ulivz:
source code: import { ComponentOptions, PluginFunction } from "vue";

This source code in the TypeScript2.0 environment will be error:
" ... node_modules / vue-router / types / vue " has no exported member 'ComponentOptions'.
" ... node_modules / vue-router / types / vue " has no exported member 'PluginFunction'.
 
This bug can be reproduced from this project： [demo](https://github.com/toxichl/vue-typescript-template-ui)

 Since I have installed Vue2.0, so I directly quoted the vuejs type file.
 **The most reasonable thing is to remove the coupling and make it run properly.**

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -1,5 +1,16 @@
 import Vue = require("vue");
-import { ComponentOptions, PluginFunction } from "vue";
+/**
+ * source code: import { ComponentOptions, PluginFunction } from "vue";
+ *
+ * This source code in the TypeScript2.0 environment will be error:
+ * " ... node_modules / vue-router / types / vue " has no exported member 'ComponentOptions'.
+ * " ... node_modules / vue-router / types / vue " has no exported member 'PluginFunction'.
+ *
+ * Since I have installed Vue2.0, so I directly quoted the vuejs type file.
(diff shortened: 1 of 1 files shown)


posva:
I think we can safely remove the comments before merging 🙂

ulivz:
My PR itself is not standardized, hope vuejs as soon as possible according to my PR modify the available type declaration file.

ktsn:
Thank you for your PR. However this actually does not solve the problem.
I've investigated it again but the root cause is that `@types/vue-resource` depends on `@types/vue` which is for Vue v1 as I said on #1170.

`@types/vue` has the internal module declaration in it's file like `declare module 'vue'`, thus the TS compiler no longer search `vue` package even if there is a `vue` npm package with v2 typings. That is the reason you can load the v2 typings in that case by rewriting the module path to relative path.

The compilation of vue-router is passed by this fix but you still cannot use vue-resource because it does not augment Vue v2 typings.

posva:
Thanks for your contribution, we'll improve typings in #1685

PR #1184, opened 2017-02-22
improve passing-props translation

kazupon:
original:
`
Try to keep the props function stateless, as it's only evaluated on route changes. Use a wrapper component if you need state to define the props, that way vue can react to state changes.
`

see: https://router.vuejs.org/en/essentials/passing-props.html

### DIFF
--- a/docs/ja/essentials/passing-props.md
+++ b/docs/ja/essentials/passing-props.md
@@ -65,7 +65,6 @@ const router = new VueRouter({
 
 url `/search?q=vue` は `{query: "vue"}` をプロパティとして SearchUser コンポーネントに渡します。
 
-ルート変更時にのみ評価されるため、プロパティの機能はステートレスにしてください。
-ラッパーコンポーネントを使用すると、状態を変更して状態を変更することができます。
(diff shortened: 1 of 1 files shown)


docs/ja/essentials/passing-props.md
```diff
@@ -65,7 +65,6 @@ const router = new VueRouter({
 
 url `/search?q=vue` は `{query: "vue"}` をプロパティとして SearchUser コンポーネントに渡します。
 
-ルート変更時にのみ評価されるため、プロパティの機能はステートレスにしてください。
-ラッパーコンポーネントを使用すると、状態を変更して状態を変更することができます。
+ルート変更時にのみ評価されるため、props 関数はステートレスにしてください。プロパティを定義するために状態を必要とする場合は、その方法で vue は状態変更に対応することができます。
```
ktsn: `use a wrapper component` に対応する文章がないように見えます。

kazupon: あ、失敬。orz ...
修正しました！

ktsn (approved):

Issue #1183, opened 2017-02-20
"props" option works differently than explained in the docs when defining multiple components on a route, 


brad426:
Given a route entry like this:
```
{
  path: '/path/:test_prop',
  components: {
    default: Bar,
    a: Foo
  },
  props: true
}
```
Neither component Foo or Bar will receive the `test_prop` prop

Working example of issue https://jsfiddle.net/6du90epg/231/

LinusBorg:
I think this is "only" a problem of lacking documentation.

Looking at the source [here ](https://github.com/vuejs/vue-router/blob/5bc255af0a99e807dd202b16b26ab82e4ffa4707/src/create-route-map.js#L56-L58)and [here](https://github.com/vuejs/vue-router/blob/a4a91e70730fa4008ed7dc8c89e75583ff053164/src/components/view.js#L63), it's clear that the intended usage with multiple components is this:

```javascript
 {
  path: '/multiple/:test_prop',
  components: {
    default: Bar,
    a: Foo
  },
  props: {
    default: true,
    a: (route) => ({ test_prop: route.params.test_prop })
  }
}
```

With that, it works as intended: https://jsfiddle.net/Linusborg/6du90epg/237/

And this makes sense, because in most scenarios, each component will/might need a different set of props.

brad426:
Yup, that works perfectly. Thank you.

LinusBorg:
Re-opened because we still have to track improvement of docs here. ;)

LinusBorg:
merged imrpoved docs section, have to republish again tonight, though.

veonua:
```
{
      path: '/a',
      components: {
        default: Reader,
        toolbar: ReaderToolbar
      },
      props: { 
        default : (route) => ({ url: route.query.u, card_ref: route.query.r }), 
        toolbar : (route) => ({ url: route.query.u, card_ref: route.query.r })
      } 
    }
```

ReaderToolbar got no props

PR #1193, opened 2017-02-23
Props option usage with named views (solve #1183)

LinusBorg:
The `props` option in route configurations works a bit differently for routes that have multiple components for named routes.

I added a description of this differing API to the docs.

### DIFF
--- a/docs/en/essentials/passing-props.md
+++ b/docs/en/essentials/passing-props.md
@@ -27,6 +27,13 @@ const User = {
 const router = new VueRouter({
   routes: [
     { path: '/user/:id', component: User, props: true }
+    
+    // for routes with named views, you have to define the props option for each named view:
+    {
+      path: '/user/:id', 
+      components: { default: User, sidebar: Sidebar },
+      props: {user: true, sidebar: true }
+    }
   ]
 })
 ```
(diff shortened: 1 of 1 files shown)


docs/en/essentials/passing-props.md
```diff
@@ -27,6 +27,13 @@ const User = {
 const router = new VueRouter({
   routes: [
     { path: '/user/:id', component: User, props: true }
+    
+    // for routes with named views, you have to define the props option for each named view:
+    {
+      path: '/user/:id', 
+      components: { default: User, sidebar: Sidebar },
+      props: {user: true, sidebar: true }
```
znck: It could be confusing. `sidebar` is okay, it matches a key on components. Where does `user` come from?

znck: Am I mis-reading something?

znck: > Object mode

> When props is an object, this will be set as the component props as-is. Useful for when the props are static.

It would set `props.user` and `props.sidebar` to `true`. Right?

LinusBorg: oh, yeah i have a typo.

znck (approved):
LGTM

PR #1209, opened 2017-02-27
Fix params passed to `scrollHandler` (fix #1207)

### LINKED ISSUE #1207: scrollBehavior ( to, from, savedPosition )   console.log(to) == console.log(from) ;    //true
## Vue.js / vue-router versions

2.2.0

## Reproduction Link

https://jsfiddle.net/wjunjie/fg0eu4ey/

## Steps to reproduce

Click link: 
`
scrollBehavior ( to, from, savedPosition ) { 
     console.log(to) == console.log(from) ;    //true
}
`

## What is Expected?

**to**: Route: the target Route Object being navigated to.

**from**: Route: the current route being navigated away from.

## What is actually happening?

**to**: Route: the target Route Object being navigated to.

**from**: Route: the target Route Object being navigated to.

//  英文不好，希望回复中文，谢谢。

fnlctrl:
Currently `from` is always equal to `to` for `scrollHandler(to, from, savedPosition)`.
Should capture `this.current`(the `from` route) before making the transition to fix this.

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -30,17 +30,19 @@ export class HTML5History extends History {
   }
 
   push (location: RawLocation, onComplete?: Function, onAbort?: Function) {
+    const { current: fromRoute } = this
     this.transitionTo(location, route => {
       pushState(cleanPath(this.base + route.fullPath))
-      handleScroll(this.router, route, this.current, false)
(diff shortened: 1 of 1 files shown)


posva:
We definitely need a test on this one 😄

Issue #1214, opened 2017-02-28
Feature Request: Option for case sensitivity


lambertkuang:
The default option for case sensitivity in `path-to-regexp` is false and I couldn't find an option to set it from a new Vue router instance.  I'm trying to redirect some old paths in a project to new ones, and some of these new paths are just capitalized versions of the old, like `/all-stores` to `/All-Stores` but this causes an infinite redirect loop with the current setting.

posva:
Hello! URIs are case insensitive (even if servers can handle them differently), so there's no reason to add an option to allow creating case sensitive routes to vue-router. It'll make things confusing and allow people to create routes that are wrong.

> Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.

(https://tools.ietf.org/html/rfc3986#section-3.1)

smitt04:
Hey @posva The specification you posted there is only talking about the scheme (http, ftp, ws, etc...) According to the rfc spec the scheme and host are case insensitive, but the path portion of the URI is supposed to be case sensitive.

From https://tools.ietf.org/html/rfc3986#section-6.2.2.1

> For all URIs, the hexadecimal digits within a percent-encoding
   triplet (e.g., "%3a" versus "%3A") are case-insensitive and therefore
   should be normalized to use uppercase letters for the digits A-F.
> 
> When a URI uses components of the generic syntax, the component
   syntax equivalence rules always apply; namely, that the scheme and
   host are case-insensitive and therefore should be normalized to
   lowercase.  For example, the URI <HTTP://www.EXAMPLE.com/> is
   equivalent to <http://www.example.com/>.  The other generic syntax
   components are assumed to be case-sensitive unless specifically
   defined otherwise by the scheme.

According to that the scheme, host and the hexidecimal parts of the path are case insensitive, but all other components should be case sensitive.

fnlctrl:
> but all other components should be case sensitive.

Sorry, but there is no such statement in the RFC.

lambertkuang:
That link above (https://tools.ietf.org/html/rfc3986#section-6.2.2.1) does say the other components are assumed to be case-sensitive.

> The other generic syntax
   components are assumed to be case-sensitive unless specifically
   defined otherwise by the scheme (see Section 6.2.3).

And Section 6.2.3 (https://tools.ietf.org/html/rfc3986#section-6.2.3) has no mention of the path, so we can assume it to be case sensitive.

[...]

PR #1215, opened 2017-02-28
Case Sensitive Option for Routes #1214

smitt04:
Added both a `caseSenstive` and `pathToRegexpOptions` property on RouteConfig.  The `caseSensitive` property will override and add to `pathToRegexpOptions` which defaults to `{}`. 

Along with that `RouteRecord` gets `pathToRegexpOptions` that comes from `RouteConfig`

### DIFF
--- a/examples/redirect/app.js
+++ b/examples/redirect/app.js
@@ -9,6 +9,8 @@ const Foo = { template: '<div>foo</div>' }
 const Bar = { template: '<div>bar</div>' }
 const Baz = { template: '<div>baz</div>' }
 const WithParams = { template: '<div>{{ $route.params.id }}</div>' }
+const Foobar = { template: '<div>foobar</div>' }
+const FooBar = { template: '<div>FooBar</div>' }
 
 const router = new VueRouter({
   mode: 'history',
@@ -51,10 +53,10 @@ const router = new VueRouter({
     { path: '/redirect-with-params/:id', redirect: '/with-params/:id' },
 
     // redirect with caseSensitive
-    { path: '/foobar', redirect: '/FooBar', caseSensitive: true },
+    { path: '/foobar', component: Foobar, caseSensitive: true },
 
     // redirect with pathToRegexpOptions
-    { path: '/FooBar', redirect: '/FOOBAR', pathToRegexpOptions: { sensitive: true }},
+    { path: '/FooBar', component: FooBar, pathToRegexpOptions: { sensitive: true }},
 
     // catch all redirect
     { path: '*', redirect: '/' }
@@ -102,13 +104,13 @@ new Vue({
         <li><router-link to="/redirect-with-params/123">
           /redirect-with-params/123 (redirects to /with-params/123)
         </router-link></li>
-        
+
         <li><router-link to="/foobar">
-          /foobar (redirects to /FooBar)
+          /foobar
         </router-link></li>
-        
+
         <li><router-link to="/FooBar">
-          /FooBar (redirects to /FOOBAR)
+          /FooBar
         </router-link></li>
 
         <li><router-link to="/not-found">
--- a/examples/redirect/app.js
+++ b/examples/redirect/app.js
@@ -54,7 +54,7 @@ const router = new VueRouter({
     { path: '/foobar', redirect: '/FooBar', caseSensitive: true },
 
     // redirect with pathToRegexpOptions
-    { path: '/FooBar', redirect: '/bar', pathToRegexpOptions: { sensitive: true }},
+    { path: '/FooBar', redirect: '/FOOBAR', pathToRegexpOptions: { sensitive: true }},
 
     // catch all redirect
     { path: '*', redirect: '/' }
--- a/examples/redirect/app.js
+++ b/examples/redirect/app.js
@@ -108,7 +108,7 @@ new Vue({
         </router-link></li>
         
         <li><router-link to="/FooBar">
-          /FooBar (redirects to /bar)
+          /FooBar (redirects to /FOOBAR)
         </router-link></li>
 
         <li><router-link to="/not-found">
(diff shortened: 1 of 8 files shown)


posva (commented):
Could you add some test case, please?
I have one concern, it's about the route cache: if we allow overwriting routes, changing the caseSensitive option wouldn't do anything. We can also point it out in docs though, I don't see an actual use case for that but it's a gotcha 😆

smitt04:
I was looking at where to add tests. I couldn't find any similar tests in `unit` and not sure where to add in `e2e`, didn't see any redirect tests (which is where we were having the problem). I did add it to the examples though. And before my code was working the examples broke. I would be happy adding other tests if you could point me in the right direction.

smitt04:
@posva I have added some tests, let me know if this is good.

smitt04:
Do we have an eta for this?

posva (approved):

posva:
Sorry for the delay 😅

posva:
I'm looking into this again, there are some conflicts I'm fixing. I should get a working pr soon

posva:
Right now, the global option doesn' work. I'm looking at that

posva:
The things that I'm not sure yet:

- Should we remove the `caseSensitive` option?
  - if yes, keeping only the `pathToRegexOptions` restrict us to using that lib
  - if no, we'd have 2 ways of setting the case sensitive option. Why the case is aliased while the others aren't?
- The naming of `pathToRegexOptions` is a bit too long and very technical, maybe `pathToRegex` is enough but it's unclear. `regexOptions` or `pathOptions` are other alternatives

posva:
Note to myself, do not forget to close #1214 when merging

posva:
ping @yyx990803 tldr: https://github.com/vuejs/vue-router/pull/1215#issuecomment-302882548

rivieiraa:
When will this be merged ?

dotBits:
I'm waiting for this merge as well. Is important to have consistency between indexed URLs with trailing slashes, otherwise there will be duplicate URLs for crawlers

MachinisteWeb:
Champagne ! (that resolve also my point: https://github.com/vuejs/vue-router/issues/1443)

Thx guys.

MachinisteWeb:
Where is the `pathToRegexpOptions` documentation ? Or maybe an example ?

For example the `{ strict: true }` mode not work as expected. It's allows only url not ending with `/` to work. And it's not work with defaultpage.

I have only : https://router.vuejs.org/en/api/options.html#routes that indicate the property exist, but I do not have any example or detail. Where I know ``{ strict: true }`  exist and what is it's true purpose ?

rivieiraa:
@Haeresis 
> Where is the pathToRegexpOptions documentation ?

[pathToRegexp github page](https://github.com/pillarjs/path-to-regexp#usage)
Try this.

MachinisteWeb:
I will add this information to the vue-router doc. Thx @rivieiraa !

rivieiraa:
It's pathToRegex(p)Options

> strict When false the trailing slash is optional. (default: false)

It is correct to use `pathToRegexpOptions: { strict: true }`

Haeresis:
So I have an issue. If I set `pathToRegexpOptions: { strict: true }`, only page « without » trailing work. And I expected page only « with » that's work...

MachinisteWeb:
So I have an issue. If I set `pathToRegexpOptions: { strict: true }`, only page « without » trailing work. And I expected page only « with » that's work...

On my server I create:

`/` and `/test/` for exemple

`/` <= OK
`/test/` <= OK
`/test` <= NOK

and with `vue-router`, by default it's that:

`/` <= OK
`/test/` <= OK
`/test` <= OK

With usage of `pathToRegexpOptions: { strict: true }` I expected this

`/` <= OK
`/test/` <= OK
`/test` <= NOK

But what I have is that

With usage of `pathToRegexpOptions: { strict: true }`

`/` <= OK
`/test/` <= NOK
`/test` <= OK

...

Normal?

rivieiraa:
I understand your situation.

> that resolve also my point: #1443

Nope, it doesn't actually. This particular pull request was about case sensitivity but also involved another pathToRegexp options as they all passed together.

MachinisteWeb:
So, if I correctly understand,
you think it's not normal and
there is another pull request dedicated to `pathToRegexp `.

You know the link to this pull request?

rivieiraa:
Sure, i believe there must be an option to strict trailing slashes more clearly (to make slash mandatory or to deprecate it). But as long as vue router uses pathToRegexp it's a problem of third party software and i'm not sure if it could be done anyways.
No, there is no PR dedicated to trailing slash, but there is an issue you pointed to earlier #1443

MachinisteWeb:
I will ask to re-open it so. Thx a lot !

posva:
@Haeresis I created an issue on the repo instead: https://github.com/pillarjs/path-to-regexp/issues/112 Feel free to share your thoughts there

rivieiraa:
@posva 
hello! what is the difference between `caseSensitive` and `pathToRegexpOptions: { sensitive: true }` options?

posva:
the same, it's just that caseSensitive seems like a very common use case

rivieiraa:
@posva 
not sure if it works.
i'm doing my route definition like that:
`...
{
    name: 'about',
    path: '/company/about',
    caseSensitive: true,
    pathToRegexpOptions: { strict: true },
    component: require('./components/About.vue')
}
...`

then i'm requesting something like /comPany/abOut and vue-router replaces my route with /company/about and displays requested page as if i typed it right. Is that intentional or i'm doing something wrong? I suppose pathToRegexp should not match this kind of address and show wildcard page if it was defined.
But, there is also another strange behavior. If i put capitalized path like /Company/About or /company/About then it shows my 404 page.

adamreisnz:
Has this been removed? `pathToRegexpOptions: {strict: true}` doesn't seem to do anything, still getting URL's with a trailing slash.

PR #1228, opened 2017-03-07
add note for passing props on nested routes

rodneyrehm:
This might seem obvious, but this sent me in circles for the first hour of using Vue and VueRouter…

### DIFF
--- a/docs/en/essentials/passing-props.md
+++ b/docs/en/essentials/passing-props.md
@@ -40,6 +40,8 @@ const router = new VueRouter({
 
 This allows you to use the component anywhere, which makes the component easier to reuse and test.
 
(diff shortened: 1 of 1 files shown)


posva:
What do you mean, they do work: http://jsfiddle.net/posva/zfqrvraf/

rodneyrehm:
oh no! I was mislead into thinking it's the sub-routes that are the problem, but it's the named router views that seem to be breaking the passing of props. tests cases: [simple named router view](http://jsfiddle.net/zyvzvtpk/2/),  [with sub-routes](http://jsfiddle.net/s0vx45tz/1/). Is that intended behavior?

PR #1233, opened 2017-03-08
append a feature help to change document.title

sqchenxiyuan:
by add `title` to route config 

like 

```
  routes: [
    { path: '/', title: 'home', component: Home },
  ]
})
```

When the current page is specified, the title will be changed

----

reason: I have seen a request in the past , but not successfully completed the merge, and there is no continuation of the merger, so I pull this merge.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -139,9 +139,16 @@ export class History {
     const prev = this.current
     this.current = route
     this.cb && this.cb(route)
+    this.updateTitle(route)
     this.router.afterHooks.forEach(hook => {
       hook && hook(route, prev)
     })
+  }
+
+  updateTitle (route: Route) {
+    if (route.title !== '') {
+      window.document.title = route.title
+    }
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/history/base.js
```diff
@@ -139,10 +139,17 @@ export class History {
...
     this.cb && this.cb(route)
+    this.updateTitle(route)
     this.router.afterHooks.forEach(hook => {
       hook && hook(route, prev)
     })
   }
+
+  updateTitle (route: Route) {
+    if (route.title !== '') {
+      window.document.title = route.title
```
znck, line 150: IMO this would cause problem with SSR.

sqchenxiyuan: my code is not good       

dose there need a judge for window ?  or how to solve  this?  can you help me?

znck: /ping @posva

znck:
This should not be the responsibility of the router, as it can be used in multiple environments.

Also it could be done with in after hook.

PR #1240, opened 2017-03-11
Add configuration settings for IIS (Windows)

AngeloAnolin:
Added web.config settings for those who would be using Internet Information Services (IIS) to host their VueJS app with Vue Router in History mode.

### DIFF
--- a/docs/en/essentials/history-mode.md
+++ b/docs/en/essentials/history-mode.md
@@ -44,7 +44,7 @@ location / {
 
 For Node.js/Express, consider using [connect-history-api-fallback middleware](https://github.com/bripkens/connect-history-api-fallback).
 
-### Internet Information Services (IIS)
+#### Internet Information Services (IIS)
 ```
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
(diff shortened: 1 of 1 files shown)


posva (commented):
I've never IIS so I can't really review this but it looks good to me to have more server configs

josemadriz:
IIS setting should be.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
  <system.webServer>
    <rewrite>
      <rules>
        <rule name="Handle History Mode and custom 404/500" stopProcessing="true">
            <match url="(.*)" />
            <conditions logicalGrouping="MatchAll">
              <add input="{REQUEST_FILENAME}" matchType="IsFile" negate="true" />
              <add input="{REQUEST_FILENAME}" matchType="IsDirectory" negate="true" />
            </conditions>
          <action type="Rewrite" url="index.html" />
        </rule>
      </rules>
    </rewrite>
  </system.webServer>
</configuration>
```
as described here note that the difference is that index.html is determined as default page. 
In case the application is served from a subdirectory
https://www.linkedin.com/pulse/hosting-vue-js-spa-build-microsoft-iis-zainul-zain/
for the application to work on IIS.

PR #1252, opened 2017-03-15
Update navigation.md

christianmalek:
The sentence above the changed text mentions that the string represents the path. But I think `literal string path` will be much clearer because not everyone wants to read the text. So it's instantly recognizeable that if you just pass a simple string it's a path and not the route name.


posva:
Thanks!

PR #1290, opened 2017-03-29
fix target _blank when anchor has children

RezaHaidari:
Couple of days ago, we have faced an issue with out non-SPA project:
Once you add a child element to a ``<router-link>`` with ``target="_blank"`` attribute, it wouldn't fire on a new tab. check this fiddle:
https://jsfiddle.net/asynchronous/dnLs1dm4/10/

I've changed ``gaurdEvent`` method in ``link`` component so it could fix this issue.

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -82,7 +82,12 @@ export default {
     return h(this.tag, data, this.$slots.default)
   }
 }
-
+// return first anchor parent of element
+function getAnchorParent (event) {
+  return event.path.find(parent => {
+    return parent.localName === 'a'
+  })
+}
 function guardEvent (e) {
   // don't redirect with control keys
   if (e.metaKey || e.ctrlKey || e.shiftKey) return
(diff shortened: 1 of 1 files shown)


posva:
What browser are you facing the issue? It's working properly on Chrome

RezaHaidari:
@posva we've checked on chrome and firefox (both ubuntu and windows).

are you sure you're clicking on span link ?

posva:
yes, both links behave the same (chrome 57)

Javid-Izadfar:
@posva anchor with child is not opening in new tab for me too. 

ezgif-1-9275d1a181

posva:
I see, it doesn't work on firefox.
Maybe we should use `currentTarget` instead of `target`
BTW, `target` should be the `a` link in that case (topmost element). It may be a Firefox bug

Javid-Izadfar:
Nah man, same thing on Chrome ... even multiple PCs :/
you are the first person who says it works fine for him

RezaHaidari:
@posva yes currentTarget fixed that issue.

posva (approved):
ping @fnlctrl @yyx990803

PR #1291, opened 2017-03-30
make a copy when query param value is an array (fix #1182)

### LINKED ISSUE #1182: $router.push({query: this.myarray}) not updating the URL
2.1.2

### Reproduction Link
http://jsfiddle.net/ub9e33fm/2/

### Steps to reproduce
1. Click first link
2. Click second link: see the difference in printed vars

### What is Expected?
Clicking the second link will also update `this.$route.query`

`this.$route.query` is not updated

javoski:
fix #1182 
Don't using the object reference as query value directly.

### DIFF
--- a/src/util/query.js
+++ b/src/util/query.js
@@ -19,21 +19,18 @@ export function resolveQuery (
   query: ?string,
   extraQuery: Dictionary<string> = {}
 ): Dictionary<string> {
-  if (query) {
-    let parsedQuery
-    try {
-      parsedQuery = parseQuery(query)
-    } catch (e) {
-      process.env.NODE_ENV !== 'production' && warn(false, e.message)
-      parsedQuery = {}
-    }
-    for (const key in extraQuery) {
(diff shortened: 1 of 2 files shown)


src/util/query.js
```diff
@@ -19,21 +19,18 @@ export function resolveQuery (
...
-    } catch (e) {
-      process.env.NODE_ENV !== 'production' && warn(false, e.message)
-      parsedQuery = {}
-    }
-    for (const key in extraQuery) {
-      parsedQuery[key] = extraQuery[key]
-    }
-    return parsedQuery
-  } else {
-    return extraQuery
```
posva, line 35: It worries me that deleting this part didn't break the test. @fnlctrl Do you have an idea about this?

javoski: There is already an empty value check in parseQuery function.

posva: Yeah, but in the other case we were skipping the whole parseQuery and for loop and returning right away the `extraQuery` instead

PR #1304, opened 2017-04-04
Add note about $router in navigation

### LINKED ISSUE #1136: Docs: Mention $router in "Programmatic Navigation" section
This may be obvious, but it took me some time to figure out.

I wanted to navigate programmatically in a component and it wasn't clear for me how to do it. The obvious answer is to use `this.$router.go` in a component method, but the "Programmatic Navigation" section in the documentation only refers to the "Router instance".

As far as I can tell, the only section that mentions the injected instance is the "Component injections" section.

Thanks.

posva:
Closes #1136


kazupon (approved):

posva:
Thanks @kazupon !
It's so cool you always review many PR 🙂

Issue #1268, opened 2017-03-17
Query parser doesn't support nested objects


danderozier:
### Vue.js / vue-router versions
2.1.10 / 2.2.1

### Reproduction Link
https://jsfiddle.net/danderozier/jhcbu84y/1/

### Steps to reproduce
In your vue-router enabled vue app, push a nested object as the router query. For example, take the following statement: 
```
this.$router.push({ query: { order: { name: 'asc' } } })
```

### What is Expected?
One would expect the query string to be built correctly. For the example above, this would be: 
```
?order[name]=asc  // url decoded for legibility
```

### What is actually happening?
Because nested objects aren't supported, the example above results in the query string 
```
?order=[object Object]  // url decoded for legibility
```

Somewhat related: in #1259 someone suggests using qs or a similar library as a querystring parser, which has object support built in.

posva:
Maybe we should use qs to parse query strings, it's only 2kb gzipped

trikadin:
@danderozier 
Small clarification: I don't suggest to change query lib, but to add ability to use any custom stringifier/parser (JSON, for example).

posva:
@danderozier Yes, I know 😄 , and we're probably going to do that. That way people will be able to use anything they want

posva:
Closing in favour of #1259

danderozier:
Looking forward to it—qs would be perfect for me in this particular case. Cheers!

Issue #1259, opened 2017-03-16
Custom querystring parser


trikadin:
Please, add the opportunity to use custom qs lib instead of built-in.

posva:
This was shipped in 2.4.0 https://router.vuejs.org/en/api/options.html#parsequery--stringifyquery

ganigeorgiev:
It is not very clear from the documentation what the parameters and their types are. From the source, I suppose that `parseQuery`  receives a `String` and `stringifyQuery` receives an `Object`.
```javascript
// https://github.com/ljharb/qs
import qs from 'qs';

const router = new Router({
    routes: [
        // ...
    ],
    // set custom query resolver
    parseQuery(query) {
        return qs.parse(query);
    },
    stringifyQuery(query) {
        var result = qs.stringify(query);

        return result ? ('?' + result) : '';
    }
});
```

rayrutjes:
I think we should allow to override the query parsing methods on a per route basis.
When you are trying to use a third party component that has special requirements regarding query parsing, you don't want to change the way existing URLs are parsed.
Does that make sense @posva , is it maybe already possible?

posva:
@rayrutjes The url being global, I really think that's a bad idea to change it based on the current component. Appart from that I don't think it's even doable

rayrutjes:
I think you are right @posva, per route doesn't make sense.
Then, I would militate to have a better url parsing by default that handles nested objects & co.
qs works perfectly ;) but we would like to not have existing users to have to change the way urls are parsed for our library.

For now, we are going to make sure that we are compatible with what is supported.

posva:
I proposed that already some time ago but vue-router is 8.1kb gziped and qs is 2.6gziped.
Most people don't need nesting in queries, so being able to remove 20% of the size for most people is definitely a good thing

rayrutjes:
Yes I agree, sadly.
We will work on a solution on our side ;)

christhofer:
@posva can you add ganigeorgiev's comment above to the documentation?

Current documentation doesn't show how to use ths option.
![image](https://user-images.githubusercontent.com/13524958/64775232-defa0180-d57f-11e9-9246-1a1e70625085.png)

omerciftcibasi:
Addition to ganigeorgiev's comment, the query parameters inside routes will be automatically converted url string and parsed as an object when accessing $router.query

tomisicm:
[...]

PR #1308, opened 2017-04-06
Fix nested query parser follow issue #1268

thangman22:
Normally when user pass query to push method like this

`this.$router.push({ query: { order: { name: 'asc' } } })`

Vue router will convert to 

`?order=[object Object]`

Instead of 

`?order[name]=asc`

I changed query parser for custom script to 'qs' module that support this use case

### DIFF
--- a/examples/active-links/app.js
+++ b/examples/active-links/app.js
@@ -60,7 +60,11 @@ new Vue({
             /users/evan?foo=bar&baz=qux
           </router-link>
         </li>
-
+        <li>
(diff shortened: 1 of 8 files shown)


posva:
Hey thanks for the PR! Unfortunately, we don't want to add `qs` to keep vue-router lean. However, if you're interested in fixing that nested object behaviour, what we do want to add is the possibility to use a custom query parser (like `qs`). The related issue is https://github.com/vuejs/vue-router/issues/1259
We haven't decided on an API yet, but you can propose some there 🙂

Issue #1312, opened 2017-04-07
Missing TS definition for router push


druppy:
### Version
2.3.0

### Reproduction link
[https://jsfiddle.net/7syt1bdu/](https://jsfiddle.net/7syt1bdu/)

### Steps to reproduce
When writing `router.push({name: 'xx', params: {id: 2}}` TS will complain about the type is not right.

### What is expected?
Problem free TS compilation :-)

### What is actually happening?
TS error 

---
adding this to `types/router.d.ts`, fix the issue :

``` typescript
export interface LocationObject {
    name?: string,
    path?: string,
    params?: Object
    query?: Object
}
export type RawLocation = string | Location | LocationObject;
```

posva:
A PR would be even better 🙂

druppy:
Ok, i make the PR danse 😁

druppy:
Location `params` and `query` may only hold string values, and I passed along a number. Sorry, move on ... nothing to see :-)

PR #1313, opened 2017-04-07
Fix for issue #1312

druppy:
Hope this will do :-)


posva:
Mhh, now that I see it all, it's weird that you get a TS error because the `Location` type should fulfil your requirements, but I may be missing something. @ktsn Can you help me out please 😇

druppy:
You are right ... the `Location` interface is just not working with my `params` object, the Dictionary template gives some problems here. My version just used the type `Object` for both `query` and `params` .

druppy:
Ok, I am sorry ... it needs to have ONLY strings as values, and I had a number. This makes a lot of sense, now that I understand.

Forget this PR ...

PR #1335, opened 2017-04-14
improved for...loop a bit

dsonet:
Since this function would be invoked a lot, it's worthy to be optimized.


billouboq:
I think most browser already optimize this for you.

yyx990803:
Yeah, this is no longer necessary in most cases, thanks for the PR anyway!

PR #1353, opened 2017-04-20
modify method when pass absolute path for option.base

thebeautyoflove:
my colleague and me use a static html to load pages from other site to prevent cross-main.
so when we want when we pass base like 'http://xxx.com' in option . vue-router can join the url accurately;
thank you

### DIFF
--- a/dist/vue-router.common.js
+++ b/dist/vue-router.common.js
@@ -561,7 +561,9 @@ function parsePath (path) {
 }
 
 function cleanPath (path) {
-  return path.replace(/\/\//g, '/')
+  var ishttp = /^\w+:\/\//.exec(path);
+  var http = Array.isArray(ishttp) ? ishttp[0] : '';
+  return http + path.substr(http.length).replace(/\/{2,}/g, '/');
 }
 
 /*  */
@@ -1759,7 +1761,9 @@ function normalizeBase (base) {
     }
   }
   // make sure there's the starting slash
-  if (base.charAt(0) !== '/') {
+  if(/^[a-z]+:\/\//i.test(base)){
+    
+  }else if (base.charAt(0) !== '/') {
     base = '/' + base;
   }
   // remove trailing slash


dist/vue-router.common.js
```diff
@@ -561,7 +561,9 @@ function parsePath (path) {
 }
 
 function cleanPath (path) {
-  return path.replace(/\/\//g, '/')
+  var ishttp = /^\w+:\/\//.exec(path);
```
blqw, line 564: The RepExpr here needs to be modified. because there's still this kind of situation:
```
//google.com
```
so the RegExpr should be it
```js
/^([a-z]+:)?\/\//i
```

dist/vue-router.common.js
```diff
@@ -1759,7 +1761,9 @@ function normalizeBase (base) {
     }
   }
   // make sure there's the starting slash
-  if (base.charAt(0) !== '/') {
+  if(/^[a-z]+:\/\//i.test(base)){
```
blqw, line 1764: The RepExpr here needs to be modified. because there's still this kind of situation:

//google.com
so the RegExpr should be it

/^([a-z]+:)?\/\//i

blqw (commented):
Url, take care of the reference. http://jsrun.net/IxkKp/edit

posva:
Sorry for the delay. It looks like #1426 is something that may be related. There's still a bit more to investigate to make sure a correct solution is provided

The modifications should be to the source files. The files on the dist folder are generated 🙂 
Thanks a lot for the PR but I'm closing this as we still need to do some checks

dist/vue-router.common.js
```diff
@@ -1759,7 +1761,9 @@ function normalizeBase (base) {
...
+    
+  }else if (base.charAt(0) !== '/') {
     base = '/' + base;
   }
   // remove trailing slash
```
hgq520, line 1769: 请问，需要修改源码才能实现在History模式下去掉http://api.com/http:/static.com/test前面的部分吗

lxx2013: 一般`/`开头的都是站内link , 如果是站外 link 的话, 可以使用`<a target="_blank"></a>`

PR #1387, opened 2017-04-30
Prevent unnecessary Exception with non-existent links 

### LINKED ISSUE #1386: Missing named routes throw TypeErrors
### Version
2.5.2

### Reproduction link
[http://jsfiddle.net/qkco53b9/1/](http://jsfiddle.net/qkco53b9/1/)

### Steps to reproduce
Run the example. Look at the console.

### What is expected?
Inserting a router link to a named route that doesn't exist should warn.

The console spews both warnings about the missing named routes and errors about not being able to access the `regex` property of a route.

---
I'm building an SPA and inserting placeholder links to pages I haven't built yet.

It looks like the problem might be here: https://github.com/vuejs/vue-router/blob/4de549a45cc38975d6ce343ecdea007972891b4f/src/create-matcher.js#L39. Seems the matcher still tries to proceed even if there's no record.

posva:
Closes #1386

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -33,10 +33,10 @@ export function createMatcher (
 
     if (name) {
       const record = nameMap[name]
-      if (process.env.NODE_ENV !== 'production' && !record) {
-        warn(false, `Route with name '${name}' does not exist`)
-        return _createRoute(null, location)
+      if (process.env.NODE_ENV !== 'production') {
+        warn(record, `Route with name '${name}' does not exist`)
       }
+      if (!record) return _createRoute(null, location)
       const paramNames = record.regex.keys
         .filter(key => !key.optional)
         .map(key => key.name)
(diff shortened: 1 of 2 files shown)


kazupon (approved):

src/create-matcher.js
```diff
@@ -33,8 +33,9 @@ export function createMatcher (
 
     if (name) {
       const record = nameMap[name]
-      if (process.env.NODE_ENV !== 'production') {
-        warn(record, `Route with name '${name}' does not exist`)
+      if (process.env.NODE_ENV !== 'production' && !record) {
```
yyx990803: behavior should be consistent between prod and dev, so we should move the env check to only apply to the warning.

yyx990803 (changes requested):

posva:
I forgot to ping you @yyx990803 😓

PR #1403, opened 2017-05-10
EN Doc `getting-started.md` : `standalone` to `full`

MachinisteWeb:
Change `standalone` to new `full` term as using into official https://vuejs.org/v2/guide/installation.html
+ update link to https://vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only


posva:
Thanks!

PR #1405, opened 2017-05-10
Doc EN : Add ` on code part to consistency and comprehension

MachinisteWeb:
All is in title ;) It's all we have see/change for getting-started.md while our translation process.

### DIFF
--- a/docs/en/essentials/getting-started.md
+++ b/docs/en/essentials/getting-started.md
@@ -17,7 +17,7 @@ Creating a Single-page Application with Vue.js + vue-router is dead simple. With
   <p>
     <!-- use router-link component for navigation. -->
     <!-- specify the link by passing the `to` prop. -->
-    <!-- <router-link> will be rendered as an `<a>` tag by default -->
+    <!-- `<router-link>` will be rendered as an `<a>` tag by default -->
     <router-link to="/foo">Go to Foo</router-link>
     <router-link to="/bar">Go to Bar</router-link>
   </p>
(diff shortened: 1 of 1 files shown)


posva:
Thanks for the contribution. Since the text won't change because it's inside of a code block, I really think it's better to have them without the backticks as it makes it easier to read
WDYT @fnlctrl ?

MachinisteWeb:
Ok but in this case that means which things like `<a>` or `to` which will be originally between « tick » which must be « without tick »because, for example, in this sentence:

```js
<!-- <router-link> will be rendered as an `<a>` tag by default -->
```

there is a lack of consitency between « &lt;router-link> » and « `<a>` »

------

Our advice is this:

```js
// short for `routes: routes`
```

is more readable than this

```js
// short for routes: routes
```

« `:` is a part of a phrase ponctuation? »

posva:
It's true it's inconsistent and I also think that one is more readable
because it's multiples words while the other is just html
It's not a big deal and really depends on tastes, docs will be readable
either ways, so I'll let Evan decide

MachinisteWeb:
Some others cases in all documentation and others PR (like following) tend to thinking Evan choice to also tick:

- https://github.com/vuejs/vue-ssr-docs/pull/33
- https://github.com/vuejs/vue-ssr-docs/pull/28

so let him save some time @posva ;)

posva:
yeah, I've been merging them for some time now. Feel free to add that to other PRs as well

PR #1425, opened 2017-05-17
Fix nested object comparison for queries

### LINKED ISSUE #1421: Deeply nested objects in query are considered equal when they are not really equal
### Version
2.5.3

### Reproduction link
[https://jsfiddle.net/awei01/5v9nqe7o/](https://jsfiddle.net/awei01/5v9nqe7o/)

### Steps to reproduce
Open the JS fiddle. 
Open console logger
Click the link, notice that the query is updated and console logs "route completed"
Click link again. Notice the query is not updated and console logs "route aborted"

### What is expected?
The route should change because the query object has changed. 

The route inspector should see that these objects are not equal. The implementation of "isEqualObject" doesn't take into account deeply nested objects. 

see these lines: https://github.com/vuejs/vue-router/blob/2f97b1930328a2610109ceac679248ac7f8b8585/src/util/route.js#L76-L83

---
I'm using my own custom implementation of parseQuery and stringifyQuery for the Router object so, I'd like to be able to allow the route to continue. Specifically I'm using 'urlon' npm package which can handle nested objects in the querystring.

posva:
Fix #1421

### DIFF
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -79,7 +79,15 @@ function isObjectEqual (a = {}, b = {}): boolean {
   if (aKeys.length !== bKeys.length) {
     return false
   }
-  return aKeys.every(key => String(a[key]) === String(b[key]))
+  return aKeys.every(key => {
+    const aVal = a[key]
+    const bVal = b[key]
+    // check nested equality
+    if (typeof aVal === 'object' && typeof bVal === 'object') {
+      return isObjectEqual(aVal, bVal)
+    }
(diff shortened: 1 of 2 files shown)


posva:
ping @yyx990803

rayrutjes:
OMG I wish I saw this earlier, just spent a decent amount of time wondering why route transition broke on certain queries.
Thought it had to do with my objects being observed but it was just as simple as upgrading to 2.6.0.

PR #1431, opened 2017-05-17
Add note about current and append in resolve

### LINKED ISSUE #1430: [FEATURE] A url generation method that returns the compiled url for the given named route populated with the given params.
This feature will solve a simple problem, having urls in translations. 
Today if i have a translation text that has anchors linking to some part of the website, i can't generate a url based on a named route.

"But you can just inject the url hardcoded" - well yes, but where's the point of having named routes if i can't generate my own url at some point to just have the "text value" of my url ?

Other use-case : a user has referral url he wants to share ton somebody, well i can't output the url to the user to copy / paste.

```
@param Object options (the route options as for $router.push())
@param Boolean prependHostName (wether to prepend or not hostname)
@return String the url of the named route

this.$router.href([options], prependHostName)
```

e.g :
```
 this.$router.href({name: 'pages.policies'}) => /policies
 this.$router.href({name: 'pages.policies'}, true) => www.yourhostname.com/policies
```

posva:
Closes #1430

### DIFF
--- a/docs/en/api/router-instance.md
+++ b/docs/en/api/router-instance.md
@@ -56,6 +56,9 @@
   }
   ```
 
+- `current` is the current Route by default (most of the time you don't need to change this)
+- `append` allows you to append the path to the `current` route (as with [`router-link`](https://router.vuejs.org/en/api/router-link.html#props))
+
 - **router.addRoutes(routes)**
 
   > 2.2.0+


posva:
This was missing from the docs @yyx990803 @fnlctrl

PR #1442, opened 2017-05-20
Silence default child warning when redirecting

decademoon:
There is no need to warn about a named route that has a default child route if the named route intentionally redirects.

### DIFF
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -77,11 +77,11 @@ function addRouteRecord (
   }
 
   if (route.children) {
-    // Warn if route is named and has a default child route.
+    // Warn if route is named, does not redirect and has a default child route.
     // If users navigate to this route by name, the default child will
     // not be rendered (GH Issue #629)
     if (process.env.NODE_ENV !== 'production') {
-      if (route.name && route.children.some(child => /^\/?$/.test(child.path))) {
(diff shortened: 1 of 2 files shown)


posva:
Hey, thanks for the contribution. Could you add a unit test on the create-map.spec.js file, please?

posva (approved):
Just out of curiosity, do you use this to redirect to a different child in your app?

decademoon:
No, my setup is pretty much like this:

``` js
{
  path: 'foo',
  name: 'foo',
  redirect: { name: 'foo-index' },
  component: require('components/foo/master.vue'),
  children: [
    {
      path: '',
      name: 'foo-index',
      component: require('components/foo/index.vue'),
    },
    {
      path: ':fooId',
      name: 'foo-show',
      component: require('components/foo/show.vue'),
    },
    // etc...
  ],
},
```

And I write links like this:

``` html
<router-link :to="{ name: 'foo' }">Foo</router-link>
```

tebs1200:
@posva, I do use a redirect to go to a specific child route. I have a route like `user/:id` which I redirect to one of the nested routes `user/:id/feed`.

posva:
@tebs1200 ok, thanks for letting me know 🙂

PR #1445, opened 2017-05-20
Update route.js

lzxb:
Feat route.search
```javascript
export default {
 ///... options
 watch: {
   '$route.search' () {
     // update data
   }
 }
}
```

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -83,6 +83,7 @@ declare type Route = {
   hash: string;
   query: Dictionary<string>;
   params: Dictionary<string>;
+  search: string;
   fullPath: string;
   matched: Array<RouteRecord>;
   redirectedFrom?: string;
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -19,8 +19,13 @@ export function createRoute (
     hash: location.hash || '',
     query: location.query || {},
     params: location.params || {},
+    search: '',
     fullPath: getFullPath(location, stringifyQuery),
     matched: record ? formatMatch(record) : []
+  }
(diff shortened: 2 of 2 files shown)


posva:
What feature are you trying to add that Vue doesn't have?

lzxb:
Yes, but this submission seems to be a bug

posva:
What are you trying to add with that search property? It looks like the `hash` property
If you have a bug please, open an issue with a repro

lzxb:
It's the same thing as $route.query, but sometimes it takes the form of a string

posva:
Thanks for the pr, but we'll pass on this for the moment. You can open a new issue explaining the feature request if you want but I don't feel as if this is something we need.

lzxb:
Because $ruote.query is a reference type, when I store a $ruote.query, when I switch back different routes back, then I compare the stored $ruote.query with the current $ruote.query and return the false. When using $ruote.search, it won't fail because it is a string type comparison

posva:
In that case you can compare both queries converting them into strings
first. Just in case, you should not modify the query object because the
$route object is immutable

lzxb:
```javascript
{
  // ... options
  created () {
    this.$router.replace({ query: { test: new Date().getTime() } })
    console.log('ok')
  },
  watch: {
    $route: {
      deep: true,
      handler: function (to, from) {
        console.log(to, from, to.query === from.query) // query change
        // $route.search It's actually converting $route.query into strings
      }
    }
  }
}
```

posva:
You should use an object comparison function in your case.

PR #1477, opened 2017-05-29
Doc EN: Consitency

MachinisteWeb:
After french translation, we have add consistency on all this English point based on others documentations.


posva (approved):

posva:
thanks

PR #1483, opened 2017-05-31
Proposed update from require.ensure() to import()

rodet:
Hello,
I just started investigating in this area so let me know if my assumptions are wrong. The [Webpack documentation](https://webpack.js.org/guides/code-splitting-async/) linked here states:

```
webpack supports two similar techniques to achieve this goal: using import() (preferred, ECMAScript proposal) and require.ensure() (legacy, webpack specific).
```

I was able to make `import()` work with a very simple change. I'm taking the assumption that we'd update this documentation to take the current preferred method.

### DIFF
--- a/docs/en/advanced/lazy-loading.md
+++ b/docs/en/advanced/lazy-loading.md
@@ -9,8 +9,8 @@ All we need to do is define our route components as async components:
 
 ``` js
 const Foo = resolve => {
-  // `require.ensure` is webpack's special syntax for a code-split point.
-  require.ensure(['./Foo.vue'], () => {
+  // The `import()` call sets a code-split point.
+  import('./Foo.vue').then(() => {
     resolve(require('./Foo.vue'))
   })
 }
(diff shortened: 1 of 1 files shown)


KaelWD:
It's actually even simpler than that. The first example can be converted to:

```JS
const Foo = () => import('./Foo.vue');
```

And instead of the mess of `require.ensure` and callback spaghetti trying to set a chunk name, you can just do this (webpack >= 2.4.0):

```JS
const Foo = () => import(/* webpackChunkName: 'group-foo' */'./Foo.vue');
const Bar = () => import(/* webpackChunkName: 'group-foo' */'./Bar.vue');
const Baz = () => import(/* webpackChunkName: 'group-foo' */'./Baz.vue');

rodet:
Nice! I can make the corresponding changes then. Thanks for the heads up on that.

rodet:
Thanks @yyx990803 for taking that up! I went up to other things and forgot to update this, but you made it much better than I could anyway.

PR #1485, opened 2017-06-01
append a letter missed in comment

Winterwrath:
correct “th” to “the” at line 23


posva:
Hey, thanks for the PR but the next branch is outdated. The fix should be against the dev branch

PR #1504, opened 2017-06-12
Add preloading support

Dadibom:
This PR adds an option to start calling beforeRouteEnter/beforeRouteUpdate on mousedown/touchstart, which reduces click-to-transition times by 100-300ms and makes fast applications become "instant". Simply add preload: true to your router config.

Because this relies on calling beforeRouteEnter/beforeRouteUpdate before we actually know if we want to transition or not (since the user can stop the click action from happening after mousedown), the callback in next will now be called in beforeRouteUpdate as it is in beforeRouteEnter. This will also allow the use of a single shared function for enter/update instead of two almost identical ones, avoiding duplication.

I have no idea how to unit test this so no test included, sorry.


callumacrae:
Have you got any data showing that the time difference between mousedown and the click event is 100-300ms? When I've tested similar things in the past with real users, it has been _much_ smaller.

(I made [futurelink](https://github.com/SamKnows/futurelink) which also works with vue-router [as a plugin](https://github.com/SamKnows/vue-futurelink))

Dadibom:
Yes, try this out on a real device: https://output.jsbin.com/yiyocavato

Dadibom:
sorry i thought you meant touchdown, mousedown to click is of course not as long but i use this in prod and switching preloading on/off makes a VERY noticeable difference

callumacrae:
It doesn't seem to be working on iOS: just saying "No time recorded".

Didn't Safari fix this [in 2015](https://webkit.org/blog/5610/more-responsive-tapping-on-ios/) in Chrome [forever ago](https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away), though? Or are we talking about different things?

Dadibom:
i'm talking about the delay between touch_start_ and click, not touch_end_ :)

Dadibom:
it appears ios doesn't support ES6, so just change let and const to var and you're good to go

maxmilton:
@callumacrae dude, vue-futurelink is amazing! I originally built something which had logic which triggered on hover but futurelink is much better. I'll be including it in my projects from now on.

P.s. Your SVG tools look sweet too! Def going to give them a try :)

P.p.s You guys should really gzip your static assets on samknows.com (sorry this comment it out of context of the github thread!)

Dadibom:
I can't even get the tests to run locally..

vue-router/src/history/abstract.js
  7:3  error  'index' is not defined  no-undef
  8:3  error  'stack' is not defined  no-undef
vue-router/src/history/base.js
  16:3  error  'router' is not defined                     no-undef
  17:3  error  'base' is not defined                       no-undef
  18:3  error  'current' is not defined                    no-undef
  19:3  error  'pending' is not defined                    no-undef
  20:3  error  'cb' is not defined                         no-undef

etc etc.

posva:
Thank you for this but I think this is worth keeping as an external plugin

PR #1512, opened 2017-06-16
support no fallback on browsers which do not support history API, fix #1456

### LINKED ISSUE #1456: SSR compatibility with IE9
### Version
2.5.3

### Reproduction link
[https://vue-hn.now.sh/top](https://vue-hn.now.sh/top)

### Steps to reproduce
open [Vue HN 2.0](https://vue-hn.now.sh/) in IE9

### What is expected?
Keep the original link

The url will change to https://vue-hn.now.sh/#/top automatically

---
Is there anyway to force enable 'history' mode in IE9 or just don't change url and treat it like IE9-?

JounQin:

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -30,6 +30,7 @@ declare type AfterNavigationHook = (to: Route, from: Route) => any
 declare type RouterOptions = {
   routes?: Array<RouteConfig>;
   mode?: string;
+  fallback?: boolean;
   base?: string;
   linkActiveClass?: string;
   parseQuery?: (query: string) => Object;
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,7 @@ export default class VueRouter {
     this.matcher = createMatcher(options.routes || [], this)
 
     let mode = options.mode || 'hash'
-    this.fallback = mode === 'history' && !supportsPushState
+    this.fallback = mode === 'history' && !supportsPushState && options.fallback !== false
     if (this.fallback) {
       mode = 'hash'
     }


posva:
Could you add a test for it in the `api.spec.js` file please?

JounQin:
@posva Sorry but it should only work on IE9, and it seems there is no spec for fallback to hash mode before, so actually I have trouble about how to write the test spec.

posva:
I meant that you can test that the mode is actually forced to 'history' even when there's no browser support for `pushState` (https://github.com/vuejs/vue-router/blob/dev/src/util/push-state.js#L6)

JounQin:
But it will always fallback to `abstract` on server/Node.js and it should not be broken.

https://github.com/vuejs/vue-router/blob/dev/src/index.js#L49

Then I pretend it as browser inside, I don't know does it meet your requirement.

posva:
Sorry if I was too pushy there, I meant to have an open discussion about it 😅 
Yes, it should default to abstract in Node. What you wrote looks pretty much like what I was thinking (without the `dist` files ofc, if you can remove that from the commit, that would be great 😄 )

JounQin:
But the spec indeed requires the *dist* files, because as I mentioned in comment: *"we do not use Router top above directly because `inBrowser` has been initialized inside"*.

yyx990803:
The dist file will be overwritten between builds... this won't work. Let's just remove the spec.

JounQin:
@yyx990803 I've reset the last commit.

PR #1518, opened 2017-06-17
Added documentation for replace in next function

ronald-d-rogers:
I was stuck on this one for a while because it was not in the documentation and want to add it for the benefit of others.


posva:
Thanks for reporting this. It's not quite the doc that should be added because the point is that the argument of the `next` function is a `RawLocation`. But the doc doesn't says make it clear about the types

ronald-d-rogers:
Yes you're probably right. I'm definitely not equal to the task of documenting RawLocation at the moment though. Perhaps an issue should be created instead.

PR #1526, opened 2017-06-20
warn empty route configuration in development

### LINKED ISSUE #1411: Warn if no routes are provided in the Router constructor options
### Version
2.5.3

### Reproduction link
[https://codepen.io/iDhruv22/pen/MmGXee?editors=1010](https://codepen.io/iDhruv22/pen/MmGXee?editors=1010)

### Steps to reproduce
```js
const Routes = [{
  path: '',
  component: Home
}, {
  path: 'users',
  component: User
}];

  routes: Routes
});
```

This is current, but now if I mistakenly write below code

```js
  Routes // Means Routes: Routes
});
```

So I don't get anything in browser, importantly not even a warning message(something like, routes key is required)

### What is expected?
There should be a warning message if one does not pass routes key.

Currently, there is no warning message.

cwang22:
closes #1411

### DIFF
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -19,6 +19,12 @@ export function createRouteMap (
   const pathMap: Dictionary<RouteRecord> = oldPathMap || Object.create(null)
   const nameMap: Dictionary<RouteRecord> = oldNameMap || Object.create(null)
 
+  if (process.env.NODE_ENV !== 'production') {
+    if (routes.length === 0) {
+      warn(false, `route configuration is empty.`)
+    }
+  }
+
   routes.forEach(route => {
     addRouteRecord(pathList, pathMap, nameMap, route)
   })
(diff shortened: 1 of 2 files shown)


yyx990803:
Not sure if this is necessary... technically you *can* create a router with no routes.

JounQin:
We have `addRoutes` method now, so is it really needed?

posva:
It's true that it's actually possible to create an empty router and dynamically add routes from the server.
The original idea was realising easily that doing

```js
new Router({
     Routes
})
```

would create no routes but I don't think adding a warning for people that dynamically retrieve the routes is a fair tradeoff. This is a very beginner problem that is absolutely unrelated to vue

This is my bad for marking the issue as FR and add the contribution welcome. I'm sorry @cwang22 I hope you will make another contribution another day 👍

PR #1568, opened 2017-07-05
Handle null values when comparing objects

### LINKED ISSUE #1566: `isObjectEqual` will throw exception when object contains null value
### Version
2.7.0

### Reproduction link
[https://jsfiddle.net/t8gpt9mk/](https://jsfiddle.net/t8gpt9mk/)

### Steps to reproduce
Click router link in page.

### What is expected?
Query in url changed.

Throw TypeError.

---
This code works well on vue-router 2.5.3, and bug introduce by pull request #1425.

posva:
Fix #1566

@callumacrae I couldn't see why the error was happening in circleci (infinite loading) and wanted to change some things, so I created a new PR. I hope this doesn't bother you.
To answer your question, the problem was there was a non-caught exception, so I tested that and equality as well. But I should test that as well, let me add a commit for that

### DIFF
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -74,6 +74,8 @@ export function isSameRoute (a: Route, b: ?Route): boolean {
 }
 
 function isObjectEqual (a = {}, b = {}): boolean {
+  // handle null value #1566
+  if (!a || !b) return a === b
   const aKeys = Object.keys(a)
   const bKeys = Object.keys(b)
   if (aKeys.length !== bKeys.length) {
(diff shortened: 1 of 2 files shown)


callumacrae:
What should happen with `{ foo: null }` and `{ foo: undefined }`? Should they be equal?

posva:
IMO no, but I also wrote that code 😆 
As a developer, you have to be consistent, I think, so either use `null` or `undefined` (I'd choose `null`)

javoski:
Maybe a shared util function `isObject` will be helpful, it also can be reused in many places in this repo.
```js
function isObject (val) {
  return val !== null && typeof val === 'object'
}
```

anxolin:
+1

posva:
ping @yyx990803 
@javoski I haven't checked, but yeah, maybe we could refactor some code

webrsb:
Resolve #1722

PR #1569, opened 2017-07-05
Fix #1188 add documentation for click

### LINKED ISSUE #1188: Add documentation doc about using @click with a router-link
Following  #650 . Sometimes we want to use a router-link component to benefit from its visited state or other things but manually handle the `@click` and it's necessary to use `native`. Even though it's said on the vue docs, it's better to repeat ourselves a bit

ATM there's no place on the docs where we mention the `@click` with a router-link so I think we could add a little note at the bottom of : https://router.vuejs.org/en/api/router-link.html

marshallshen:
#1188 
Update documentation for `@click.native` for `router-link`


posva:
Thanks for the proposal. This is something we'll add on a guide later on, but being something relative to Vue core, we'll do it very differently

Issue #1156, opened 2017-02-11
Dynamically add child routes to an existing route


miki2826:
I am building a large scale application with "mini applications" hosted within the application.
Each mini application is being developed by another team.
I would like to dynamically add nested child routes to an existing route to allow dynamic registration of each of the mini apps.

for example: 
```javascript
const routes = [
        {path: 'a', component: ComponentA, name: 'a'},
        {path: 'b', component: ComponentB, name: 'b'},
];

//MiniApp.js
export default Vue.extend({
    beforeCreate () {
        this.$router.addChildRoutes(this.$router.currentRoute.path, routes);
    },
    template: `
<div>
  <div class="page-host">
  <router-link to="{name: 'a'}">a</router-link>
  <router-link to="{name: 'b'}">b</router-link>
  <transition name="fade" mode="out-in">
    <keep-alive><router-view class="view"></router-view></keep-alive>
  </transition>
  </div>
</div>`,
});

//The app itself
Vue.use(Router);
import MiniApp from "./MiniApp";

const config = {
    linkActiveClass: 'active',
    scrollBehavior: () => ({x: 0, y: 0}),
    routes: [
        {path: '/mini', component: MiniApp, name: 'mini'}},
    ]
};
let router = new Router(config);
const vue = new Vue({
   router,
   template: `
<div>
  <div class="page-host">
  <router-link to="/mini">a</router-link>
  <transition name="fade" mode="out-in">
    <keep-alive><router-view class="view"></router-view></keep-alive>
  </transition>
  </div>
</div>`,
}).$mount('#app');
```

fnlctrl:
Since the `MiniApp` is loaded synchronously (not lazy loaded), why not simply expose its routes and use it when creating the router?
e.g.
```js
//MiniApp.js
export const MiniAppRoutes = [
        {path: 'a', component: ComponentA, name: 'a'},
        {path: 'b', component: ComponentB, name: 'b'},
];

export default Vue.extend({...});

//////////////////////////////////////////////////////////

//The app itself
Vue.use(Router);
import MiniApp, { MiniAppRoutes } from "./MiniApp";

const config = {
    routes: [
        {path: '/mini', component: MiniApp, name: 'mini', children: MiniAppRoutes }},
    ]
};
let router = new Router(config);
```

miki2826:
@fnlctrl you are right when the app is loaded synchronously, but this is just an example.
The real world example will require lazy loading.

fnlctrl:
[...]

PR #1576, opened 2017-07-08
[Feature]: Async Route Children

patrickhousley:
Async loading of children allows complex child route configurations to be asynchronously loaded into an existing route. This allows developers to more easily split their applications into lazy loaded modules through the likes of Webpack. Only those async children that are routed to are loaded, no unnecessary HTTP calls or processing of children. The only limitation I can see so far with this feature is that nested async named routes are not supported since they are not loaded into the nameMap until the routes are resolved and injected into the router instance.

Example configuration:
```ts
// src/routes.ts
export const routes: VueRouter.RouteConfig[] = [
  {
    path: '/lazy',
    children: async () => new Promise<Vue>(resolve => {
      require.ensure(
        [],
        async require => {
          resolve((require('./lazy/routes') as { routes: VueRouter.RouteConfig[] }).Routes);
        },
        'lazy'
      );
    })
  }
];

// /lazy/routes.ts
export const routes: VueRouter.RouteConfig[] = [
  {
    {
      path: '/one',
      component: OneComponent
    },
    {
      path: '/two',
      component: TwoComponent
    }
  }
];
```

### DIFF
--- a/examples/async-children/app.js
+++ b/examples/async-children/app.js
@@ -1,7 +1,7 @@
 import Vue from 'vue'
 import VueRouter from 'vue-router'
 
-Vue.use(VueRouter);
+Vue.use(VueRouter)
 
 const Home = { template: '<div>home</div>' }
 const Foo = { template: '<div>foo</div>' }
@@ -32,7 +32,7 @@ const router = new VueRouter({
   mode: 'history',
   base: __dirname,
   routes
-});
+})
 
 new Vue({
   router,
--- a/examples/async-children/app.js
+++ b/examples/async-children/app.js
@@ -1,0 +1,55 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+Vue.use(VueRouter);
+
+const Home = { template: '<div>home</div>' }
+const Foo = { template: '<div>foo</div>' }
+const Bar = { template: '<div>bar</div>' }
+
+const routes = [
+  { path: '/', component: Home },
+  { path: '/foo', component: Foo },
+  { path: '/bar', component: Bar },
+  {
+    path: '/basic',
+    component: { template: '<div><router-view class="async-view"></router-view></div>' },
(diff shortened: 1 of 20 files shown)


posva:
Why don't you just use the async component syntax? How can the totality of your children routes config be so big that it needs to be lazy loaded?
If you need to dynamically add complex children routes, you can use `addRoutes` already. For children [this fr](https://github.com/vuejs/vue-router/issues/1156) should also address your problem

patrickhousley:
I would not dismiss the idea that child routes could become large. Doing so would dismiss the idea that Vue and Vue Router could be used for large enterprise applications. That said, I can certainly try that FR. Is there work underway on it? If not I can try adding it. Thinking out loud, that FR would require:
- Update `create-route-map.js` `createRouteMap` to take a parent RouteRecord.
- Update `create-matcher.js` `addRoutes` to take a parent path. It will use `pathMap` to resolve the correct RouteRecord. It will use `nameMap` to resolve the correct named RouteRecord.
- Update `index.js` `addRoutes` to take a parent RouteRecord.
- Update `create-route-map.js` `compileRouteRegex` to match any and all children for routes that have no children. Deep linking to a child route that is not yet loaded into the router is a must have.
- Will this ensure all named and aliased routes are updated?
- Would it be better to update the route records in the async component `beforeRouteEnter` instead of `beforeCreate`. Or maybe a separate, new component guard method. How can we ensure the children are only added once?
- The router will need to reconsider the current route being loaded to ensure there are not additional components that should be loaded.

marcusamargi:
Thank you everyone for such quality work and awesome library!

Any chance this will get merged in soon?

greabock:
any progress on this?

lilong7676:
any progress on this?

EvanBurbidge:
Any idea when this feature will be introduced?

posva:
Thank you for the efforts put on this but I don't think this is worth pursuing given the points mentioned.
If you think otherwise, this is worth going through the process of an RFC: https://github.com/vuejs/rfcs

PR #1579, opened 2017-07-08
Also hint that you can use named routes in guards

mitar:


posva:
As noted at #1518, we should say somewhere that it's pretty much like using `router.push`

posva:
Added in a5ecfa1

PR #1586, opened 2017-07-11
Fix scroll when going back to initial route

### LINKED ISSUE #1585: scrollBehavior doesn't work for initial route in Firefox
### Version
2.7.0

### Reproduction link
[https://github.com/vuejs/vue-router/tree/dev/examples/scroll-behavior](https://github.com/vuejs/vue-router/tree/dev/examples/scroll-behavior)

### Steps to reproduce
Sorry, because this requires history navigation I can't give too great of a link. This manifests itself in the scroll-behavior example of the vue-router repo, though.

Basically, open a new tab to the scroll behavior example: http://localhost:8080/scroll-behavior/bar

Scroll just a little bit down and click on the "bar" link. It will correctly scroll to the top again. When you press back it will stay at the top, though, and not go to the previous scroll. By the way, for me this doesn't manifest in Chrome. I believe they are doing their own scrolling by default for history navigation through this: https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration

Firefox should manifest the issue, though.

### What is expected?
I expect it to be able to scroll properly even for initial route.

It doesn't scroll correctly.

---
The reason it happens is since the first request wasn't done through a pushState there is no state object stored. It needs the state object to store the scroll position.

This issue is probably known, however I think we may be able to fix it by doing a replaceState right away with a new state object, yeah?

hworld:
Should fix: https://github.com/vuejs/vue-router/issues/1585

I added a new e2e test that sets scrollRestoration to manual. Without the fix the test correctly fails.

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -7,6 +7,7 @@ import { getStateKey, setStateKey } from './push-state'
 const positionStore = Object.create(null)
 
 export function setupScroll () {
+  // Fix for #1585 for Firefox
   window.history.replaceState({ key: getStateKey() }, '')
   window.addEventListener('popstate', e => {
     saveScrollPosition()
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -7,6 +7,7 @@ import { getStateKey, setStateKey } from './push-state'
 const positionStore = Object.create(null)
 
 export function setupScroll () {
+  window.history.replaceState({ key: getStateKey() }, '')
   window.addEventListener('popstate', e => {
     saveScrollPosition()
     if (e.state && e.state.key) {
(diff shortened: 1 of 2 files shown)


posva (commented):
~I tested this fix on Firefox and the bug is still happening though...~ I thought you used dev but you created a new branch 👍

posva (approved):
Tested manually as well
@yyx990803 I cannot see anything else, can you take a look? (1 line bug fix for Firefox)

posva:
ping @yyx990803

djune:
@hworld 
Throw an exception use IE9 test 
`// Fix for #1585 for Firefox
  window.history.replaceState({ key: getStateKey() }, '');`

image

JounQin:
@djune It has been fixed in #1835 which has not been released.

PR #1591, opened 2017-07-13
Replace NPM by npm

MachinisteWeb:
Because we replace « NPM » by « npm » in french version (as it is named by official npm team : https://www.npmjs.com/), I propose to you this update.

Feel free to not accept it if you prefer kept the non official terms.


posva (approved):

posva:
Thanks

PR #1597, opened 2017-07-13
Doc EN : `props` vs "props" in `passing-props` for better comprehension.

MachinisteWeb:
Use coerce word in replacement of cast — « In computer science, type conversion, type casting, and type coercion are different ways of changing an entity of one data type into another. » JavaScript use coercion, not casting : https://en.wikipedia.org/wiki/Type_conversion

Better readibility between "props" as a concept and `props` as the option.

That is not totaly clear for me so feel free to correct that if I do mistake on some occurences.


docs/en/essentials/passing-props.md
```diff
@@ -59,7 +58,7 @@ const router = new VueRouter({
 
 ### Function mode
 
-You can create a function that returns props. This allows you to cast the parameter to another type, combine static values with route-based values, etc.
+You can create a function that returns props. This allows you to coerce the parameter to another type, combine static values with route-based values, etc.
```
posva: You can coerce to other values, but here we still want to keep _cast_

posva (approved):

docs/en/essentials/passing-props.md
MachinisteWeb: tl;dr => It's ok to merge like that for me :)

But

I'm not sure to well understand difference between « cast » and « coerce » :D

For me, « cast » is the action to explicitly convert a type value into another with some keyword usage.
And « coerce » is the action to implicitly convert a type value with the usage of a context or operand.

You means « with the usage of operand » is not "coercing" but is "casting" too? Even the conversion is « implicite » (no « int » or « string » keyword for example)?

Or I not understand the action behind « This allows you to cast the parameter to another type ».

So in JavaScript you coerce, in C# you cast (same for TypeScript no?)

PR #1603, opened 2017-07-14
feat(router): add 'replaceRoutes' method

### LINKED ISSUE #1234: Remove existing routes
Right now it is possible to dynamically add routes by calling `router.addRoutes([/* routes */])`. It would be nice if it was possible to also remove/replace routes on the fly. In my app user may define routes by himself (on the fly). I keep user configuration in the Vuex store, and when it is updated I want to completely replace whole router routes configuration (though replacing/removing single route would be also handful but I guess then it should also handle removing/replacing/adding child routes instead of only dealing with top-level routes, which would be quite complicated). I imagine that after replacing routes router would check if current path matches any of the new routes and navigated to it.

It could be for example method `router.replaceRoutes([/* routes */])` or `router.setRoutes` (in addition to method `router.addRoutes` which already exists).

Please note that this situation can be handled without VueRouter by setting one global route `/*` that will match every possible path, performing route matching inside this global route component (based on dynamic config stored in Vuex store for example) and dynamic rendering of component(s) that matched current path. However this is very ugly solution which looses all benefits of VueRouter, and handling child routes would be complicated.

In fact I wanted to handle it by patching VueRouter manually, but it seems that `createMatcher` (https://github.com/vuejs/vue-router/blob/dev/src/create-matcher.js#L16) does not expose `pathMap` and `nameMap` so it is not possible modify routes configuration in other way than by calling `router.addRoutes` (I guess it's intentional :).

alekbarszczewski:
'replaceRoutes' method replaces all routes in router

close #1234

### DIFF
--- a/docs/en/api/router-instance.md
+++ b/docs/en/api/router-instance.md
@@ -65,6 +65,12 @@
 
   Dynamically add more routes to the router. The argument must be an Array using the same route config format with the `routes` constructor option.
 
+- **router.replaceRoutes(routes)**
+
+  > 2.8.0+
+
+  Dynamically replace all routes in the router. The argument must be an Array using the same route config format with the `routes` constructor option.
+
 - **router.onReady(callback, [errorCallback])**
 
   > 2.2.0+
--- a/examples/index.html
+++ b/examples/index.html
@@ -25,6 +25,7 @@
     <li><a href="auth-flow">Auth Flow</a></li>
     <li><a href="discrete-components">Discrete Components</a></li>
     <li><a href="nested-router">Nested Routers</a></li>
+    <li><a href="replace-routes">Replace Routes</a></li>
   </ul>
 </body>
 </html>
--- a/examples/replace-routes/app.js
+++ b/examples/replace-routes/app.js
@@ -1,0 +1,103 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+Vue.use(VueRouter)
+
+const RouteX = { template: '<div>Route X</div>' }
+const RouteY = { template: '<div>Route Y</div>' }
+const RouteZ = { template: '<div>Route Z</div>' }
+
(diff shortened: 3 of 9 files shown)


alekbarszczewski:
Can someone help me? I don't know why but on my local machine tests are passing while they fail on CircleCI...

screen shot 2017-07-14 at 23 30 02

nickforddev:
@alekbarszczewski Did you look at CircleCI's log? I just glanced at it and there is one case failing.

```
 ✖ Testing if element <.view> contains text: "Route Y".  - expected "Route Y" but got: "Route X"
    at Object.replace routes (/home/ubuntu/vue-router/test/e2e/specs/replace-routes.js:30:15)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)

alekbarszczewski:
@nickforddesign yeah I saw it but I don't get it... When I run tests on my computer all tests pass successfuly. I spent some time to figure out why this is happening without any luck.

nickforddev:
https://github.com/vuejs/vue-router/blob/a85c6168e404cea3bd82d5365d7182f1070acbab/test/e2e/runner.js

Line 13: If CircleCI, use PhantomJS, otherwise use Chrome. So your tests are being executed differently locally than on CircleCI.

I am able to get the test to fail locally by changing the config to always use PhantomJS. Sorry I don't have more information, hopefully helps a little in debugging this.

lightswitch-sa:
Please please

rathga:
Please can this be fixed up and added...?  Need this feature

posva:
Thanks for the PR! After going through an RFC, we are going in a different direction. If you are still interested in implementing this, see https://github.com/vuejs/vue-router/issues/3087#issuecomment-626143022

PR #1615, opened 2017-07-18
Update history-mode.md

Alex-Sokolov:
Make link refer to page with language list


posva:
Thanks but I actually think it's better to point to the english docs from the english docs.

Issue #1640, opened 2017-07-26
preserving wrong key when popstate gives null state in history mode, resulted in buggy scroll behavior


sunziping2016:
### Version
2.7.0

### Reproduction link
[https://jsfiddle.net/sunziping2016/uc1uf9xa/](https://jsfiddle.net/sunziping2016/uc1uf9xa/)

### Steps to reproduce
Just run the script.

It has two pages. We start with page1 and then jump to page2, get back to page1, scroll down page1 and then forward to page2.

### What is expected?
page2 stay in the origin position.

### What is actually happening?
page2 scrolled to the same position as page1

---
Check the code at https://github.com/vuejs/vue-router/blob/dev/src/util/scroll.js line 12. Discard the key when getting null state may be better.

I'm still confused why it's designed like that. Why don't you just save the position information into state? These problems can soon be finished. In your implementation, a page refresh can easily clean your saved positions and cause many bugs.

sunziping2016:
And vue-router has occupied the history state. I have to store the position by myself. I'm really feeling headache about this. Have I missed something?

posva:
This actually happens because you are not starting from the right place because of you `*` redirect. If you look at the `to` and `from` content, you'll see that they're probably not what you expect (there's the `/_display/` route)

By simply pushing a route right after starting, it will solve your issue: https://jsfiddle.net/v8n6d2rc/

Not closing this because I think it should work nevertheless

sunziping2016:
Yeah, I see. I don't notice it before.

So `/_display/` represents the route temporary route?

However, I think there still exists the bug. A page refresh can destroy the positions saved in `scroll.js`, but html5 `history` is still preserved. So even changing the first `push` to `replace` may help little, as user can always go back to some unknown pages even before the `/_display` route.

https://jsfiddle.net/sunziping2016/kf50z3np/ Initially at page2, click `back` so as to get back to `page1`. Scroll down and click `forward`. Page2 then changes it position.

I've got a mouse with some navigation buttons, so I can do this even without changing your code.

It can be strange if that's the expected behavior.

posva:
Coming back at this, I looked at the last repro but I don't see what is wrong with the scroll, it seems to be restoring the right position. Maybe it was fixed in the meantime

PR #1641, opened 2017-07-26
temporary fixed issue #1640

sunziping2016:
Fixed issue #1640 

Uses session storage to save scroll positions and regenerates key when encountering an unknown page


sunziping2016:
Note: that this pull has failed on one e2e test scoll behavior. I just run build for my own usage

posva:
thanks for the pull request but we're not going to add a temporary fix. You can use this in your app if you want though

Issue #1149, opened 2017-02-09
Expose child routes on currentRoute object


JoshuaDoshua:
Any issue with adding the route children to the $router.currentRoute object? Adding routes dynamically might make this tricky, maybe vuex-router-sync would be more appropriate for this.
I'm not sure how to currently access nested routes without doing a find in the complete routes array.

Access to the current routes children would allow us to easily list child routes in a sub-navigation menu.

Ex:
```
computed: {
  children: () => this.$router.currentRoute.children
  }
}
...
<nav class="sub-nav">
  <router-link
    v-for="child in children"
    :to="{name: child.name}"
    >
    {{ child.meta.label }}
  </router-link>
</nav>

<router-view></router-view>
```

posva:
Closing in favour of #1156

JoshuaDoshua:
@posva These issues don't seem the same

I am requesting adding the predefined child routes to the currentRoute, not dynamically adding routes

posva:
Sorry for that. Adding all children routes to the currentRoute looks like a handy hack for your case but I'd rather just export them and import where necessary. @fnlctrl WDYT?

Currently, it should be possible to access the children routes records with:

```js
$route.matched[0].children
```

fnlctrl:
I think it seems reasonable for this use case.. Updated the title to be less ambiguous.

tmcdos:
So what is the current progress with this ? About half an year passed ...

NoraGithub:
Would there be any solution for this? It's such a long time.

rchipka:
👍

rannooo:
This would allow me to get rid of many ugly workarounds.

b4dnewz:
I really don't know why thing like this are not the default..

From my point of view it's pretty obvious that __currentRoute__ should match the real route as if found on __$router.options.routes__ or at least have the same properties..

Jo-Hansen:
Any news regarding this feature request? 😊

MrAdam:
I'd also like to give a vote for this feature.
Being able to dynamically generate navigation is great!

Gonzalo2310:
Any change, news or novelty of this subject?.

sysebert:
I would like this as well.

Given a route that has children, it would be nice to generate a nav knowing what children routes are available.

edit: A workaround is basically like this:

```js
computed: {
  routeChildren() {
    const matched = this.$route.matched;
    const routePath = matched[matched.length - 1].path;
[...]

PR #1650, opened 2017-07-27
Expose child routes on currentRoute object; add property for a CSS class on inactive routes

tmcdos:
It is not easy to get access to child routes when you need to render nested menus with `<router-link>`
Publishing **children** on the **$route.currentRoute** solves a real problem. The **parent** property is required so that the route can see its siblings. The **title** property is optional - it is useful instead of **name** when you want to name a route having a default child. I personally use **name** in `beforeRouteEnter/beforeRouteUpdate` to update the **document.title** in browser - and I need it on all routes (even with default child), so I had to add **title** property.

This closes issue #1149

There are non-trivial situations where playing with CSS selectors becomes too inconvenient. Consider the following markup:
```vue
    <div>
      <div class="tabs_account">
        <div class="tab_spacer">&nbsp;</div>
        <template v-for="page in $router.currentRoute.parent.children" v-if="page.can_show">
          <router-link tag="div" active-class="tab_active" inactive-class="tab" v-bind:to="page.path">{{ page.title }}</router-link>
          <div class="tab_spacer">&nbsp;</div>
        </template>
        <div class="tab_fill tab_spacer">&nbsp;</div>
      </div>
      <router-view></router-view>
    </div>
```
You can achieve the same effect like this
```vue
<router-link :class="{'router-link-active': $route.fullPath ==='/' || $route.fullPath === '/a', 'router-link-inactive': $route.fullPath !=='/' && $route.fullPath !== '/a'}" to="/a">
Text
</router-link>
```
but it looks hacky (you have to keep the PATH strings in sync between HTML and JavaScript code) while the proposed solution feels more elegant.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "lint": "eslint src examples",
     "test": "npm run lint && flow check && npm run test:unit && npm run test:e2e && npm run test:types",
     "test:unit": "jasmine JASMINE_CONFIG_PATH=test/unit/jasmine.json",
-    "test:e2e": "node test/e2e/runner.js",
+    "test:e2e": "xvfb-run node test/e2e/runner.js && rm -rf ./xvfb-run.* && rm -rf ./.com.google.* && rm -rf ./.org.chromium.*",
     "test:types": "tsc -p types/test",
     "docs": "cd docs && gitbook install && gitbook serve",
     "docs:deploy": "bash ./build/update-docs.sh",
--- a/test/e2e/nightwatch.config.js
+++ b/test/e2e/nightwatch.config.js
@@ -33,7 +33,10 @@ module.exports = {
       'desiredCapabilities': {
         'browserName': 'chrome',
         'javascriptEnabled': true,
-        'acceptSslCerts': true
+        'acceptSslCerts': true,
+        "chromeOptions": {
+          "args" : ["--no-sandbox"]
+        }      
       }
     },
 
--- a/test/unit/specs/children.spec.js
+++ b/test/unit/specs/children.spec.js
@@ -1,0 +1,78 @@
+import Vue from 'vue'
+import VueRouter from '../../../src/index'
+
+Vue.use(VueRouter)
+
+describe('currentRoute', () => {
+  describe('children[]', () => {
+    it('should work', () => {
+      const router = new VueRouter({
+        routes:
+        [
+          {
+            path: '/jobs',
+          },
+          {
+            path: '/setup',
+            children:
+            [
+              {
+                path: 'lists',
+              },
+              {
+                path: 'periods',
+              }
+            ]
+          }
+        ]
+      })
+  
+      router.push('/setup')
+  
+      expect(router.currentRoute.fullPath).toEqual('/setup')
+      expect(router.currentRoute.children.length).toEqual(2)
+      expect(router.currentRoute.children[0].path).toEqual('lists')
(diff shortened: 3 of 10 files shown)


posva:
Hey thanks for the PR 🙂 
If you want to add the feature, please, only add **that** feature, so don't add the `title` feature which is a personal thing. Also, please, don't add the `// TMCDOS` at the end of every line you added
Remember to add a test case for children

tmcdos:
Okay, I followed your recommendations:

- removed the `// TMCDOS` marks
- removed the custom **title** property
- added a unit test for **parent** and **children[]** properties
- added a Selenium test for **inactive-class** and **exact-inactive-class**
- fixed some settings to allow headless browser testing with ChromeDriver

Hopefully the PR is now acceptable.

package.json
```diff
@@ -28,7 +28,7 @@
     "lint": "eslint src examples",
     "test": "npm run lint && flow check && npm run test:unit && npm run test:e2e && npm run test:types",
     "test:unit": "jasmine JASMINE_CONFIG_PATH=test/unit/jasmine.json",
-    "test:e2e": "node test/e2e/runner.js",
+    "test:e2e": "xvfb-run node test/e2e/runner.js && rm -rf ./xvfb-run.* && rm -rf ./.com.google.* && rm -rf ./.org.chromium.*",
```
posva, line 31: Please let it as it was

test/e2e/nightwatch.config.js
```diff
@@ -33,7 +33,10 @@ module.exports = {
       'desiredCapabilities': {
         'browserName': 'chrome',
         'javascriptEnabled': true,
-        'acceptSslCerts': true
+        'acceptSslCerts': true,
```
posva, line 36: Can you remove this as well please

test/unit/specs/children.spec.js
```diff
@@ -0,0 +1,78 @@
+import Vue from 'vue'
+import VueRouter from '../../../src/index'
+
+Vue.use(VueRouter)
+
+describe('currentRoute', () => {
+  describe('children[]', () => {
```
posva, line 7: I don't think a describe is necessary here. a test with _should have children_ and another with _should have parent.children_ is enough

posva (changes requested):
I thought you were going to remove the other extra feature as well 😅 . Can you remove the inactive class feature from this PR, please?
Also, could you add the typings to the `flow/declarations.js` file, please?

Frondor:
I believe this is overkill, child routes should never be exposed in `currentRoute` object. A new method like `this.$route.getChildren()` should be better in my opinion.

tmcdos:
Child routes are exposed in every **route** object - just like the **parent** property. Why do you think a method is better than a property ?

Frondor:
Actually, no. They aren't.
The parent route is just one referenced object, and if you are exposing children for those route objects, you're wasting a lot of memory for something that most people don't need. That's why **in my opinion** it should be called/built on demand instead.

tmcdos:
Forgive me my profanity - perhaps using the original JavaScript objects as defined when instantiating the router instead of partially cloning them and exposing these partial clones would solve the problem with memory waste ?

kanatkubash:
Are not js objects just referenced not copied, it shouldnt be wasting memory.

posva:
Closing in favour of #1991

PR #1662, opened 2017-08-02
enhance hashHistory to support scrollBehavior

ariesjia:
if pushState is supported , enhance hashHistory to support scrollBehavior


posva:
Why would you use hash mode if pushState is supported? Why not using history mode?

PR #1677, opened 2017-08-11
fix #1675

### LINKED ISSUE #1675: when using history mode in IE9 and fallback:false, page will flash before refresh
### Version
2.7.0

### Reproduction link
the problematic  code is at src/history/base.js   line 64
 ```
 transitionTo (location: RawLocation, onComplete?: Function, onAbort?: Function) {
    const route = this.router.match(location, this.current)
    this.confirmTransition(route, () => {
      this.updateRoute(route) // problem here
      onComplete && onComplete(route)
      this.ensureURL()

      // fire ready cbs once
      if (!this.ready) {
        this.ready = true
        this.readyCbs.forEach(cb => { cb(route) })
      }
    }, err => {
      if (onAbort) {
        onAbort(err)
      }
      if (err && !this.ready) {
        this.ready = true
        this.readyErrorCbs.forEach(cb => { cb(err) })
      }
    })
  }
```
when in ie9 history mode and set fallback:false, after page inited, all of the router change should work like a regular a page jump(just like location.href or location.replace (code is at ./src/util/push-state.js line 53))
```
window.location[replace ? 'replace' : 'assign'](url)
```

thats say,  this.updateRoute(route) shouldn't be invoked when invoke history.push(except the time before page inited)

because this.updateRoute(route) will render the target component, and highlight the target router-link.

for example when I am at the home link('/'), then
1.  I click the nav tab help link('/help');
2. then the code invoke this.updateRoute(route);
3. the help nav is highlighted and the help page shown.;
4. after a while, the code run into ./src/util/push-state.js and invoke the window.location method, the page will jump to the target link. after server response, page refreshed.

the step 3 should not happen in this story, we should just jump, otherwise the page will flash: show the targe page,  highlight the target link, then after server response, do those task again. isn't it terrible?

### Steps to reproduce
ie9, history mode, fallback: false

### What is expected?
page won't show as if is't just a page jump(like location.href or location.replace)

page show before server response, lead to a page flash(like location.reload()), looks really terrible

---

### PR DESCRIPTION
**Pull request by hxlniada**
fix https://github.com/vuejs/vue-router/issues/1675

I add a shouldUpdateRoute method which is always true except in html5.js when the route **isready** and the browser **don't support history pushstate**(such as ie9)

because only history mode have the chance to disable the updateRoute method
which is force ie9 to history mode(fallback: false,  for server render)

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -193,8 +193,8 @@ export class History {
     })
   }
 
-  shuoldUpdateRoute ():boolean {
-    return true;
+  shuoldUpdateRoute (): boolean {
+    return true
   }
 }
 
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -64,7 +64,7 @@ export class History {
   transitionTo (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     const route = this.router.match(location, this.current)
     this.confirmTransition(route, () => {
-      if (this.shuoldUpdateRoute()) {
+      if (this.shouldUpdateRoute()) {
         this.updateRoute(route)
       }
       onComplete && onComplete(route)
@@ -193,7 +193,7 @@ export class History {
     })
   }
 
-  shuoldUpdateRoute (): boolean {
+  shouldUpdateRoute (): boolean {
     return true
   }
 }
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -64,7 +64,9 @@ export class History {
   transitionTo (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     const route = this.router.match(location, this.current)
     this.confirmTransition(route, () => {
-      this.updateRoute(route)
+      if (this.shuoldUpdateRoute()) {
+        this.updateRoute(route)
+      }
       onComplete && onComplete(route)
       this.ensureURL()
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by callumacrae** [comment]
shuoldUpdateRoute -> shouldUpdateRoute 😀

**Comment by hxlniada** [comment]
when will you merge this pr?

**Comment by hxlniada** [comment]
@LinusBorg @posva  hello, when will this pr be merged?
or is there some reason that this pr won't be merged?

I will use this feature soon.

**Comment by posva** [comment]
I still have to properly test this. It's a bit weird to never use the `fallback` option in the check

**Comment by hxlniada** [comment]
@posva 
in src/index.js line 45

```
this.fallback = mode === 'history' && !supportsPushState && options.fallback !== false
if (this.fallback) {
    mode = 'hash'
}
```
we know when we are in history mode(src/history/html5.js) and we are in a browser that doesn't support push state, then we know we have set the fallback option to false manully. if not, we will use src/history/hash.js .

so we don't need to check the fallback option, just need to check if we are in history mode and we are in a browser that doesn't support push state. 

the logical is: when we are in history mode, and the browser doesn't support push state, and the router has already initialized. we should just jump(window.location\[replace ? 'replace' : 'assign'\](url)).

thank you.

ps: I have tested this on my ie9 and it works well.

hxlniada:
@posva @LinusBorg hello, is this ok? should I do something? /(ㄒoㄒ)/~~

PR #1686, opened 2017-08-15
add this.$router information

connor11528:
Include information about the this.$router syntax. For many Vue.js projects vue-router will be configured in the main.js file and programmatic routes will be called from component files. This syntax is incredibly helpful and not obvious in the documentation. I found out about it because of [this thread](https://laracasts.com/discuss/channels/vue/redirect-from-method-in-vuejs-with-vue-router). Figured it should be accessible for all! Thank you

### DIFF
--- a/docs/en/essentials/named-routes.md
+++ b/docs/en/essentials/named-routes.md
@@ -28,4 +28,10 @@ router.push({ name: 'user', params: { userId: 123 }})
 
 In both cases, the router will navigate to the path `/user/123`.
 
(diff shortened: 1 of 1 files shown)


posva:
Hi, thanks a lot for your interest in improving the project. This information appears already in the previous doc page: https://router.vuejs.org/en/essentials/navigation.html
If added somewhere else, I think it should go in the _Getting Started_ section. I'll add it right away 🙂

connor11528:
Okay cool! Thank you, it is a great feature

posva:
it's at #1687 btw

PR #1687, opened 2017-08-15
docs($router): Note that router === this.$router

posva:


posva:
@chrisvfritz Can I have your review on this, please?
More precisely, what do you think about introducing those note right at the beginning. Is it too early?

connor11528:
The computed code block might not even be necessary though that is another great insight for getting params! Super helpful knowing we can refer to this.$router (a lot of the answers on the internets do things with this.$route as a legacy from vue 1.0)

posva:
@connor11528 That's a bit terrifying... It's a very basic feature. This means it's probably not clear in the docs 😞

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,27 @@ const app = new Vue({
...
+      // We will see what params are shortly
+      return this.$route.params
+    }
+  },
+  methods: {
+    someMethod () {
+      // Redirecting to another route
+      this.$router.push('/redirect')
+    }
+}
```
chrisvfritz: Looks like we're missing a `}` here.

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,27 @@ const app = new Vue({
...
+```
+
+Thoroughly the docs, we will often use the `router` instance. Keep in mind that `this.$router` is exactly the same as using `router`. The reason we use `this.$router` is because we don't want to import the router in every single component that needs to manipulate routing.
```
chrisvfritz: `Thoroughly` -> `Throughout`

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,27 @@ const app = new Vue({
...
+By injecting the router, we get access to it as `this.$router` as well as the current route as `this.$route` inside of any component:
+
+```js
+// Home.vue
+export default {
+  computed: {
+    params () {
+      // We will see what params are shortly
+      return this.$route.params
+    }
```
chrisvfritz: It might be nice to update this to a more real-world example. For example:

``` js
username () {
  // We will see what params are shortly
  return this.$route.params.username
}
```

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,27 @@ const app = new Vue({
...
+  },
+  methods: {
+    someMethod () {
+      // Redirecting to another route
+      this.$router.push('/redirect')
+    }
```
chrisvfritz: Again, a more real-world example could be helpful, so that readers can begin imagining when they might use `this.$router`. For example:

``` js
goBack () {
  window.history.length > 1
    ? this.$router.go(-1)
    : this.$router.push('/')
}
```

chrisvfritz: Another note: I'd probably avoid the word "redirect" here, as I think it implies an immediate transfer to another route in most cases - "navigate" or simple "go" are probably more appropriate.

docs/en/essentials/getting-started.md
chrisvfritz: Hmm, I'm actually wondering now if this paragraph is even necessary. I feel like the idea that `router === this.$router` is made clear by:

> By injecting the router, we get access to it as `this.$router`

chrisvfritz (changes requested):

docs/en/essentials/getting-started.md
posva: I wonder as well, let's share with the team

posva: It's the kind of text I'd put inside of a warning `p`

posva:
Thanks a lot for the nice review @chrisvfritz 🙂

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,29 @@ const app = new Vue({
...
 ```
 
+By injecting the router, we get access to it as `this.$router` as well as the current route as `this.$route` inside of any component:
+
+```js
+// Home.vue
+export default {
+  computed: {
+    username () {
+      // We will see what params are shortly
```
phanan: It should be 

```
// We will see what `params` is shortly
```

since `params` here refers to the attribute's name.

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,29 @@ const app = new Vue({
...
+        ? this.$router.go(-1)
+        : this.$router.push('/')
+    }
+  }
+}
+```
+
+Throughout the docs, we will often use the `router` instance. Keep in mind that `this.$router` is exactly the same as using `router`. The reason we use `this.$router` is because we don't want to import the router in every single component that needs to manipulate routing.
+
 You can also checkout this example [live](http://jsfiddle.net/yyx990803/xgrjzsup/).
```
phanan: It should be `check out` (with a space), as "checkout" is a noun.

posva: Ah, thanks, I'll fix other _checkout_s as well but in another commit

docs/en/essentials/getting-started.md
```diff
@@ -64,6 +64,29 @@ const app = new Vue({
...
+    }
+  }
+}
+```
+
+Throughout the docs, we will often use the `router` instance. Keep in mind that `this.$router` is exactly the same as using `router`. The reason we use `this.$router` is because we don't want to import the router in every single component that needs to manipulate routing.
+
+You can also check out this example [live](http://jsfiddle.net/yyx990803/xgrjzsup/).
 
 Notice that a `<router-link>` automatically gets the `.router-link-active` class when its target route is matched. You can learn more about it in its [API reference](../api/router-link.md).
```
phanan, line 93: Is it `router-link.md` or `router-link.html`?

posva: gitbook transforms it automatically

PR #1688, opened 2017-08-15
add beforeRouteUpdate use case to navigation doc

jsgv:
Adding this since I spent a considerable amount of time wondering why my component was not reloading.

Also, it looks like the documentation site is not being updated to current repo contents:
https://router.vuejs.org/en/essentials/navigation.html


posva:
Thanks a lot for your interest in improving the docs!

This is already said in a previous page of the docs: https://router.vuejs.org/en/essentials/dynamic-matching.html#reacting-to-params-changes so I adapted the note. I'm still unsure about this because it has nothing to deal with programmatic navigation but at the same time, a little reminder is always helpful

jsgv:
Ahh, missed that page! Sounds good 👍

Issue #1582, opened 2017-07-10
beforeRouteUpdate never calls next callback


bdoms:
### Version
2.7.0

### Reproduction link
[http://jsfiddle.net/znmLks2w/2/](http://jsfiddle.net/znmLks2w/2/)

### Steps to reproduce
Create a component with a `beforeRouteUpdate` guard that passes a callback to the `next` function.

### What is expected?
The callback function should be called.

### What is actually happening?
The callback is never called.

---
In my example I have both `beforeRouteEnter` and `beforeRouteUpdate` pointing to the exact same function. `beforeRouteEnter` works exactly like I expect, but `beforeRouteUpdate` does not. Click on the links to try to add one to the displayed value and you'll notice it works the first time but never any of the subsequent times.

If this behavior is purposeful then I would consider it a documentation bug - the docs need to be absolutely clear that `next` behaves differently depending on where you're calling it from. But even as I type this I'm realizing that if that's the case then it's terrible design. It should work the same way regardless.

posva:
The difference between both hooks is that enter doesn't have access to `this` so it needs the callback, while `update` has access to this. But I agree it should be more consistent, it's quite usual to have the same function in both hooks.

thatguystone:
I'd say that this is a bug, not an improvement. Consider when using `<router-view :key="...">`, it becomes impossible to access the new component, so you can't inject data. See: http://jsfiddle.net/thatguystone/nypp9jma/1/

amertum:
I've just step on this problem. Could it be explained in the documentation on the navigation-guards page ?

arnabrahman:
Just discovered this problem.

justin-hackin:
Also observing this problem

arnabrahman:
Why is this issue closed? is it solved?

LinusBorg:
I understand your confusion, but if you look at the documentation closesly, it states that passing a callback to `next()` is only necessary (and supported) in `beforeRouteEnter`, because we can't access `this` in that hook.

Both `beforeRouteUpdate` and `beforeRouteLeave` *do* have access to `this`, so the callback is unnecessary and therefore not supported. 

Solution: just use `this.yourdataProp = 'yourvalue'` to change data.

I will see if I can come up with a quick PR for the docs to stress this difference as this has come up a couple of times already.

bdoms:
[...]

PR #1704, opened 2017-08-25
Improve in-component guard examples

LinusBorg:
The section about in-component navigation guards regularly seems to confuse people. We have had several issues and topics on the forum where people tried to passing a callback to `next()` in `beforeRouteUpdate`, event though it has access to this and therefore doesn't support that callback.

Examples: 

* #1582 
* https://forum.vuejs.org/t/ssr-best-hookpoint-to-load-synchronous-component-specific-data/16491/5?u=linusborg

### DIFF
--- a/docs/en/advanced/navigation-guards.md
+++ b/docs/en/advanced/navigation-guards.md
@@ -116,7 +116,27 @@ beforeRouteEnter (to, from, next) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


posva (approved):

posva:
I think that we should support the callback everywhere for consistency though

PR #1702, opened 2017-08-25
Send props not defined on the route component in $attrs. Fixes #1695.

lbogdan:
Fixes https://github.com/vuejs/vue-router/issues/1695.

PR #1713, opened 2017-08-31
chore: fix invalid type annotation in typings test

ktsn:
The current typings test case has invalid type annotation, so the latest TypeScript compiler (v2.4.2) does not pass the test case. I've just fix the annotation to pass the test.

ref #1702

### DIFF
--- a/types/test/index.ts
+++ b/types/test/index.ts
@@ -163,7 +163,7 @@ router.go(-1);
 router.back();
 router.forward();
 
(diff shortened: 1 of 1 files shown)


posva (approved):

posva:
@ktsn Thanks 🖖 
Do you prefer to wait for someone's else review?

ktsn:
@posva No, since it's a tiny fix 😀

PR #1723, opened 2017-09-07
typo fix

FranckFreiburger:
Maybe a luckless copy-past


posva:
Thanks but it's actually right: `/users/1` (or any other value) gets matched

PR #1729, opened 2017-09-11
add configuration for Firebase hosting

ReneCode:
when hosting a vue-cli app on Firebase hosting, than the server has to be configured.

I've found a solution here:
https://github.com/vuejs-templates/pwa/issues/38

That works fine for me.

### DIFF
--- a/docs/en/essentials/history-mode.md
+++ b/docs/en/essentials/history-mode.md
@@ -102,6 +102,22 @@ rewrite {
 }
 ```
 
+#### Firebase hosting
+
+```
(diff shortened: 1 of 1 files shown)


posva (approved):

amaralDaniel:
So the issue with firebase is solved? I've deployed my PWA and the SW is not showing up.

posva:
That has nothing to deal with vue router though, I recommend you to use other means of communication where you are more likely to get answers and help 🙂

PR #1758, opened 2017-09-25
Make scrollBehavior work with transitions by handling a promise (fixes #1263)

### LINKED ISSUE #1263: Handle scrollBehaviour with transitions
> from #1262 

2.1.2 / 2.1.1

### Reproduction Link

http://jsfiddle.net/qw4hbgek/

### Problem
Clicking on Home, then on post1 will not scroll down because of the transition

### Feature request
We should at least provide a way of doing it userland in the docs. Ideally, scrollbehaviour should trigger once the new view is inserted into the dom

homerjam:
As stated in the discussion in #1263 this currently works by enabling an arbitrary delay to be used within the `scrollBehavior` handler. A better solution would use an event handler on the router view or at least allow us to get the transition duration. Unfortunately I don't think this is possible - AFAIK we don't have access to the view from within `scrollBehavior`.

### DIFF
--- a/examples/scroll-behavior/app.js
+++ b/examples/scroll-behavior/app.js
@@ -20,12 +20,13 @@ const Bar = {
 // - only available in html5 history mode
 // - defaults to no scroll behavior
 // - return false to prevent scroll
-const scrollBehavior = (to, from, savedPosition) => {
+const scrollBehavior = async function (to, from, savedPosition) {
   if (savedPosition) {
     // savedPosition is only available for popstate navigations.
     return savedPosition
   } else {
     const position = {}
+    let delay = 500
     // new navigation.
     // scroll to anchor by returning the selector
     if (to.hash) {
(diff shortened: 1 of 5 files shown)


homerjam:
Not sure why tests are working locally but not on CI, anyone?

blake-newman (changes requested):
@homerjam just rerun the tests to ensure it was not something CI end. Looks like it is a consistent failure. Could be something to do with the test implementation. 

Could you try something more specific that just a setTimeout. Perhaps use the transition API to hook into when the animation is complete. https://vuejs.org/v2/guide/transitions.html#JavaScript-Hooks

There are events you can hook into, that should give you the desired affect. https://vuejs.org/v2/api/#transition

homerjam:
Hi @blake-newman, I tracked down the test issue to me using async/await which didn't work in phantomjs.

Re. the transition API - that was one of my initial thoughts, however by my understanding we're unable to use it programatically from the `scrollBehavior` hook. But I think the user could maybe use it to emit an event which the `scrollBehavior` hook then listens for (I'm pretty new to Vue so assuming it's possible to emit/listen to events on the root/app).

homerjam:
Hey @pi0 any thoughts on this re. how it tackles https://github.com/nuxt/nuxt.js/issues/1376?

yyx990803:
Although this does not automatically solve the transition problem, it does expose a useful async primitive for users to tackle this problem in userland.

hackuun:
Will it be documented, how to use this primitive?

yyx990803:
See https://router.vuejs.org/en/advanced/scroll-behavior.html at bottom

PR #1759, opened 2017-09-25
Add event listeners to router links

therufa:
This PR should add the possibility to add any kind of event listeners to router-links, since they did not support this.


callumacrae:
Isn't this already possible with `.native`?

therufa:
@callumacrae wow, you're right. i totally ignored that modifier

PR #1761, opened 2017-09-26
[Doc EN]: `VueRouter` vs. Vue Router // [Doc FR]: Reflect this changes

MachinisteWeb:
Hi @posva!

Something goes wrong with `essentials/getting-started.md`, line 33.

The options are: 

```
If using a module system (e.g. via vue-cli), import `Vue` and `VueRouter` 
```

*Speaking about Class name*

or

```
If using a module system (e.g. via vue-cli), import Vue and Vue Router
```

*Speaking about Product name*.

We have chosen to speak about Product name because the real Class syntax is given just after with `Vue.use(VueRouter)`.

I suggest that change in this PR.

What your thoughts about that?

### DIFF
--- a/docs/en/essentials/getting-started.md
+++ b/docs/en/essentials/getting-started.md
@@ -30,7 +30,7 @@ Creating a Single-page Application with Vue + Vue Router is dead simple. With Vu
 ### JavaScript
 
 ``` js
-// 0. If using a module system (e.g. via vue-cli), import Vue and VueRouter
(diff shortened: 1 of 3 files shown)


posva:
You can also say _import vue and vue-router_ because it matches the package name but at the same time you `import Vue from 'vue'` and `import VueRouter from 'vue-router'`, so I think we can keep it the way it is

MachinisteWeb:
Ok. Maybe that's work in english because of « import » word. I remove my modification for english documentation :) !

PR #1801, opened 2017-10-12
Update SUMMARY.md

Alex-Sokolov:
Bump to version 3 :)


Jinjiang:
Speak to this issue, I suggest just remove the 2|3 version number in all `/docs/{lang}/SUMMARY.md` and also update `/docs/LANGS.md` but retain 0.x version number. Seems we won't split the 2.x and 3.x docs away right? 😄

PR #1804, opened 2017-10-12
Call scrollBehavior with app context

homerjam:
This allows us to listen for events in the scroll behavior hook, thus enabling us to resolve a promise returned by the hook upon triggering of some custom event (for example one emited on the $root by an after-leave transition event).

```js
// in scrollBehavior
      this.$root.$once('triggerScroll', () => {
        resolve(position)
      })

// in component

  methods: {
    afterLeave () {
      this.$root.$emit('triggerScroll')
    }
  }
```
```html
<!-- in template -->
      <transition name="fade" mode="out-in" @after-leave="afterLeave">
        <router-view class="view"></router-view>
      </transition>
```

Sorry for hastily done PR.

Ref: https://github.com/nuxt/nuxt.js/issues/1376#issuecomment-336182111

### DIFF
--- a/examples/scroll-behavior/app.js
+++ b/examples/scroll-behavior/app.js
@@ -25,23 +25,27 @@ const scrollBehavior = function (to, from, savedPosition) {
     // savedPosition is only available for popstate navigations.
     return savedPosition
   } else {
-    return new Promise(resolve => {
-      const position = {}
-      let delay = 500
-      // new navigation.
-      // scroll to anchor by returning the selector
-      if (to.hash) {
-        position.selector = to.hash
+    const position = {}
 
-        // specify offset of the element
-        if (to.hash === '#anchor2') {
-          position.offset = { y: 100 }
-        }
+    // scroll to anchor by returning the selector
+    if (to.hash) {
+      position.selector = to.hash
 
-        if (document.querySelector(to.hash)) {
-          delay = 0
-        }
+      // specify offset of the element
+      if (to.hash === '#anchor2') {
+        position.offset = { y: 100 }
       }
+
+      if (document.querySelector(to.hash)) {
+        return position
+      }
+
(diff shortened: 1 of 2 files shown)


homerjam:
Hey @yyx990803 any chance of squeezing this one in with the next patch release? 😃

yyx990803:
Maybe passing the instance in as the 4th argument is better.

homerjam:
I was wondering if app was the right context, I kind of followed nuxt conventions as that is how I've entered Vue land. Should I update the PR then?

If you don't mind me asking - what makes you think 4th arg is better?

Thanks!

posva:
No other functions in vue-router have the context set to the root, and it's a function declared on the router options so it wouldn't be intuitive to have the context set to the app. I also think an extra argument makes more sense

homerjam:
And should the extra argument be `app`? Bit of a Vue noob here so just
making sure this is best.

homerjam:
@posva @yyx990803 agreed that app doesn't make sense for the context, it should be router if anything.

Why is 4th argument better? I would expect the context to be set in router function/hooks etc. and it seems strange to pass router object to a function of itself.

homerjam:
Happy to change this guys, just let me know.

yyx990803:
`this` pointing to the router instance seems to make sense.

zamakkat:
Thank you so much for this fix it's a life saver! The only thing that worked flawlessly with Vue transitions and vue-router to make silky smooth fades from 1 page to another.

This fix has been around for quite some time without being released, wondering if there are plans for a new release of vue-router soon? It was quite confusing because there was an example in the live documentation but it didn't work in code because the latest official release is `3.0.1` from Oct 14, 2017 (almost 10 months ago!). I am currently resorting to use a manual build against the dev branch...

PR #1812, opened 2017-10-14
docs(router-link): add how to use @click with router-link

### LINKED ISSUE #1188: Add documentation doc about using @click with a router-link

Bilie:
closes #1188


posva:
Note: can be closed if router-link is transformed into a functional component #2021

posva:
At the end we are transforming the router-link into a functional one

PR #1814, opened 2017-10-15
feat(scroll): use manual scrollRestoration with scrollBehavior

Dzinlife:
Browser default to perform a scroll when popstate happens which cause a glitch especially when using vue-router async scroll.

`
window.history.scrollRestoration = 'manual';
`

Fix by adding this line.

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -8,7 +8,7 @@ const positionStore = Object.create(null)
 
 export function setupScroll () {
   // Prevent browser scroll behavior on History popstate
-  window.history.scrollRestoration = 'manual';
+  window.history.scrollRestoration = 'manual'
   // Fix for #1585 for Firefox
   window.history.replaceState({ key: getStateKey() }, '')
   window.addEventListener('popstate', e => {
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -7,6 +7,8 @@ import { getStateKey, setStateKey } from './push-state'
 const positionStore = Object.create(null)
 
 export function setupScroll () {
+  // Prevent browser scroll behavior on History popstate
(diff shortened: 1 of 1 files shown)


posva:
Could you provide a repro illustrating the problem, please? You can use [this fiddle](http://jsfiddle.net/posva/9r6xhqbp/) or provide an HTML page

Dzinlife:
http://jsfiddle.net/Lfeg5dtw/

I use a router transition so you can see it clearly.
Scroll to the bottom of /Foo and click the back button, the browser will perform an unexpected scroll.

yyx990803:
You can simply set this up in your own code - forcing it to always manual seems less flexible.

yyx990803:
On another thought - maybe this is a good default to have.

kchq:
Let's get this bad boy in here, eh?

src/util/scroll.js
```diff
@@ -7,6 +7,8 @@ import { getStateKey, setStateKey } from './push-state'
 const positionStore = Object.create(null)
 
 export function setupScroll () {
+  // Prevent browser scroll behavior on History popstate
+  window.history.scrollRestoration = 'manual'
```
posva:
```suggestion
  if ('scrollRestoration' in window.history) {
    window.history.scrollRestoration = 'manual'
  }
```

jellehak:
Not sure if this is a perfect solution. As when a user goes "outside the SPA" and uses the back button it scroll position isn't restored in "manual" mode. If you desire to restore the users scroll position use the following to set it back:

```
window.history.scrollRestoration = 'auto'
```

Perhaps this could be explained in the docs somewhere. Or passed as a config to the VueRouter.

PR #1818, opened 2017-10-16
fix(resolveAsyncComponents): return object when module is empty

clarkdo:
When `resolvedDef.default` is `undefined`, `match.components[key]` will be `undefined` instead of a object like `{_Ctor: {},  inject: {}}` in v2.7.
I thought this behavior may break compatibility with previous versions, pls feel free to review and discuss.

### DIFF
--- a/src/util/resolve-components.js
+++ b/src/util/resolve-components.js
@@ -20,7 +20,7 @@ export function resolveAsyncComponents (matched: Array<RouteRecord>): Function {
         pending++
 
         const resolve = once(resolvedDef => {
-          if (isESModule(resolvedDef)) {
+          if (isESModule(resolvedDef) && resolvedDef.default) {
             resolvedDef = resolvedDef.default
           }
           // save resolved on async factory in case it's used elsewhere


posva:
When do you find yourself with `undefined` as the default export of a component?

clarkdo:
@posva Thank you for the review!

In [nuxt error test](https://github.com/nuxt/nuxt.js/blob/dev/test/fixtures/error/pages/index.vue), there is an error component which includes a undefined variable `not_defined`.

After webpack building, component module `default` will be like:
```js
exports.default = {
  not_defined
};
```
At first time loading, component resolving will throw a [exception `not_defined is not defined`](https://github.com/nuxt/nuxt.js/blob/dev/test/error.test.js#L26), then `export.default` will be cached as `undefined`.

clarkdo:
It's just a test for error condition, not a common case, so if it's not necessary to consider, I can open a pr for nuxt to fix the test
@Atinux @pi0 If you have time, could you please take a look ?

yyx990803:
Even if we add this check, it's still likely an error that should be thrown explicitly.

PR #1835, opened 2017-10-19
fix #1834

### LINKED ISSUE #1834: scroll.js: replaceState() is not supported in IE9
### Version
3.0.1

IE9 throw the following error: `Object doesn't support this property or method "replaceState"`.
This was introduced here:
commit: https://github.com/vuejs/vue-router/commit/c166822cdc6cadd9f3c01085b877c86d74362b95
related issue: https://github.com/vuejs/vue-router/issues/1585

JounQin:
fix issue on browsers which do not support history API

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -5,15 +5,16 @@ import { History } from './base'
 import { cleanPath } from '../util/path'
 import { START } from '../util/route'
 import { setupScroll, handleScroll } from '../util/scroll'
-import { pushState, replaceState } from '../util/push-state'
+import { pushState, replaceState, supportsPushState } from '../util/push-state'
 
 export class HTML5History extends History {
   constructor (router: Router, base: ?string) {
     super(router, base)
 
     const expectScroll = router.options.scrollBehavior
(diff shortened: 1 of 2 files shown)


callumacrae:
What happens in browsers that don't support history API? Are they just stuck on the wrong URL now?

JounQin:
@callumacrae As the origin issue mentions, it throws an error on IE9: `Object doesn't support this property or method "replaceState"`.

src/history/html5.js
```diff
@@ -5,15 +5,16 @@ import { History } from './base'
...
 
 export class HTML5History extends History {
   constructor (router: Router, base: ?string) {
     super(router, base)
 
     const expectScroll = router.options.scrollBehavior
+    const supportsScroll = supportsPushState && expectScroll
 
-    if (expectScroll) {
+    if (supportsScroll) {
```
yyx990803, line 17: should also change to `supportScroll` on line 33

yyx990803 (changes requested):

src/history/html5.js
JounQin: Done.

PR #1837, opened 2017-10-20
Fix object notation error in navigation docs

wgpsutherland:


posva:
Thanks but it's not an error, it's an ES6 feature (object property shorthand)

PR #1851, opened 2017-10-27
Prevent banner from being removed by minifiers

scottbedard:
Trying to debug a webpack bundle splitting setup was made a bit trickier because it was hard to tell where this package was being included. Adding this should make the process a bit simpler.


posva (approved):
Thanks!

PR #1873, opened 2017-11-09
webpack version detail in the doc

TongDaDa:
Webpack began after version 2.6 + support import dynamic module is introduced.

Novice in practice, these version details easily out of a big pit, after about tool version, please consider carefully, maintain the Vue community together😄

detail see site -->  https://webpack.js.org/api/module-methods/#import-


posva:
Given that the syntax is now standard, I think we can keep this. Thank you for the proposal!

PR #1884, opened 2017-11-15
Update to nginx example

maxmckenzie:
If you have multiple `sites-enabled` on the same nginx server then setting `try_files $uri $uri/ /index.html;` causes a 500 error on the other sites if the first site root directory it tries is empty

https://serverfault.com/questions/505098/what-does-this-nginx-error-rewrite-or-internal-redirection-cycle-mean\?newreg\=96a16475fb24461b973b6c9bec2f4308

To support history fallback and make sure no other sites are affected  setting `try_files $uri /index.html $uri/ =404;` will route all requests to index. If it is not found, then it will serve a 404 error and avoid the infinite loop which causes other sites on the server to error and not load

### DIFF
--- a/docs/en/essentials/history-mode.md
+++ b/docs/en/essentials/history-mode.md
@@ -36,7 +36,7 @@ Not to worry: To fix the issue, all you need to do is add a simple catch-all fal
 
 ```nginx
 location / {
-  try_files $uri $uri/ /index.html;
+  try_files $uri /index.html $uri/ =404;
 }
 ```
 


hsxfjames:
Hello, I have a question: why `/index.html` goes front of `$uri/` ? Would `try_files $uri $uri/ /index.html =404;` makes more sense? Because `try_files` directive will follow the specified order, see [nginx docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#try_files):

> Checks the existence of files in the specified order and uses the first found file for request processing;
> ...

maxmckenzie:
@hsxfjames I'm not sure, They give an example in the link you provided `try_files $uri $uri/index.html $uri.html =404;`
which seems very similar to what you propose. 

If the order needs to be changed I guess that makes sense, if my memory serves me, history-api-fallback was involved in the need for the order I suggested.

hsxfjames:
@maxmckenzie Umm, I don't understand why you use nginx's `try_files` with history-api-fallback, but for spa, it always exists an `index.html` file at root. So it will always lose the chance to try `$uri/` since the `/index.html` has higher priority in your order. How do you think?

maxmckenzie:
Very good point. your right I do not use the history-api-fallback. I was thinking of something else. 

Sounds right, I'll try to get an example of this together somehow to show the issue.

It was specifically if you had multiple sites on one nginx instance/server. If one site (index.html) was not there it would fail to load the others

PR #1894, opened 2017-11-19
Typo fixes

miyanokomiya:
There are a few wrong variables.


posva:
Thanks! It was actually the other way around (based on en docs)

PR #1906, opened 2017-11-26
add the replace attribute in location for hooking in beforeEach

### LINKED ISSUE #1620: ability to check push vs replace in navigation guard
I am trying to create a single page mobile app with a push/pop/modal navigation. This navigation behavior is similar to the iOS SDK's navigation controller behavior.
To get this to work in my code I have added a new field(navigationType: push | pop | modal) in the params object. Then in router.afterEach I check for this new field.

It would be nice to have this be part of the vue router api. Considering the push/replace is already separate API.

usage might be:
router.afterEach((to, from) => {
  if(to.navigationType == 'push'){
  } else if(to.navigationType == 'replace'){
  }
}

wanyaxing:
Closes https://github.com/vuejs/vue-router/issues/1620

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -52,6 +52,12 @@ export class HTML5History extends History {
 
   replace (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     const { current: fromRoute } = this
+    if (typeof location === 'string'){
+      location = {path:location}
+    }
+    if (typeof location === 'object' && !location.replace){
+      location.replace = 1
+    }
     this.transitionTo(location, route => {
       replaceState(cleanPath(this.base + route.fullPath))
       handleScroll(this.router, route, fromRoute, false)
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -52,10 +52,10 @@ export class HTML5History extends History {
 
   replace (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     const { current: fromRoute } = this
-    if (typeof location === 'string'){
-      location = {path:location}
+    if (typeof location === 'string') {
+      location = { path: location }
     }
-    if (typeof location === 'object' && !location.replace){
+    if (typeof location === 'object' && !location.replace) {
       location.replace = 1
     }
     this.transitionTo(location, route => {
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -56,7 +56,7 @@ export class HTML5History extends History {
       location = { path: location }
     }
     if (typeof location === 'object' && !location.replace) {
-      location.replace = 1
+      (location: Object).replace = true
     }
     this.transitionTo(location, route => {
       replaceState(cleanPath(this.base + route.fullPath))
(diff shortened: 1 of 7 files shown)


wanyaxing:
e...
I should to study something about CircleCI now...
to be continued.

posva:
Hey, sorry for the delay. What are you trying to achieve? The issue you linked is closed. Why are you always setting replace to true if location is an object?

wanyaxing:
If somebody reset the path in beforeEach, the new path will be open as push type, actually, it will be puzzled if the original router path is replace type.

In this PR, the replace setting of location in the $router.replace() will be set  as true, so the developer can find the route's type in the beforeEach hooking.


image

posva:
Aaah, is the issue you're referring to #1620?

wanyaxing:
yes.

And the vue-router has supported the replace type checking in the route guards next()  in this commit: https://github.com/vuejs/vue-router/commit/d53ec3c29a5f05fcb3753b20e12ac37e5546ece2

This pr is an expand of that commit.

posva:
ok, could you add some tests to make sure the property is added in different scenarios but not others? (push/replace/calling it with next)

wanyaxing:
I have added an example for this commit.
Hopes to be useful.

luozhihua:
Is there any progress? I waiting for this PR to be merged for a long time.

choegyumin:
I am waiting for this PR to be merged too..

posva:
This will probably come along with other changes that would allow knowing what kind of navigation happened (back, forward, push, replace)

hxlniada:
@posva is there any plan to add these features? we are waiting for this feature too too too too too long. almost a year.

reckert:
Around one year after the last post, I would like to ask if this branch will be merged some day.

src/history/html5.js
```diff
@@ -52,6 +52,12 @@ export class HTML5History extends History {
 
   replace (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     const { current: fromRoute } = this
+    if (typeof location === 'string') {
+      location = { path: location }
+    }
+    if (typeof location === 'object' && !location.replace) {
+      (location: Object).replace = true
+    }
```
posva, line 60: Since location is an object at this point, we can directly set `replace` to true

```suggestion
  (location: Object).replace = true
```

src/util/location.js
```diff
@@ -52,10 +52,13 @@ export function normalizeLocation (
     hash = `#${hash}`
   }
 
+  const replace = next.replace || false
```
posva, line 55: no need to create a new variable. It can also be written as `!!next.replace`

src/util/route.js
```diff
@@ -23,6 +23,7 @@ export function createRoute (
     meta: (record && record.meta) || {},
     path: location.path || '/',
     hash: location.hash || '',
+    replace: location.replace || false,
```
posva, line 26:
```suggestion
    replace: !!location.replace,
```

posva (changes requested):
Thanks! There is no need to add a new e2e spec, we can add a test to the existing `basic` e2e test instead

PR #1918, opened 2017-12-01
Fix #1572 router-link in nested component behaves differently than one in a page-route

### LINKED ISSUE #1572: Should warn when params are not provided
### Version
2.7.0

### Reproduction link
[http://jsfiddle.net/forzLu80/30/](http://jsfiddle.net/forzLu80/30/)

### Steps to reproduce
1. Click on the button to generate a random param
2. Click on the first link (param / url stays the same)
3. Click on the second link (param / url changes)

### What is expected?
Both `router-link` components should behave exactly the same as their data is equal.

`router-link` components in nested components behave differently when param values are updated.

lzxb:
Fix #1572

The reason for bug is that the state of the parent component is modified

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -12,6 +12,7 @@
     <li><a href="nested-routes">Nested Routes</a></li>
     <li><a href="named-routes">Named Routes</a></li>
     <li><a href="named-views">Named Views</a></li>
+    <li><a href="nameed-change">Nameed Change</a></li>
     <li><a href="route-matching">Route Matching</a></li>
     <li><a href="active-links">Active Links</a></li>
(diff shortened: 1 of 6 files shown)


posva:
Sorry I didn't answer yet. The reason I didn't merge it yet is that using `{ name: 'app' }` without provided all of the params should not be allowed. I'm also concerned that navigating to a route with optional parameters that way could reuse parameters from the current route. I need a bit more time to investigate this one. See https://github.com/vuejs/vue-router/issues/1572#issuecomment-350041942

posva:
Thanks for the PR, at then end we are going with #2029

PR #1931, opened 2017-12-08
docs: add example of nested named routes

### LINKED ISSUE #1921: Document Named Views and Nested Routes
### Version
3.0.1

### Reproduction link
Full problem Example
[https://jsfiddle.net/oaxs8ek8/](https://jsfiddle.net/oaxs8ek8/)

### Steps to reproduce
I have created a Homepage with  2 Named Views called for example RouteA and RouteB. 
Now RouteA and RouteB have 2 other nested routes like so RouteA1 and RouteA2 - RouteB1 and RouteB2.

```
  routes: [
    {
      path: '/',
      components: {
        RouteA : Bar,
        RouteB: Baz
      }
    }
  ]
})
```

this work very well for the first route but then i cant continue the routing chain for the 2 differente route since i cant only do this: 

```
  routes: [
    {
      path: '/',
      components: {
        RouteA : Bar,
        RouteB: Baz
      },
      children: [
        {
          path: '' // Load RouteA1 on RouteA
          component: UserProfile
        },
        {
          path: '', // Load RouteB1 on RouteB
          component: UserPosts
        }
      ]
    }
    }
  ]
})
```

There is no way to specify the different route.

A solution could be enable this for example:

```
  routes: [
    {
      path: '/',
      components: {
        RouteA : {
                     path: ''
                     component: Bar,
                     children:{
                        ....
                     }
              },
        RouteB:  {
                     path: ''
                     component: Baz,
                     children:{
                        ....
                     }
      }
    }
  ]
})
```

### What is expected?
The router should allow nested route with named view router

it's not routeing correctly

posva:
Closes #1921

### DIFF
--- a/docs/en/essentials/named-views.md
+++ b/docs/en/essentials/named-views.md
@@ -50,7 +50,7 @@ It is possible to create complex layouts using named views with nested views. Wh
 
 **Note**: _Let's forget about how the HTML/CSS should look like to represent such layout and focus on the components used_
 
-The `<template>` section for `UserSettings` component in above layout would be something like this:
+The `<template>` section for `UserSettings` component in the above layout would look something like this:
 
 ```html
 <!-- UserSettings.vue -->
--- a/docs/en/essentials/named-views.md
+++ b/docs/en/essentials/named-views.md
@@ -60,7 +60,7 @@ While being quite complex, it is possible to combine named views with nested vie
 </div>
 ```
 
-_The nested view components are omitted here but you can find the full example at the end in the jsfiddle_
+_The nested view components are omitted here but you can find the full example [here](https://jsfiddle.net/posva/22wgksa3/)_
 
 Then you can achieve the layout above with this route configuration:
 
--- a/docs/en/essentials/named-views.md
+++ b/docs/en/essentials/named-views.md
@@ -30,7 +30,7 @@ A working demo of this example can be found [here](https://jsfiddle.net/posva/6d
 
 ## Nested Named Views
 
-While being quite complex, it is possible to combine named views with nested view. When doing so, you will also need to name nested `router-view` components used. Let's take a Settings panel example:
+It is possible to create complex layouts using named views with nested views. When doing so, you will also need to name nested `router-view` components used. Let's take a Settings panel example:
 
 ```
 /settings/emails                                       /settings/profile
@@ -50,6 +50,8 @@ While being quite complex, it is possible to combine named views with nested vie
 
 **Note**: _Let's forget about how the HTML/CSS should look like to represent such layout and focus on the components used_
 
+The `<template>` section for `UserSettings` component in above layout would be something like this:
+
 ```html
 <!-- UserSettings.vue -->
 <div>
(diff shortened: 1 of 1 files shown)


docs/en/essentials/named-views.md
```diff
@@ -27,3 +27,58 @@ const router = new VueRouter({
...
+<!-- UserSettings.vue -->
+<div>
+  <h1>User Settings</h1>
+  <NavBar/>
+  <router-view/>
+  <router-view name="helper"/>
+</div>
+```
+
+_The nested view components are omitted here but you can find the full example at the end in the jsfiddle_
```
Akryum: It's unfortunate that the jsfiddle link is at the end. Reader may ask "what jsfiddle?" here.
Maybe add the link here too?

docs/en/essentials/named-views.md
```diff
@@ -27,3 +27,58 @@ const router = new VueRouter({
...
 
 A working demo of this example can be found [here](https://jsfiddle.net/posva/6du90epg/).
+
+## Nested Named Views
+
+While being quite complex, it is possible to combine named views with nested view. When doing so, you will also need to name nested `router-view` components used. Let's take a Settings panel example:
+
+```
+/settings/emails                                       /settings/profile
++-----------------------------------+                  +------------------------------+
```
znck: We should put some image instead.

docs/en/essentials/named-views.md
```diff
@@ -27,3 +27,58 @@ const router = new VueRouter({
 ```
 
 A working demo of this example can be found [here](https://jsfiddle.net/posva/6du90epg/).
+
+## Nested Named Views
+
+While being quite complex, it is possible to combine named views with nested view. When doing so, you will also need to name nested `router-view` components used. Let's take a Settings panel example:
```
znck: Re-wording.
---
It is possible to create complex layouts using named views with nested views. For example, let us see following example; a settings panel.

docs/en/essentials/named-views.md
```diff
@@ -27,3 +27,58 @@ const router = new VueRouter({
...
+| |     |                         | |                  | |     | UserProfilePreview | |
+| +-----+-------------------------+ |                  | +-----+--------------------+ |
++-----------------------------------+                  +------------------------------+
+```
+
+- `Nav` is just a regular compnonent
+- `UserSettings` is the view comopnent
+- `UserEmailsSubscriptions`, `UserProfile`, `UserProfilePreview` are nested view components
+
+**Note**: _Let's forget about how the HTML/CSS should look like to represent such layout and focus on the components used_
```
znck, line 51: Re-wording.
---
The `<template>` section for `UserSettings` component in above layout would be something like this:

docs/en/essentials/named-views.md
```diff
@@ -27,3 +27,58 @@ const router = new VueRouter({
...
+<!-- UserSettings.vue -->
+<div>
+  <h1>User Settings</h1>
+  <NavBar/>
+  <router-view/>
+  <router-view name="helper"/>
+</div>
+```
+
+_The nested view components are omitted here but you can find the full example [here](https://jsfiddle.net/posva/22wgksa3/)_
```
znck: Re-wording.
---
You can find the complete source code for above example [here](https://jsfiddle.net/posva/22wgksa3/).

docs/en/essentials/named-views.md
posva: Let's keep this for the moment. An Image would be indeed better

posva:
@znck Thanks! I updated the relevant parts the best I could

docs/en/essentials/named-views.md
posva: I added this just below

znck (approved):

PR #1941, opened 2017-12-15
Adding children and parent to currentRoute

ianwalter:
Also adding .vscode (IDE settings file) to .gitignore

@posva I'd like to have the feature #1650 was trying to add but it looks like there are too many unrelated changes in that PR and it hasn't been updated with your comments for a little while so I decided to try re-work it here. Thank you for your consideration.


src/util/route.js
```diff
@@ -21,6 +21,8 @@ export function createRoute (
   const route: Route = {
     name: location.name || (record && record.name),
     meta: (record && record.meta) || {},
+    children: (record && record.children) || [],
+    parent: (record && record.parent) || {},
```
posva: Isn't it better to have `null` or `undefined` if the parent does not exist?

posva:
Sorry for taking so much time in getting this a read!
Thank you very much for the contribution!

src/util/route.js
ianwalter: Yes, I think it would be. I'll update it.

posva:
Opening #1991 to replace this

PR #1952, opened 2017-12-23
If the head of the hash value is numeric, an error will occur

hiro0218:
If the head of the hash value is numeric, the queryselector fails.

`#a1234` is  good, however `#12345` is  not good.

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -111,13 +111,15 @@ function isNumber (v: any): boolean {
 function scrollToPosition (shouldScroll, position) {
   const isObject = typeof shouldScroll === 'object'
   if (isObject && typeof shouldScroll.selector === 'string') {
-    const el = document.querySelector(shouldScroll.selector)
-    if (el) {
-      let offset = shouldScroll.offset && typeof shouldScroll.offset === 'object' ? shouldScroll.offset : {}
-      offset = normalizeOffset(offset)
-      position = getElementPosition(el, offset)
-    } else if (isValidPosition(shouldScroll)) {
-      position = normalizePosition(shouldScroll)
+    if (!isFinite(shouldScroll.selector.substr(1, 1))) {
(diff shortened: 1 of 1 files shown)


posva:
Hey thanks for the pr but an id should not start with a number as an id-selector cannot start with a number by-spec

hiro0218:
@posva Thank you for confirmation.

> Hey thanks for the pr but an id should not start with a number as an id-selector cannot start with a number by-spec

I agree with you.
but is not mentioned in w3c reference to "hash id should not start with a number".
https://www.w3.org/TR/2011/WD-html5-20110525/urls.html#parsing-urls

I think it can be assumed that numbers are specified.

posva:
the error doesn't come from vue-router, it comes from chrome

hiro0218:
Ah, I see. Thanks.

tolgap:
@posva I'm running into this error as well in Chrome. I understand that only Chrome throws this exception, but I'm using Nuxt.js, and routes are generated for me. I don't have access to the routes, so it would be nice if this error was eventually caught inside vue-router. In other browsers, it simply doesn't scroll.

posva:
You still have to make sure your id is valid and therefore doesn't start with a number 🙂 
It should be feasible but that's more of a Nuxt issue than vue-router

tolgap:
@posva Thanks for your speedy response.

The querySelector is ran as soon as you enter the page `#` by vue-router. It doesn't stop my visitors from simply append a non-existent ID in the URL fragment, which will result in a 500 error from within vue-router. When/Where do I have the chance to check whether this ID exists? Is there something I can hook into before the querySelector call is done so I can do a try-catch?

posva:
I'm sorry I don't know how your app is structured nor how ids are set. You should ask for help on the forums but may sure to provide some code if you want people to help you easily

PR #1980, opened 2018-01-11
Add typing for async components

CKGrafico:
I think that this typing is necessary to work with async components.

`component: () => import('./cities.component')`

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -73,7 +73,7 @@ export interface PathToRegexpOptions {
 export interface RouteConfig {
   path: string;
   name?: string;
-  component?: Component;
+  component?: () => Promise<any> | Component;
   components?: Dictionary<Component>;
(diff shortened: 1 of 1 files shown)


types/router.d.ts
```diff
@@ -73,7 +73,7 @@ export interface PathToRegexpOptions {
 export interface RouteConfig {
   path: string;
   name?: string;
-  component?: Component;
+  component?: () => Promise<any> | Component;
```
posva: shouldn't we use `Promise<Component>`?

posva:
It looks like async component are already handled at https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L3 . Not sure if it supports plain Promises or just calls from dynamic `imports`

edit: Seems to work correctly, what were you trying to fix?

CKGrafico:
Probably my fault

PR #1988, opened 2018-01-12
fix(errors): added check if IE11 then use hashchange event listener F…

### LINKED ISSUE #1849: Manual change of hash into the URL doesn't trigger the route in IE11
### Version
3.0.1

### Reproduction link
[https://codepen.io/Aymkdn/pen/MENZPx](https://codepen.io/Aymkdn/pen/MENZPx)

### Steps to reproduce
1. Use IE11 
2. Open codepen at https://codepen.io/Aymkdn/pen/MENZPx 
3. Click on "Open demo"
4. Click on "Go to foo" or click on "Go to bar" to see it works correctly
5. Manually change the hash into the URL from the IE11 address bar (for example replace "#/foo" with "#/bar")

### What is expected?
The component to be loaded based on the current hash into the URL

The changes are not detected under IE11. It works with Firefox.

---
When the page is already opened and I want to copy/paste a link into the same page which has the same url but a different path/hash

margokopli:
fix #1849

IE11 does not fire the popstate event when the URL's hash value changes. Changed, in IE11 case, from `propstate` event listener to `hashchange`.

### DIFF
--- a/src/history/hash.js
+++ b/src/history/hash.js
@@ -28,7 +28,10 @@ export class HashHistory extends History {
       setupScroll()
     }
 
-    window.addEventListener(supportsPushState ? 'popstate' : 'hashchange', () => {
+    const ua = window.navigator.userAgent
+    const ie11 = ua.indexOf('rv:11.0') !== -1
+
+    window.addEventListener(supportsPushState && !ie11 ? 'popstate' : 'hashchange', () => {
       const current = this.current
       if (!ensureSlash()) {
         return


posva:
Hey, thanks for the contribution. I will need to run the test suit against IE11 to make sure no other features are being broken

pacop:
Some news? I am having this very same problem

thepill:
Im also facing the same problem - would be great if this solves it!

SniperSister:
Same issue here.

hugywin:
IE10 also has this problem

N2ET:
any progress?

Krabi:
@posva Any results? Almost one yer ago you promissed to run tests.

asomervell:
@posva how is this not merged yet? Was it resolved already somewhere else? @aclifford85

PR #1995, opened 2018-01-14
fix(match): use pathMatch for the param of * routes

### LINKED ISSUE #1994: "'0' is not a valid attribute name" when navigating to a wildcard route with "props: true"
### Version
3.0.1

### Reproduction link
[https://codesandbox.io/s/oo6x710k39](https://codesandbox.io/s/oo6x710k39)

### Steps to reproduce
 - click 'foo'
 - click 'bar'
 - observe the error
 - click 'home'
 - note that the page title no longer updates

### What is expected?
Either give a console warning, or silently ignore the numeric props

The following error is thrown, breaking the whole page
image

---
In the actual app everything is wrapped in another dynamic route, which is what has `props: true`

posva:
Fix #1994
Combining an asterisk route (*) with props: true would result in an
error because the name of the param would be the number 0 making it an
invalid attribute + the fact that vue passes props as attributes
if they're not declared as props

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -189,7 +189,8 @@ function matchRoute (
     const key = regex.keys[i - 1]
     const val = typeof m[i] === 'string' ? decodeURIComponent(m[i]) : m[i]
     if (key) {
-      params[key.name] = val
+      // Fix #1994: using * with props: true generates a param named 0
+      params[key.name || 'fullPath'] = val
     }
   }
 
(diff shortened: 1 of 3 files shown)


skjnldsv:
Bump?
I'm experiencing the same issue :/
Can I help @posva ?

Alternative fix: `path: '/:index(index.php/)?settings/users',`
Thanks to @danxuliu

``` js
export default new Router({
	mode: 'history',
	routes: [
		{
			path: '/(index.php/)?settings/users',
			component: Users,
			props: true,
			name: 'users',
			children: [
				{
					path: ':selectedGroup',
					name: 'group',
					component: Users
				}
			]
		}
	]
});
```

PR #2006, opened 2018-01-16
Update app.js

Alanhliu:
what the difference between static and attrs


Alanhliu:
what the difference between static and attrs?

posva:
I think there's some refactoring necessary in the tests and one of them may be reduntant. Thanks for bringing my attention to it. I'll give it a look soon when refactoring tests :)

Issue #1858, opened 2017-10-31
Rewriting URLs


ClassicOldSong:
### What problem does this feature solve?
Much like redirect, but a programmable alias route. Such as when I'm visiting '/me', actually I'm visiting '/user/:username'.

### What does the proposed API look like?
{
  path: '/alias',
  aliasTo: to => 'path/I/would/like/to/be/aliased/to'
}

posva:
After thinking about it, I think it's a good idea to have something like this. Having `alias` and `aliasTo` can be confusing though. It could also be just `alias` with no component option but I still think it's confusing:

```js
{ path: '/me', alias: '/users/posva' }
```

Maybe something called `replace` instead

ClassicOldSong:
Maybe `rewrite` could be better?

mattbryanswan:
`rewrite` is exactly right, I think. This appears to fit my use case as well. For example, I need a detail page for events, so the no-brainer url looks like this:

`/event/:id`

However, that'll be a little ugly. Ideally, I want it to be something more like this:

`/event/name-of-event`

Currently, I cannot do this using aliases unless I explicitly create every route and every alias. Having dynamic aliases to be able to connect :key to :id would be fantastic.

SerikDM:
Please check if this solution works for you:
https://github.com/SerikDM/vue-router

description:
add '*' route, in 'beforeEnter' handler find out what route you want to redirect to, pass object with 'ninjaPath'(=to.path) property to 'next()'
i.e.:
next({ path : '/user/123', ninjaPath : '/user/me' })

brophdawg11:
This would be an awesome addition and I think would solve the use-case I'm working with now.  Our e-commerce app has a "catch-all" `/:slug` route that serves a few different types of pages (i.e., `/dresses` generates a tiled view of products in the dresses category whereas `/fall-fashion` may lead to a CMS-driven page of content regarding fall fashions.  This is all driven by `route.params.slug` and we look up what type of page it should be.

Our homepage is just a CMS-driven route for the slug `homepage`, so `/homepage` is the same as what we want rendered for `/`.  I thought the following would do the trick but it doesn't seem like it actually populates the proper `route.params.slug` field as I had hoped it would when going to `/`:

```
route = [{
    path: '/homepage',
    alias: '/',
    component: CatchAll,
}, {
    path: '/:slug',
    component: CatchAll,
)];
```

[...]

PR #2013, opened 2018-01-20
Mask for urls

SerikDM:
Related issue #1858 

Adding support to hide real path by 'ninja' mask
Pass object with 'ninjaPath' property to 'push' or 'replace' method, 'ninjaPath' will be displayed in address bar.

Use case:
Dynamic short urls like 'domain/fancyShortUrl' which will be caught in '*' route and lead to 'domain/user/userId' or 'domain/note/noteId' based on response from server, but URL in address bar will remain 'domain/fancyShortUrl'

ps. I'm beginner.


posva:
Thank you for the PR, please see https://github.com/vuejs/vue-router/issues/1858#issuecomment-626546063

PR #2025, opened 2018-01-29
chore: replace extend method with vue built-in

QingWei-Li:
To reduce the package size. 😂

reference `router-link` 
https://github.com/vuejs/vue-router/blob/63e00c1e8bd406c84fc1ead7cf10712f88d0d3ee/src/components/link.js#L91-L94

### DIFF
--- a/src/components/view.js
+++ b/src/components/view.js
@@ -1,4 +1,5 @@
 import { warn } from '../util/warn'
+import { _Vue } from '../install'
 
 export default {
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

posva:
Forgot about this one, at the end we prefer having this util in vue router to not depend on vue utils for such a small utility function as it could change in the future breaking vue-router for every user updating only vue

Issue #2038, opened 2018-01-31
Throw exception or provide info message when route's component is null


mortyccp:
### What problem does this feature solve?
Currently, route's component can be null and it give no information and error. Provide information or raising error can help users to catch they have some import issues when they use
```
import * as components from './components'

export default [
  {
    path: '/accounts',
    component: components.AccountsListView,
    name: 'accountsListView'
  },
  {
    path: '/accounts/create',
    component: components.CreateEditAccounts,
    name: 'createAccounts'
  }
]
```

So if `components.CreateEditAccounts` is null, they will know what went wrong easier.

### What does the proposed API look like?
Raise error when component is null or log info message.

posva:
for anybody picking this up: make sure to check it only in dev mode

serge1peshcoff:
I will try to implement this the next week if nobody would be faster than me

posva:
There's already a pr 😉

posva:
Note to myself, this could be solved through typings

deckar01:
This issue also produces unexpected behavior when routes without components have child routes. See #2908.

@posva The MR for this is 18 months old. Can we close that and remove the `has PR` label?

agm1984:
+1 I like the idea of type safety here. I fouynd this searching unrelated-ish stuff, but I feel like ES Lint might pick this error up. `eslint-plugin-import` and/or `eslint-import-resolver-alias` might supply an IDE with enough girth and mustard.

amitkeinan9:
I see that no one had touched this issue since March. May I try?

Robert-Kovalcuk:
is it still possible to solve the issue?

raj-adi00:
Is it still open. I would like to work on it??

yash07-bit:
is it still open . I would like to work on it?

PR #2041, opened 2018-02-02
Throw error when component is undefined or null (#2038)

AlbertLucianto:
Add assertion when a route component is null or undefined, a fix for issue #2038.

Really appreciate for a review.

### DIFF
--- a/test/unit/specs/create-map.spec.js
+++ b/test/unit/specs/create-map.spec.js
@@ -80,9 +80,8 @@ describe('Creating Route Map', function () {
 
   it('in development, throws if component is null or undefined', function () {
     process.env.NODE_ENV = 'development'
-    expect(() => {
-      maps = createRouteMap([{ path: '/' }])
-    }).toThrowError(/route config "component" for path/)
+    maps = createRouteMap([{ path: '/' }])
+    expect(console.warn).toHaveBeenCalled()
   })
 
   it('in production, it has not logged this warning', function () {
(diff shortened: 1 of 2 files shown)


posva (changes requested):
I simplified the condition but you still need to add more checks because, the component option can actually be absent if other options are present (like `redirect`)

AlbertLucianto:
Oh sorry I missed that. Will try to work on that.

posva:
Keep in mind you can run the tests locally too, it will save you some time 😄

AlbertLucianto:
Just now, I've tried to trace through what might be throwing error when it shouldn't. Then I found in `Route Matching` that the routes does not have any components nor any other options like redirections, aliases. So I think it is probably better to just warn instead of asserting error.

You might want to review it again. Anyway thanks for the guide.

test/unit/specs/create-map.spec.js
```diff
@@ -78,6 +78,12 @@ describe('Creating Route Map', function () {
     }).toThrowError(/"path" is required/)
   })
 
+  it('in development, throws if component is null or undefined', function () {
+    process.env.NODE_ENV = 'development'
+    maps = createRouteMap([{ path: '/' }])
+    expect(console.warn).toHaveBeenCalled()
+  })
+
```
vinibrsl, line 86: I would create an spec similar to this one, but to test that this is not affecting production env, with `not.toHaveBeenCalled` 👍

vinibrsl (changes requested):

PR #2043, opened 2018-02-02
Update docs to expose redirectedFrom on $route object

therealpecus:
This PR describes the `vm.$route.redirectedFrom` property that is made available if the current route is the result of a redirect.
It also updates the docs to clarify that props and meta fields are ignored for redirect route definitions.

### DIFF
--- a/docs/en/api/route-object.md
+++ b/docs/en/api/route-object.md
@@ -90,4 +90,4 @@ The route object can be found in multiple places:
 
 - **$route.redirectedFrom**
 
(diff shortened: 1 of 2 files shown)


therealpecus:
A note on CI checks: no code being tested has been changed (this is a fork from develop), but I cannot relaunch the tests. Locally, all tests run fine.

posva:
Hey, thanks for the PR and sorry for the delay.
I removed the note about ignoring props and meta field because I'm not sure what you're scenario you're talking about, do you mind sharing a fiddle showcasing what you mean so we can improve the docs?

posva (approved):

PR #2049, opened 2018-02-06
Update app.js

ghost:
Fix (comment) for transition issue

### DIFF
--- a/examples/transitions/app.js
+++ b/examples/transitions/app.js
@@ -19,6 +19,8 @@ const Parent = {
     }
   },
   beforeRouteUpdate (to, from, next) {
+    // expect to.path and from.path to be without / in the end, becouse /example/word will give the same result as /example/
+    // or add this filter to.path.split('/').filter(n => !!n).length and from.path.split('/').filter(n => !!n).length 
     const toDepth = to.path.split('/').length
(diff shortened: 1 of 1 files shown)


posva:
What are you trying to improve in the examples. What wasn't clear?

ghost:
```
let to.path = '/someroute/'; // if we are coming back from deeper route, check this trailing slash
let from.path = '/someroute/something';
const toDepth = to.path.split('/').length // 3 | ["", "example", ""]
const fromDepth = from.path.split('/').length // 3 | ["", "example", "something"]
this.transitionName = toDepth < fromDepth ? 'slide-right' : 'slide-left' // will be false as a result slide-left, but we expect it to be slide-right
```

posva:
I'd appreciate a bit more communication with sentences 😅

ghost:
Sorry :) My english isn't perfect :) I'm working right now on mobile vue app. And had an issue with this example. Becouse when you go back from a child route to parent you will have a trailing slash and this example logic will not work. This why I wrote this comment.

posva:
No worries!
Thanks for the PR but I think it will be better to note in docs that the trailing slash is removed by default but that it can be controlled by passing options to path-to-regexp (`strict: true`) 🙂

PR #2050, opened 2018-02-07
Query/Location can be an array of strings. (fix #1932)

### LINKED ISSUE #1932: Type declaration incomplete for Location.query
### Version
2.7.0

### Reproduction link
[https://jsfiddle.net/50wL7mdz/82013/](https://jsfiddle.net/50wL7mdz/82013/) (This doesn't actually work, but shows the code you'd need locally to repro the bug)

### Steps to reproduce
0. Note that JSFiddle doesn't appear to give Typescript compiler errors, so the repro linked will have to be ported into a local Typescript Vue project.

1. In a Typescript Vue project:
```
router.push({ query: { array: ["1", "2", "3"] }})
```

### What is expected?
This should compile without error 

You get an error:
```
 error TS2345: Argument of type '{ name: string; query: QueryParams; }' is not assignable to parameter of type 'RawLocation'.
  Type '{ name: string; query: QueryParams; }' is not assignable to type 'Location'.
    Types of property 'query' are incompatible.
      Type 'QueryParams' is not assignable to type 'Dictionary<string> | undefined'.
        Type 'QueryParams' is not assignable to type 'Dictionary<string>'.
          Index signatures are incompatible.
            Type 'string | string[] | undefined' is not assignable to type 'string'.
              Type 'undefined' is not assignable to type 'string'.
```

zigomir:
Query and Location can be arrays as can be deduced from a spec [here](https://github.com/vuejs/vue-router/blob/dev/test/unit/specs/query.spec.js#L6-L8) and [here](https://github.com/vuejs/vue-router/blob/dev/test/unit/specs/query.spec.js#L19-L20).

I guess for custom parsing user needs to provide own types and override ones coming from router somehow.

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -107,7 +107,7 @@ export interface Location {
   name?: string;
   path?: string;
   hash?: string;
-  query?: Dictionary<string>;
+  query?: Dictionary<string | string[]>;
   params?: Dictionary<string>;
   append?: boolean;
   replace?: boolean;
@@ -117,7 +117,7 @@ export interface Route {
   path: string;
   name?: string;
   hash: string;
-  query: Dictionary<string>;
+  query: Dictionary<string | string[]>;
   params: Dictionary<string>;
   fullPath: string;
   matched: RouteRecord[];
--- a/types/test/index.ts
+++ b/types/test/index.ts
@@ -105,7 +105,7 @@ const route: Route = router.currentRoute;
 const path: string = route.path;
 const name: string | undefined = route.name;
 const hash: string = route.hash;
-const query: string = route.query["foo"];
+const query: string | string[] = route.query["foo"];
 const params: string = route.params["bar"];
 const fullPath: string = route.fullPath;
 const redirectedFrom: string | undefined = route.redirectedFrom;
@@ -148,7 +148,8 @@ router.push({
     foo: "foo"
   },
   query: {
-    bar: "bar"
+    bar: "bar",
+    foo: ["foo1", "foo2"]
   },
   hash: "hash"
 });


ktsn (approved):
👍

posva:
Oh thanks!

Kingwl:
it's very helpful to me ! (after release

tarunmangukiya:
Any plans when this is going to be released? It's been more than 6 months since last release!

fuzzzerd:
I see that this was released in the last day or two and I've started getting typescript errors. 

With v3.0.1 I was able to do the following:

` const myId: string = this.$route.query.queryStringVariable;` 

after updating to v3.0.2 the above code throws an error that type `string | string[]` is un-assignable to type `string`. 

Was my example above working by accident? It seems that the only way to get this to work is to rewrite it as 

` const myId: string = this.$route.query.queryStringVariable as string;` 

which is sub-optimal, but perhaps I'm doing something wrong?

dinvlad:
Same issue here, our code has been assuming `this.$route.query.queryStringVariable` is a `string` for many months now, but looks like this is a breaking change?

Seems like for safety, we should change the code to `Array.isArray(queryStringVariable) ? queryStringVariable[0] : queryStringVariable` now.

Alternatively, had the breaking change been to switch to `string[]` everywhere now, then we could simply use `queryStringVariable[0]` from now on.

zigomir:
It was working by coincidence – see `console.logs` below in this [runkit example](https://runkit.com/zigomir/5bfcb9860cdb0800121974f8) – where I copied default `parseQuery` function (without flow types) from [vue-router's query.js file](https://github.com/vuejs/vue-router/blob/dev/src/util/query.js#L37-L63).

Otherwise I agree it could be nicer to always return same type `string[]`.

dinvlad:
Yeah, makes sense. I can also see that the "new" definition (`string | string[]`) is only breaking for TypeScript users, and provides a compatibility bridge for vanilla JS. Perhaps we could change to `string[]` only in 4.0?

dinvlad:
I think the string vs array was not clearly documented in https://github.com/vuejs/vue-router/tree/dev/docs/api#route-object-properties (there, `query` is only vaguely defined as an `Object` of key-value pairs).

fuzzzerd:
Is splashing `Array.isArray(queryStringVariable) ? queryStringVariable[0] : queryStringVariable;` around all of the views where query string parameters are processed considered _the_ typescript solution until v4 is released with only `string[]`?

khuguet-tfg:
I'm experiencing this issue as well.  It feels like [semantic versioning ](https://semver.org/)rules were not followed here.  A patch release that contains a type change on a public API is normally not allowed.  I know I'm pointing out the obvious, but just venting my frustration a bit.

zigomir:
I'd argue that public API wasn't changed – typings were wrong in the first place. But I can see your frustration – I felt it before, because typings were wrong.

fuzzzerd:
I suppose that's what we get when we're adding Typescript in after the fact; but it never the less is a breaking change since definitions are provided. 

In any event, lets all look forward to v4 where this will get a correction.

Akimyou:
Disappointment. I have to change my code about the query, such as:

```
const { addShowItem } = this.$route.query
if (addShowItem && typeof addShowItem === 'string') {
}
```
Hope can fix it in the future.

PR #2060, opened 2018-02-13
Navigate when a new route is added while on a wildcard

### LINKED ISSUE #1862: using addRoutes() when you already have a wildcard route fails to detect if the CURRENT url matches one of the new routes
### Version
3.0.1

### Reproduction link
[https://jsfiddle.net/c4jkw82o/5/](https://jsfiddle.net/c4jkw82o/5/)

### Steps to reproduce
The JSFiddle shows the broken case.

Given an app/router that includes a wildcard '*' route, calling addRoutes() to dynamically add more routes does not correctly render the page if the CURRENT url matches one of the newly added routes.

If you comment out the wildcard '*' route and re-run then it works as expected.

### What is expected?
The "Test" component should be rendered

The "NotFound" component is rendered

---
BACKGROUND...

I am attempting to use webpack code splitting to keep all of my test related code in a different bundle, that includes vue components that are only used in the tests. I wanted to try to dynamically add my "/test" route from within this separate bundle and discovered that serving up that page showed my 404 not found component instead of my (top level) test component.

DEBUGGING NOTES...

From what I can tell, when the router is first constructed - in my main application.js bundle - it checks against the current URL "/test" and matches the wildcard route as $router.currentRoute. Later when the separate test.js bundle adds the new routes it checks again, this time it matches the new test route and tries to transition, but during confirmTransition() it decides that isSameRoute() is true (both matches had path=/test) and does not update.

... after checking isSameRoute() there is actually a small attempt to see if the routes were dynamically changed...

    // in the case the route map has been dynamically appended to
    route.matched.length === current.matched.length

.. but in this case this check also (incorrectly) returns true because while the length is the same (1) the details of route.matched[0] are different - current.matched is pointing to the wildcard route but route.matched is pointing to the new test route.

P.S. THANK YOU FOR ALL YOUR WORK ON VUE - it is awesome!

SteffenDE:
As I am also relying on having dynamically added routes and a fallback 404 route, I tried fixing #1862 by checking not only if the matched routes length changed but also if the matched route itself changed:
`route.matched[0] === current.matched[0]`

Is this a valid way to fix this issue? Or do we have to check if there is more than one matched route?

Closes #1862

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -100,7 +100,8 @@ export class History {
     if (
       isSameRoute(route, current) &&
       // in the case the route map has been dynamically appended to
-      route.matched.length === current.matched.length
+      route.matched.length === current.matched.length &&
+      route.matched[0] === current.matched[0]
     ) {
       this.ensureURL()
       return abort()


posva:
Fixed with #3228

Issue #2090, opened 2018-03-05
Update apache-example for history-mode (documentation)


doppelreim:
### What problem does this feature solve?
For HTML5 History Mode, there are some example server-configurations in the documentation.

See https://github.com/vuejs/vue-router/blob/dev/docs/en/essentials/history-mode.md#apache

I think the apache-example could be simplified by using https://httpd.apache.org/docs/2.4/mod/mod_dir.html#fallbackresource instead of `mod_rewrite`?

### What does the proposed API look like?
```
FallbackResource /index.html
```

posva:
Hey, thanks for pointing it out. I don't use Apache `FallbackResource` looks indeed simpler. We are open to PRs for this kind of information but I will also like to have more people review (Will probably post the link to the PR in discord or twitter). Could you please open the PR instead to open up discussion? Thanks!

christianmarzian:
When deploying in subfolders it works for me with an additional '.'
`FallbackResource ./index.html`

PR #2091, opened 2018-03-05
change example for apache-config

doppelreim:
This is a PR for issue #2090

Instead of using `mod_rewrite`, it may be possible to use `FallbackResource` (see https://httpd.apache.org/docs/2.4/mod/mod_dir.html#fallbackresource).  
That would be more concise and would not depend on `mod_rewrite`.

Disclaimer: I just started with vue and am not an expert with apache :)

### DIFF
--- a/docs/en/essentials/history-mode.md
+++ b/docs/en/essentials/history-mode.md
@@ -22,8 +22,16 @@ Not to worry: To fix the issue, all you need to do is add a simple catch-all fal
 #### Apache
 
 ```apache
-FallbackResource /index.html
+<IfModule mod_rewrite.c>
+  RewriteEngine On
+  RewriteBase /
+  RewriteRule ^index\.html$ - [L]
+  RewriteCond %{REQUEST_FILENAME} !-f
+  RewriteCond %{REQUEST_FILENAME} !-d
+  RewriteRule . /index.html [L]
+</IfModule>
 ```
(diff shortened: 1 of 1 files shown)


phanan:
I don't think this is necessary. `mod_rewrite` is the de-facto standard – can't imagine any Apache installation without it, when `FallbackResource` requires ^2.4 AFAIK and is arguably much less known. We _may_ add it as a note, which I'm against still, simply because our doc isn't a web server's doc.

posva:
I was also sceptical about adding it to the dos for the same reason but I think it does help people so often that it is worth having it in the official docs. I personally think a very short note with a clickable link but no code about it is useful. Something along the lines of _If you are using Apache ... you can also use [FallbackResource](#fake-link)_

phanan:
OK I was wrong. `FallbackResource` was available _before_ 2.4 :P Still, my point stands that it's a much less known directive.

doppelreim:
I image having *any* kind of hint would be nice. People just starting out (with vue or developing fullstack in general) would be less likely to trip over that caveat.

Whether that's `mod_rewrite` or `FallbackResource`... ultimately, I do not care. Just wanted to give an "fyi" :)

posva:
I think it's a good _FYI_, so let's add it as an extra note 😄

doppelreim:
I reverted the example and added an fyi-notice :)

PR #2140, opened 2018-04-03
Add section about handling loading state to lazy-loading doc

### LINKED ISSUE #1534: question of advanced async components and vue-router
following sentence are copied from the doc
>Note that when used as a route component in `vue-router`, these properties will be ignored because 
 async components are resolved upfront before the route navigation happens. You also need to use 
 `vue-router` 2.4.0+ if you wish to use the above syntax for route components.

I wonder what it truly means?

if these properties will be ignored, why i use it for route components?

chrisvfritz:
Partially resolves https://github.com/vuejs/vuejs.org/issues/1534.

### DIFF
--- a/docs/en/advanced/lazy-loading.md
+++ b/docs/en/advanced/lazy-loading.md
@@ -45,3 +45,52 @@ const Baz = () => import(/* webpackChunkName: "group-foo" */ './Baz.vue')
 ```
 
 webpack will group any async module with the same chunk name into the same async chunk.
+
+### Handling Loading State
+
+Async component factories can also [return an object](https://vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State), specifying alternative components to use when loading or if an error occurs. This is a great way to improve user experience on slow or intermittent connections.
+
+Unfortunately, routes can currently only resolve to a single component in Vue Router. So to make this work, you'll need a helper function to create an intermediary component, like this:
+
+```js
+function lazyLoadView (AsyncView) {
+  const AsyncHandler = () => ({
+    component: AsyncView,
+    // A component to use while the component is loading.
+    loading: require('./Loading.vue').default,
+    // A fallback component in case the timeout is exceeded
+    // when loading the component.
+    error: require('./Timeout.vue').default,
+    // Delay before showing the loading component.
+    // Default: 200 (milliseconds).
+    delay: 400,
+    // Time before giving up trying to load the component.
+    // Default: Infinity (milliseconds).
+    timeout: 10000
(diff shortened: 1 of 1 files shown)


docs/en/advanced/lazy-loading.md
```diff
@@ -45,3 +45,52 @@ const Baz = () => import(/* webpackChunkName: "group-foo" */ './Baz.vue')
...
+    error: require('./Timeout.vue').default,
+    // Delay before showing the loading component.
+    // Default: 200 (milliseconds).
+    delay: 400,
+    // Time before giving up trying to load the component.
+    // Default: Infinity (milliseconds).
+    timeout: 10000
+  })
+
+  return Promise.resolve({
```
posva, line 72: Doesn't it work without returning a promise and just doing `component: lazyLoadView(...)`

chrisvfritz: If the `import(...)` isn't behind a function to be evaluated later, won't all the chunks be downloaded at startup time, regardless of whether the user needs them?

blake-newman: As what I know only on execution its fetched (always forget how it works)

blake-newman: So ies needs to be in a function, so it is only fetched when needed

posva (approved):

franzwilhelm (approved):

docs/en/advanced/lazy-loading.md
```diff
@@ -45,3 +45,52 @@ const Baz = () => import(/* webpackChunkName: "group-foo" */ './Baz.vue')
...
 webpack will group any async module with the same chunk name into the same async chunk.
+
+### Handling Loading State
+
+Async component factories can also [return an object](https://vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State), specifying alternative components to use when loading or if an error occurs. This is a great way to improve user experience on slow or intermittent connections.
+
+Unfortunately, routes can currently only resolve to a single component in Vue Router. So to make this work, you'll need a helper function to create an intermediary component, like this:
+
+```js
+function lazyLoadView (AsyncView) {
```
Mighty683, line 56: This solution cause to disable any navigation guards in loaded component

CurryChou777:
My version of VueRouter is 3.4.6，it can't loading page.But i try by 3.3.2, it's OK.i can't find the reason

Glandos:
See https://github.com/vuejs/v2.vuejs.org/issues/1534#issuecomment-1785408650 for another solution.

PR #2190, opened 2018-05-09
Update navigation-guards.md

trainiac:
Be consistent with referring to guard function names.


trainiac:
Not sure what the process is for translations, but first let me know if this is approved and I can do what is next for updating the rest of the langs

posva (approved):
Thanks!

Issue #2040, opened 2018-02-02
Option to match exactPath that ignores query parameters and/or hashes


zamakkat:
### Use case
I have a nested view of pages that looks like this:
```
Foo (links to /pages/foo)
|-- Bar (links to /pages/foo/bar)
```
I need to highlight the current selected page (and only 1 item can be highlighted). So if the user is on `/pages/foo`, only `Foo` should be highlighted. If the user is on `/pages/foo/bar`, only `Bar` should be highlighted. However, this page also has pagination that when selecting the 2nd page, navigates to `/pages/foo/bar?page=2`.

### Context
vue-router currently has the option to do an `exact` match of the path upon which the router will automatically add the class specified in the `linkExactActiveClass` option (which is `router-link-exact-active` by default).

In the above example, the exact match will match `/pages/foo/bar` but will not match `/pages/foo/bar?page=2` since it includes the query params in the comparison. However, if I don't use the `exact` match then both `Foo` and `Bar` will be both shown as `active`.

My solution has been to add a custom check to match only the path but would be nice for `vue-router` to have the option by default.

This request is for the ability to do an exact match but only on the path, not the whole URL, i.e. the `exactPath` match will ignore any query params in the URL. 

### What does the proposed API look like?
The current implementation of `exact` uses the `exact` directive to add the `router-link-exact-active` class.

A proposed solution would be to add the `exact-path` directive which will in term add the `router-link-exact-path-active` class.

vishnunair17:
I was searching for a way to do this and luckily found this issue, should probably update the docs https://router.vuejs.org/en/api/router-link.html . Such an useful feature.

throrin19:
any news about that ? I have the same problem here

Djaler:
I need this feature too. Can we get it? I do not think it is very difficult

busheezy:
This would be great to have.

solonik1:
Same issue:(

posva:
The api around adding active classes is starting to get bloated so I will take more time to think about this and proposals are welcome too

For the moment, adding a `name` to the route and comparing the name with current route is the workaround: `:class="$route.name === 'some-name' && 'router-link-active'"`

posva:
The RFC is up at https://github.com/vuejs/rfcs/pull/35

posva:
I created a new RFC, way simpler that eliminates the problem by design, but is also a breaking change: https://github.com/vuejs/rfcs/pull/136

I would appreciate some feedback!

[...]

PR #2070, opened 2018-02-20
Implements issue #2040 <router-link .. exact-path />

samgre:
I've needed this feature many times too. Since there is an issue open, I decided to give it a try solving it even though it's not yet tagged "contribution welcome".

Please take a look and see if it fits to vue-router.

PR #2193, opened 2018-05-12
feat(link): add 'exact-path' matching option

emanuelmutschlechner:
Improves unfinished pull request #2070 and closes issue #2040


posva:
Thanks! I need to check some things first about the API. I'll come back at you 😅

cvharris:
It looks like this was not touched upon in awhile, could it be merged soon?

One great use case for this is when individual components are not aware of the current route or need to add extra logic to become aware of it. Since `<router-link>` already has logic to determine if the current path matches the link, this PR would help tailor the exact matching to the important path and ignore query params and hashes.

Hoping it makes it in soon! 😅

JuroOravec:
Any updates on this?

posva:
Thanks! This was partially merged at cbb552c946af4ff0946e5e2f0091c8877850befd

PR #2217, opened 2018-05-25
Fix a broken link that throws a 404

jumpalottahigh:


posva:
Thanks, it wasn't working so it set the link to `./guide/`

PR #2251, opened 2018-06-11
Russian docs: translate menu item "Languages"

lex111:


Alex-Sokolov (approved):
@kazupon done

P.S.: But I think that this menu item better to keep in english everywhere. If somebody accidentally change locale — this will be easy to find where switch back

Issue #2125, opened 2018-03-23
Hash mode places # at incorrect location in URL if current query parameters exist on page load


chasegiunta:
### Version
3.0.1

### Reproduction link
[https://codesandbox.io/s/zlvz65rplp](https://codesandbox.io/s/zlvz65rplp)

### Steps to reproduce
Navigate to myapp.com?foo=bar

### What is expected?
Hash is appended before query parameters, to get the final result of myapp.com/#/?foo=bar

### What is actually happening?
Hash is placed after parameters myapp.com?foo=bar#/ (also making the query parameters inaccessible with `this.$route.query`)

---
Note on codesandbox link - unable to place query parameter for initial URL in editor preview - please remove `/#/` & append `?foo=bar`

Running across this in an instance where user authentication is done outside our app, but client/user identifier is passed to our app in query parameter upon authorization.

posva:
Next time, please take the time to complete the repro, it would have taken you 2 minutes.

chasegiunta:
@posva Would have done it if I knew how to technically make it happen with codesandbox!

> Note on codesandbox link - unable to place query parameter for initial URL in editor preview - please remove /#/ & append ?foo=bar

posva:
What I meant is completing the code : adding one route, injecting the router and displaying query params.

Firgot to say, a workaround for this is setting up a route rewrite in your server

chasegiunta:
@posva Respectfully, I'm not sure we're on the same page. The only thing needed to reproduce this issue is importing vue-router & setting the correct path, which I did here:
```
export default new Router({
  path: "/"
});
```
Adding anything else would be extraneous fluff.

The prominent issue here (and this is my fault for a poor title) isn't necessarily that it cannot detect the query parameters that precede the hash — the issue is that upon initially landing on an application that contains query parameters in the URL, the hash is being injected into the URL at the wrong place (after the query params instead of preceding them).

A side-effect of this is `this.$route.query` not being able to detect the parameters, but I'm not convinced it should. From a maintainers perspective, keep it as simple as possible, right?

I appreciate the workaround! As a matter of due diligence, I already researched this issue prior and was aware of the workaround, but going back to my original issue:
>Running across this in an instance where user authentication is done outside our app, but client/user identifier is passed to our app in query parameter upon authorization.

I probably didn't make this clear enough, but in my specific case, I'm not able to control the server or the URL directing to our app - so unfortunately I cannot apply the workaround.

[...]

PR #2253, opened 2018-06-11
Fix "#" placed incorrectly when query string is present on load (#2125)

devin-brenton:
When navigating to a URL with a query string like `myapp.com?foo=bar` in hash mode, the hash is currently placed after the query string `myapp.com?foo=bar#/`

This fix correctly parses URLs with query strings so that the above URL will be `myapp.com/#/?foo=bar` after loading.

### DIFF
--- a/src/history/hash.js
+++ b/src/history/hash.js
@@ -108,8 +108,15 @@ export function getHash (): string {
 function getUrl (path) {
   const href = window.location.href
   const i = href.indexOf('#')
-  const base = i >= 0 ? href.slice(0, i) : href
-  return `${base}#${path}`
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Hey, thanks for your contribution. Unfortunately, this will implicitly keep query parameters when navigating between routes

devin-brenton:
@posva thanks for the quick review. I noticed the behavior you're describing when navigating routes, but I wasn't sure what the desired behavior should be. Thanks for clearing that up.

emanuelmutschlechner:
I've submitted a PR #2262 which addresses this issue

PR #2262, opened 2018-06-19
Fix hash mode deep links (#2125)

### LINKED ISSUE #2125: Hash mode places # at incorrect location in URL if current query parameters exist on page load

emanuelmutschlechner:
Fixes #2125 

@posva can you confirm it doesn't introduce some weird side effects?

### DIFF
--- a/examples/hash-scroll-behavior/app.js
+++ b/examples/hash-scroll-behavior/app.js
@@ -52,6 +52,7 @@ const scrollBehavior = (to, from, savedPosition) => {
 
 const router = new VueRouter({
   mode: 'hash',
+  base: __dirname,
   scrollBehavior,
   routes: [
     { path: '/', component: Home, meta: { scrollToTop: true }},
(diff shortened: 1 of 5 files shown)


posva:
Thanks! This is not exactly it because we cannot just ignore the `fallback` option 😄 I'll have a look

emanuelmutschlechner:
The `fallback` option isn't ignored in the router itself. Deep link fixes were only applied when the router did a fallback from `history` to `hash` mode in IE9?

https://github.com/gonoware/vuejs-vue-router/blob/dev/src/index.js#L44-L48

appsecmobile:
@posva Hey, what I need to do for merging this PR. I need this )

persononomo:
@posva Hey, what I need to do for merging this PR. I need this )

kolaente:
Any updates here? I'm looking forward to this.

PR #2269, opened 2018-06-26
Fix an error on the document: 

YuJianrong:
param value should be string instead of number


posva:
Thanks for this. I'm not sure we want to force users pass strings. I do think params should be normalised though

YuJianrong:
@posva Em... I think maybe it's better to force the user to use string for the router params, consider:
1. When the param is part of the url, like `/abc/123` for router `/abc/:id`, the `id` will be a string instead of number, it may not be a good idea for the user to pass the param as a number pragmatically.
2. The param member is defined as `string` in the TypeScript description file so set the param as a number is conflict with the description `router.d.ts`:
```Typescript
export interface Route {
  path: string;
  name?: string;
  hash: string;
  query: Dictionary<string>;
  params: Dictionary<string>; // <- this one
  fullPath: string;
  matched: RouteRecord[];
  redirectedFrom?: string;
  meta?: any;
}
```

posva (approved):

PR #2277, opened 2018-07-05
Warn on duplicate path inside `routes` array (#668)

J13007847:
* warn on duplicate path inside `routes`

* fix trailing colon

* use jasmine `helper` config for `to-have-been-warned`


posva:
Not sure what you were trying to do here

PR #2279, opened 2018-07-05
refactor: do not modify array while loop

codetalks-new:
modify array while loop seems harder to understand.
we can use two filter and two push to implement this small part of function.


posva:
Thanks for your concern! However, this solution goes through the original array multiple times and create multiple arrays

codetalks-new:
well, that's fine. while, I think the length of pathList would not be two long. so I prefer the more readable version.

PR #2286, opened 2018-07-07
fix(normalizeLocation): create a copy with named locations

### LINKED ISSUE #2121: implicit params ({ name: 'sub-route' }) don't update when navigating between nested routes
### Version
3.0.1

### Reproduction link
[https://jsfiddle.net/67xnq5to/](https://jsfiddle.net/67xnq5to/)

### Steps to reproduce
1. click /foo/1/sub1
2. click /foo/2/sub1
3. click sub2

### What is expected?
id==2

id is 1

like #1255

posva:
Fix #2121

### DIFF
--- a/examples/nested-routes/app.js
+++ b/examples/nested-routes/app.js
@@ -36,15 +36,22 @@ const Quy = {
     </div>
   `
 }
-const Quux = { template: '<div>quux</div>' }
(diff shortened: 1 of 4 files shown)


posva (commented):
I need to add e2e test

PR #2292, opened 2018-07-10
feat(url): call afterEach hooks after url is ensured

### LINKED ISSUE #2079: afterResolve global navigation hook
As opposed to `beforeResolve`, `afterResolve` would wait until the navigation is done and trigger the registered callbacks.

The main reason for that is to be able to track events in the right moment, with the proper url in place, globally.

```js
routerInstance.afterResolve(cb);
function cb(from, to, next) {...}
```

I'd love to submit a PR if that's something you'd like to have.

posva:
Closes #2079

This helps integrating with external scripts and makes more sense as the
navigation is finished.  However, onComplete callback passed to
`router.push` is now called before `afterEach` hooks. This shouldn't be
a problem as there's no guarantee provided anywhere in the docs about
this


themeler:
Will this be pulled any time soon? It's quite relevant feature missing.

hectorguo:
Could I know when it will be released? Really need this fix for our project. Thanks!

PR #2295, opened 2018-07-12
fix: workaround replaceState bug in Safari

### LINKED ISSUE #2195: replaceState needs third argument
### Version
3.0.1

### Reproduction link
[https://github.com/raxxarr/router-demo.git](https://github.com/raxxarr/router-demo.git)

### Steps to reproduce
- in latest Safari(OS X or iOS)
- add base element: `<base href="/app/">`
- set router mode: history
1. open page 1 via url: localhost:8080/app/p1
2. navigate to /app/p2 by clicking button
3. click history back button on browser

### What is expected?
jump back to /app/p1

jump to /app/

---
This issue is caused by using replaceState without giving a third argument when trying to fix a Firefox scroll issue (https://github.com/vuejs/vue-router/blob/dev/src/util/scroll.js#L9)； This works fine on most browsers except for Safari.
According to whatwg spec，if the 3rd argument is not given，browser should set current url into history，but Safari set the base url.
Here is Webkit source code: https://trac.webkit.org/browser/webkit/branches/safari-605.2.8.1-branch/Source/WebCore/page/History.cpp#L157
Here is Blink source code: https://chromium.googlesource.com/chromium/blink/ /master/Source/core/frame/History.cpp#153
As is seen, what Safari do is not what we expect when we omit the 3rd argument.

Although this is not really a vue-router bug, we can add a third argument(current path) to avoid something weird happen.

ps: this bug is confirmed by Webkit https://bugs.webkit.org/show_bug.cgi?id=182678

posva:
Fix #2195

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -8,7 +8,8 @@ const positionStore = Object.create(null)
 
 export function setupScroll () {
   // Fix for #1585 for Firefox
-  window.history.replaceState({ key: getStateKey() }, '')
+  // Fix for #2195 Add optional third attribute to workaround a bug in safari https://bugs.webkit.org/show_bug.cgi?id=182678
+  window.history.replaceState({ key: getStateKey() }, '', window.location.href.replace(window.location.origin, ''))
   window.addEventListener('popstate', e => {
     saveScrollPosition()
     if (e.state && e.state.key) {


posva:
Still need to check in IE 9-11

PR #2298, opened 2018-07-13
Standardize kebab-case in route names by example

FossPrime:
I have been mistakenly using PascalCase in all my projects, when clearly, kebab-case is becoming the standard in route names. An example where a distinction between camelCase, snake_case and kebab-case should be included.

https://github.com/gothinkster/vue-realworld-example-app/blob/master/src/router/index.js

### DIFF
--- a/examples/named-routes/app.js
+++ b/examples/named-routes/app.js
@@ -12,7 +12,7 @@ const router = new VueRouter({
   base: __dirname,
   routes: [
     { path: '/', name: 'home', component: Home },
-    { path: '/foo', name: 'foo', component: Foo },
+    { path: '/foo-main', name: 'foo-main', component: Foo },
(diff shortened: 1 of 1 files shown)


posva:
Thanks for the PR, there's no such standard, actually... If you have uppercase routes, make sure to add the caseSensitive flag though (https://router.vuejs.org/api/#routes)

FossPrime:
I know there is none... but maybe there should be... otherwise Vue developers can and will go wild with different casing on their route names, we seek your guidance. The example already takes a side against PascalCase in route names ... note I'm not speaking about paths

posva:
oh, didn't get that. Then it's even freer as some people will prefer one over the other, but it doesn't matter at all. If, for some reason, my opinion matters on this, I use PascalCase names because they look different from paths. Cheers

PR #2350, opened 2018-08-13
Add note for props in named views

MoritzKn:
I searched for this is the docs for quite some time, had to have a coworker tell me and finally find the solution via [stack overflow](https://stackoverflow.com/a/42398362/5587737).


posva:
Sorry for the delay. This has been incorporated into the props section of the docs instead

PR #2357, opened 2018-08-21
Add clarification for child props

crswll:
Just wanted to clarify that when props are used on child routes that the parent props are also passed.


posva:
Thanks for the suggestion. I think this could be presented in another way. I will refactor this with something a bit different

PR #2365, opened 2018-08-27
Update type define of RouteConfig.props

xlfsummer:
In route config, we can pass an object to 'props' property, which key is the name of router view, and type of value is one of boolean, object or function (RoutePropsFunction).


posva (approved):
Sorry for the delay and thanks!

posva (changes requested):
Okay, I will need to check the typings tests. Feel free to do if you can!

posva (approved):
of course! I copy/pasted too fast
Thank you

posva:
oh btw, if you want to add some typing tests, you can otherwise I would add them later on

posva:
I this and it seems like the current type already allows passing an object. If someone finds a failing case, please open an issue or a pr with the failing test case (it can be added to types/test/index.ts

PR #2370, opened 2018-08-31
修复d.ts不能支持tsx的bug

yangdan8:


posva:
When vue-router is added and used, you can safely admit you have those :)

PR #2372, opened 2018-09-03
fix: go(-1) for AbstractHistory by initial stack

xuqingkuang:
In the original implementation, stack of AbstractHistory was initialized by empty array, it makes go(-1)[1] can't work correctly.

In most situation that can't go back, it doesn't matter, but in a special environment(such as running in JavascriptCore of Native App), go(-1) can't work will be a bug.

The problem is I haven't tested in SSR, the patch may break something, a better NodeJS determine method is required.

[1] https://github.com/vuejs/vue-router/blob/dev/src/history/abstract.js#L31

### DIFF
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -2,6 +2,7 @@
 
 import type Router from '../index'
 import { History } from './base'
+import { START } from '../util/route'
 
 export class AbstractHistory extends History {
   index: number;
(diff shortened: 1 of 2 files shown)


posva:
Hey, do you have a small repro showing what should work which isn't working right now?

xuqingkuang:
Sorry, I decided to fork it and make vue-router meets the requirement.

PR #2375, opened 2018-09-04
fix initial url path for non ascii urls

aldarund:
Here is demo of issue https://codesandbox.io/s/m2xnyj5zx

If click links withing app - active class applies fine for both links
If load page on /hello - active class applies to hello link fine
If load page on /тест - active class applies not applied to /тест link and the component not loaded

The problem is because on initial load vue-router use window.location.pathname for path, which is uri encoded, so it wont match routes because they are not uri encoded.

### DIFF
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -72,7 +72,7 @@ export class HTML5History extends History {
 }
 
 export function getLocation (base: string): string {
-  let path = window.location.pathname
+  let path = decodeURIComponent(window.location.pathname)
   if (base && path.indexOf(base) === 0) {
     path = path.slice(base.length)
   }
--- a/src/history/html5.js
+++ b/src/history/html5.js
@@ -72,7 +72,7 @@ export class HTML5History extends History {
 }
 
 export function getLocation (base: string): string {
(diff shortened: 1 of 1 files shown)


src/history/html5.js
```diff
@@ -72,7 +72,7 @@ export class HTML5History extends History {
 }
 
 export function getLocation (base: string): string {
-  let path = window.location.pathname
```
posva, line 75: I think we should use `decodeURI` to keep separators (such as `/` and `&`) encoded

aldarund: @posva changed

posva (approved):
Thanks, I will add a test locally and make some changes to hashmode as well to make it work

AbdallahSabri:
Does anyone solve this issue?

aldarund:
@AbdallahSabri what u mean?

AbdallahSabri:
@aldarund I mean does anyone fix initial url path for non ascii urls?

aldarund:
@AbdallahSabri as you can see this PR was merged. So its fixed. But it wasnt released to npm

PR #2391, opened 2018-09-16
add this <router-link :to="-1" go>

mlboy:
声明式 | 编程式
-- | --
\<router-link :to="..." go\>  | router.go(-1)

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -4,7 +4,7 @@ import { createRoute, isSameRoute, isIncludedRoute } from '../util/route'
 import { extend } from '../util/misc'
 
 // work around weird flow bug
-const toTypes: Array<Function> = [String, Object]
+const toTypes: Array<Function> = [String, Object,Number]
 const eventTypes: Array<Function> = [String, Array]
 
 export default {
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -18,6 +18,7 @@ export default {
       type: String,
       default: 'a'
     },
+    go: Boolean,
     exact: Boolean,
     append: Boolean,
     replace: Boolean,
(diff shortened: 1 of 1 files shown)


posva:
The point of the router-link being resolving links and displaying the actual url, this won't make sense, at least at the current state of the router. Instead, this should be use on some kind of button with `@click=$router.back()`

JounQin:
@posva `to` is a required prop for `router-link`, and I tried `<router-link to="" @click.prevent="$router.back()">`, it does not work.

---

My temporary workaround:

```vue
<router-link to="">
  <v-icon @click="$router.back()">{{ leftSvg }}</v-icon>
</router-link>
```

posva:
my point was: don't use a router-link if it's not going to render an `a` tag with the correct `href` attribute

JounQin:
@posva I added an `afterEach` hook:

```ts
import { Plugin } from '@nuxt/types'
import Vue from 'vue'
import Router from 'vue-router'

Object.defineProperty(Vue.prototype, '$prevRoutePath', {
  configurable: __DEV__,
  get(this: { $router: Router }) {
    const { previousRoute } = this.$router
    return previousRoute ? previousRoute.fullPath : '/'
  },
})

const routerPlugin: Plugin = ({ app: { router } }) => {
  router.afterEach((_to, from) => {
    router.previousRoute = from
  })
}

export default routerPlugin
```

But `@click.prevent` still does not work:

```vue
<router-link :to="$prevRoutePath" @click.prevent="$router.back()">
  <v-icon>{{ leftSvg }}</v-icon>
</router-link>
```

So I still have to do like following:

```vue
<router-link :to="$prevRoutePath">
  <v-icon @click="$router.back()">{{ leftSvg }}</v-icon>
</router-link>
```

Is that a bug that `router-link` could not be prevented?

Issue #2243, opened 2018-06-05
Allow a user to read and write state on history entries (history.state)


crswll:
### What problem does this feature solve?

Mobile apps that want to save scroll position for elements besides window.

This would also allow the user to save whatever they want to history state. A project I'm working on now needs to keep track of what's "focused" per page and this seems like the correct place for it. It could also be a nice place for temporary form data. I'm sure there's a lot of use cases outside of this.

### What does the proposed API look like?

I'm still thinking about it, though I think someone else thinking about it would be better. :)

I checked out `react-router` a bit and it looks like we can make state part of the Location object, so we can set it in `$router.push|replace({...})` or `next({...})` calls. They use an additional argument but we might not need to?

I took a quick look at react-router and it seems to allow for this: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/history.md

Also, this seems to talk directly to the situation we're facing with scrolling:
https://reacttraining.com/react-router/web/guides/scroll-restoration

Also some additional reading: https://github.com/ReactTraining/history#properties

Hopefully this is useful and doesn't seem like useless rambling.

jnields:
My app relies on being able to associate extra state in the history that isn't in the URL and this prevents me from being able to use `vue-router`

`react-router` uses [`history`](https://www.npmjs.com/package/history), which does something like:

```
history.replaceState(
  {  key: createKey(),  state: 'user-provided state' },
  null,
);
```

and provides access to location something like:

```
const location = {
  pathname: window.location.pathname,
  search: window.location.search,
  hash: window.location.hash,
  state: history.state.state,
}
```

users aren't expected to access `history.state` directly, though I suppose you could

719media:
This would be extremely useful as a way to pass data to the scrollBehavior handler to allow different router-links that share the same `to=` string/object so that they can differ in regards to scrollBehavior.

J-Rojas:
To workaround this, I currently have to track changes to the $route globally, inspect and copy values from `window.history.state`, then use `window.history.replaceState()` to push the new application state into the window history.  It would be helpful if this is done within the Router framework.

jimut:
There are some scenarios where this is impossible to workaround. Our app has some popups that need to close on the device back press. Right now we add a query in the URL to do that but as the number of popups increases, our URLs would become nasty.

[...]

PR #2397, opened 2018-09-18
Add access to history state

jnields:
This allows developers access to history.state, which you can use for associating extra data to the URL without updating the visible URL, in history mode.  It does not work for hash mode.

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -84,6 +84,7 @@ declare type Location = {
   path?: string;
   hash?: string;
   query?: Dictionary<string>;
+  state?: mixed;
   params?: Dictionary<string>;
   append?: boolean;
   replace?: boolean;
@@ -96,6 +97,7 @@ declare type Route = {
   name: ?string;
   hash: string;
   query: Dictionary<string>;
+  state: mixed;
   params: Dictionary<string>;
   fullPath: string;
   matched: Array<RouteRecord>;
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -1,3 +1,4 @@
+/* @flow */
 declare var document: Document;
 
 declare class RouteRegExp extends RegExp {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16847 @@
+{
+  "name": "vue-router",
+  "version": "3.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0-beta.44",
+      "resolved": "http://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz",
+      "integrity": "sha512-cuAuTTIQ9RqcFRJ/Y8PvTh+paepNcaGxwQwjIDRWPXmzzyAeCO4KqS9ikMvq0MCbRk6GlYKwfzStrcP3/jSL8g==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "7.0.0-beta.44"
+      }
+    },
+    "@babel/core": {
+      "version": "7.0.0-beta.47",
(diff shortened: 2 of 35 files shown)


posva:
Why are there 31 files modified? 😨
I appreciate your efforts on this but we don't want to add invisible state like this to html history because we have seen it becomes a pattern difficult to maintain in apps. We may add some way yo add data to the object passed to popstate but directly having a state in the route is exactly the pattern that has been used in other routers and led to any maintenance problems

jnields:
@posva - eslint wasn't running on test files so I just added it for that and did a `--fix`. It's a separate commit, so feel free to drop it if you like!

Is this really such a maintainability concern for _you_? Whether or not using `history.state` makes an app maintainable seems like the user's concern!

And as it stands right now I am using a home-brewed router for access to `history.state` to keep track of analytics data.

It's an appropriate place for this, and there are several other completely valid use cases for it as well (which is why it was added to the native API!)

I don't think it makes maintainability of this project much more difficult - so why prevent users from accessing it? It's preventing adoption in at least one case ...

jnields:
There is some work to finalize this - it could use some more tests and thinking through edge cases - but if you give me the green light I'll go ahead and add them and get this ready. The one commit I added is the bulk of it, and as you can see it wouldn't be that much

jakob-e:
To me state is for keeping track of navigation on a 'per browser tab' level.
At the moment Vue provides a state key but in order to detect back and forward navigation a key list 
and a previous key would be handy–e.g. to handle slide left/right transitions. 
````js
{
  keys:["1", "2", "3", "4", "5"],  // visited keys (matches browser history)
  oldKey: "5"                      // last route key  
  key: "2"                         // current route key
}
````
The above would indicate a go(-3) navigation–that could be used to set a slide-right transition 
in the `beforeEnter`guard.

I don't see get/set state as being part of the routes but as a method on the router

I'm sorry if I've missed an existing feature to handle this :-)

**Comment by emanuelmutschlechner** [comment]
@jnields I think the router shouldn't be responsible for managing state which is not required for routing.

It's possible to add extra state without changes in the router:
```javascript
const Foo = { template: '<div>foo</div>' }
const Bar = { template: '<div>bar</div>' }

const routes = [
    { 
        path: '/foo',
        component: Foo,
        state: { a: 1 }
    },
    {
        path: '/bar',
        component: Bar,
        state: { b: 2 }
    },
]

    routes
})

function urlToState(url, originalState) {
    const route = undefined; // TODO match url with routes.path, 
                                            // e.g. using regex or router.resolve
    const somethingElse = false;
    if (route) {
        return route.state || {}
    } else if (somethingElse) {
        return { random: 123 }
    }
    return { default: 999 }
}

function modifyState(data, url) {
    const state = urlToState(url, data)
    return {
        ...data,
        ...state,
    };
}

const { pushState, replaceState } = window.history;
window.history.pushState = (data, title, url) => pushState.call(window.history, modifyState(data, url), url);
window.history.replaceState = (data, title, url) => replaceState.call(window.history, modifyState(data, url), url);
```

rzane:
Is there any possibility that Vue Router will support this behavior? I assumed this was provided by Vue Router (because it's provided by every other client-side routing package that I've used), and I was really disappointed to find that it's not supported.

Because Vue Router controls all interaction with the history API, I'm concerned that if I use `history.pushState` directly, I'd break compatibility Vue Router. So now, it feels like Vue Router has actually taken away my access to stateful routing.

I also don't think that monkey-patching `history.pushState` is a viable solution.

posva:
Things have got a little bit outdated but I want to get a second look at this and want to know about the different use cases that require access to the state like applying transition on using the data on scrollBehavior.  The discussion can be found at https://github.com/vuejs/vue-router/issues/2243

I need to make sure people don't use this to put request data that should go somewhere else and maybe even come up with something new for that

PR #2405, opened 2018-09-23
Add Node.js(Express) Usage Example to Docs

HikaruSato:
I thought that I needed documentation on usage examples in Node.js (Express).
So I will suggest adding a usage example in Node.js (Express).

### DIFF
--- a/docs/guide/essentials/history-mode.md
+++ b/docs/guide/essentials/history-mode.md
@@ -70,6 +70,29 @@ http.createServer((req, res) => {
 
 For Node.js/Express, consider using [connect-history-api-fallback middleware](https://github.com/bripkens/connect-history-api-fallback).
 
+##### Example
+```bash
+mkdir express_project
+cd express_project
+npm init
+npm i express connect-history-api-fallback -s
+touch app.js
+```
+
+```js
+// express_project/app.js
+
+const express = require('express')
+const history = require('connect-history-api-fallback')
+const app = express()
+const server = app
+    .use(history())
+    .use(express.static('dist'))
+    .listen(80, () => {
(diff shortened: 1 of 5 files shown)


kazupon (commented):
Thank you for your PR!

If it's possible, we would be grateful if you could update the `zh`. 🙏

docs/guide/essentials/history-mode.md
```diff
@@ -70,6 +70,29 @@ http.createServer((req, res) => {
 
 For Node.js/Express, consider using [connect-history-api-fallback middleware](https://github.com/bripkens/connect-history-api-fallback).
 
+##### Example
+```bash
+mkdir express_project
```
posva: could you use `express-project` instead?

docs/guide/essentials/history-mode.md
```diff
@@ -70,6 +70,29 @@ http.createServer((req, res) => {
...
+cd express_project
+npm init
+npm i express connect-history-api-fallback -s
```
posva: can you remove the `-s` and use `install` instead, please

posva (changes requested):

docs/guide/essentials/history-mode.md
HikaruSato: yes, I fixed.

docs/guide/essentials/history-mode.md
HikaruSato: Yes, I fixed.
```bash
npm install express connect-history-api-fallback
```

HikaruSato:
@kazupon I amended `Example` of chinese translation.
Do you recognize it? Please confirm.🙇‍♂️

PR #2407, opened 2018-09-25
Forward args as props to custom child

### LINKED ISSUE #1678: Support for inheriting attrs in router-link tag
router-link can render a specified tag which could happen to be a component. This component could have its own props but currently there is no way to pass them down. Eg.

```
// the primary prop will not be passed down to my-button
<router-link tag="my-button" primary></router-link>
```

Unless I am missing something I would like to request a way to solve this situation.

wregis:
When rendering a custom child ("tag" prop) forward unhandled args to the component, like a button or something complex.

I believe this will not break any existing functionality while giving some new capabilities.

Guess this will fix #1678 and can be related to #2029.

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -9,6 +9,7 @@ const eventTypes: Array<Function> = [String, Array]
 
 export default {
   name: 'RouterLink',
+  inheritAttrs: false,
   props: {
     to: {
       type: toTypes,
@@ -95,6 +96,7 @@ export default {
       } else {
         // doesn't have <a> child, apply listener to self
         data.on = on
+        data.props = this.$attrs
       }
     }
 


sqal:
This is more complicated than this simple change you made. Firstly, it will break existing functionality, because you assign `$attrs` to `data.props` instead of `data.attrs`, so any custom attribute added to the link will be removed, e.g:

 `<router-link to="/" data-custom-attr="somevalue">/homepage</rotuer-link>`

data-custom-attrs will no longer be present on the link. 

As as said, `$attrs` needs to be assigned to `data.attrs`. This way, we solve also another potential problem. Imagine we have Button component that has `size` prop:

```
const Button = {
  props: {
    // button's size
    size: String,
  },
}
```

and we're using router-link like this:

`<router-link to="/" :tag="Button" size="lg" type="button">/homepage</router-link>`

With your change, we would correctly pass `size` prop with the value `lg` to the Button component. but not the `type="button"` attribute (`this.$attrs` in the component would be an empty object). Why not? Because `$attrs` are passed explicitly as props (`data.props = $attrs`) and Button does not have specified `type` prop. By assigning $attrs to data.attrs we resolve this issue because Vue always normalize those attributes. By normalize I mean, attributes that are defined as props will become props, others will go to `this.$attrs`.  If my explanation is not clear enough for you, please take a look at a simple demo I prepared https://jsfiddle.net/sfonLhva/. Open console and see what's printed out before and after fixing my comments. Also you forgot to change tag's prop type. You should add Object type  if we want this prop to accept component's object.

wregis:
Right, maybe it was a little naive of me to do this with no further investigation once it worked for my simple test case. I agree with your arguments and I'll be more careful next time.

In my case, I wanted to receive props on my custom componente, but as a side effect I see the attributes will be lost.

PR #2417, opened 2018-10-08
[vue-router] must call Vue.use(VueRouter) before creating a router

chiaweilee:
Assert warning if forget to call Vue.use(VueRouter) before creating.


posva:
Thanks but it can actually be called after too :)

Issue #2418, opened 2018-10-09
[Docs] HTML5 History Mode Caveat Clarification


ataylor32:
### Version
3.0.1

### Reproduction link
[https://router.vuejs.org/guide/essentials/history-mode.html#caveat](https://router.vuejs.org/guide/essentials/history-mode.html#caveat)

### Steps to reproduce
The documentation correctly states that using history mode and configuring the server with a catch-all fallback route has the caveat that the server will no longer report 404 errors. It says "To get around the issue, you should implement a catch-all route within your Vue app to show a 404 page". In my opinion, that wording implies that implementing a catch-all route within Vue completely fixes the problem. However, with that approach, if a user goes to an invalid URL, the server will respond with a status code of 200 even though the URL is invalid. On some sites (especially dynamic ones), there is a good chance that a URL that is valid today could be invalid tomorrow. Maybe some search engines are smart enough to recognize that it's a "Page Not Found" page and treat it as such rather than a valid URL, but I'm guessing that not all search engines are that smart. I think it's important for developers to be aware of this, so I think it would be a good addition to the "Essentials > HTML5 History Mode >  Caveat" documentation.

### What is expected?
The documentation should mention that the server will respond with a status code of 200 for invalid URLs.

### What is actually happening?
The documentation does not mention this.

posva:
You are more than welcome to submit a proposal for this :)

PR #2419, opened 2018-10-10
docs: improve "Essentials > HTML5 History Mode > Caveat" (#2418)

ataylor32:


posva:
Thanks for the proposal but you are pretty much repeating what is already said. Keep in mind, you can customise your server to still answer with a 404 status code and the index.html file

ataylor32:
@posva I'm not repeating what is already said. Consider the scenario where the server is not Node.js. When the server receives a request that's not for a static asset, it will blindly serve index.html regardless of if the URL is valid or not. Ideally, it should be like this:

1. For valid URLs, serve index.html with a status code of 200

1. For invalid URLs, serve index.html with a status code of 404

But in order to do that, you would have to have the routes shared between Vue and the server. Even then, it's still not perfect because with a route like `/user/:id`, the URL `/user/1` might be valid and the URL `/user/2` might be invalid, so the server would have to validate the ID before serving index.html in order to know which status code to return.

The current wording in the documentation is telling developers that the problem is solved by configuring the server to serve index.html with a status code of 200 for any URL, valid or not. That might be OK for some projects, but not all.

PR #2423, opened 2018-10-11
 Fix dates inside query objects being converted to empty objects while cloning

dword-design:
It happens that you want to add a Date as a query parameter. However, when creating the route, the query object is cloned. The clone function does not support dates since it converts dates to objects. This fix clones the Date in a separate conditional branch.


posva:
Thanks, but you are supposed to cast the Date to a String if you want to put it in the URL 😄

PR #2459, opened 2018-10-31
docs(ja): Fix incorrect bold style

oohira:
I'm not sure I also have to fix `docs-gitbook/ja/essentials/passing-props.md`. Would you give me an advice?

* Before
    image
* After
    image


posva:
Thanks!

Issue #2461, opened 2018-11-01
The path of the RouteConfig is the same as the alias content, which will cause an infinite loop.


chenfengjw163:
### Version
3.0.1

### Reproduction link
[https://codepen.io/chenfeng163/pen/qJzxGW](https://codepen.io/chenfeng163/pen/qJzxGW)

### Steps to reproduce
The path of the RouteConfig is the same as the alias content

### What is expected?
no error

### What is actually happening?
infinite loop

---
The route of alias overrides the route of the path, causing the path and matchAs to be the same. When the url is parsed, an infinite loop occurs.

chenfengjw163:
PR #2462 
before:

``` javascript

if (route.alias !== undefined) {
  const aliases = Array.isArray(route.alias)
    ? route.alias
    : [route.alias]

  aliases.forEach(alias => {
    const aliasRoute = {
      path: alias,
      children: route.children
    }
    addRouteRecord(
      pathList,
      pathMap,
      nameMap,
      aliasRoute,
      parent,
      record.path || '/' // matchAs
    )
  })
}

if (!pathMap[record.path]) {
  pathList.push(record.path)
  pathMap[record.path] = record
}

```
after:

``` javascript
if (!pathMap[record.path]) {
  pathList.push(record.path)
  pathMap[record.path] = record
}

if (route.alias !== undefined) {
  const aliases = Array.isArray(route.alias)
    ? route.alias
    : [route.alias]

  aliases.forEach(alias => {
    const aliasRoute = {
      path: alias,
      children: route.children
    }
    addRouteRecord(
      pathList,
      pathMap,
      nameMap,
      aliasRoute,
      parent,
      record.path || '/' // matchAs
    )
  })
}
```

posva:
I think a warning makes more sense here as I don't see a reason to have a path with the same value as an alias

chenfengjw163:
Yes, but the problem now is that if path and alias are the same, an error will be reported. In #2462 , I have added warning.
``` javascript
if (route.alias === path) {
  warn(
    false,
    `Path with the same value as an alias: ` +
    `{ path: "${path}" }`
  )
}
```

PR #2462, opened 2018-11-01
Fix the error caused by the same path and alias of RouteConfig

chenfengjw163:

### DIFF
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -118,24 +118,32 @@ function addRouteRecord (
   }
 
   if (route.alias !== undefined) {
-    const aliases = Array.isArray(route.alias)
-      ? route.alias
-      : [route.alias]
-
-    aliases.forEach(alias => {
-      const aliasRoute = {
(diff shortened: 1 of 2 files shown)


chenfengjw163:
#2461

src/create-route-map.js
```diff
@@ -112,32 +112,40 @@ function addRouteRecord (
...
-    })
-  }
-
   if (!pathMap[record.path]) {
     pathList.push(record.path)
     pathMap[record.path] = record
   }
 
+  if (route.alias !== undefined) {
+    if (route.alias === path) {
```
posva: The warn should only appear on dev mode. Also could you add a test for it?

posva (changes requested):

src/create-route-map.js
chenfengjw163: OK, I have already added

posva:
Hey, I added a check for arrays too and directly merged to `dev`. Thank you!

PR #2473, opened 2018-11-12
Warn only when there is no matching path

minaseem:
Otherwise it pollutes console with warnings


posva:
Thanks but `warn` already does that check against `record`

minaseem:
yeah!
Apologies for bothering

posva:
no worries 🙂 !

PR #2497, opened 2018-11-25
feat(types): NavigationGuardNext

beary:
Expose a `Next` type for this usage:
```typescript
// ...
import { Route, Next } from 'vue-router'

@Component
export default class Verify extends Vue {
  // ...
  private beforeRouteEnter(to: Route, from: Route, next: Next) {
    next(vm => {
      vm.$data.username = decodeURIComponent(to.query.username || '')
    })
  }
}
```


kevlarr:
Exactly the PR I was about to submit.. @posva How do PRs get reviewed/merged?

posva:
I will come at it when I have time, after other Bugs and features

voool:
As a workaround, to grab the `next` interface with Typescript you can do
```
import { NavigationGuard } from 'vue-router';
type Next = Parameters<NavigationGuard>[2];
```

kevlarr:
Nice @voool, that's really clean! I didn't know that was possible - I've been copying the type def into my own code, which is more verbose and also more brittle (in case the def changes at all).

andredewaard:
Any progress on when this will be merged?

faustbrian:
@posva any progress on this?

d2phap:
Any update?

types/router.d.ts
```diff
@@ -7,10 +7,11 @@ type ErrorHandler = (err: Error) => void;
 export type RouterMode = "hash" | "history" | "abstract";
 export type RawLocation = string | Location;
 export type RedirectOption = RawLocation | ((to: Route) => RawLocation);
+export type Next<V extends Vue = Vue> = (to?: RawLocation | false | ((vm: V) => any) | void) => void;
```
```suggestion
export type NavigationGuardNext<V extends Vue = Vue> = (to?: RawLocation | false | ((vm: V) => any) | void) => void;
```

types/index.d.ts
```diff
@@ -12,5 +12,6 @@ export {
   RouteRecord,
   Location,
   Route,
-  NavigationGuard
+  NavigationGuard,
+  Next
```
posva:
```suggestion
  NavigationGuardNext
```

types/router.d.ts
```diff
@@ -7,10 +7,11 @@ type ErrorHandler = (err: Error) => void;
...
 export type NavigationGuard<V extends Vue = Vue> = (
   to: Route,
   from: Route,
-  next: (to?: RawLocation | false | ((vm: V) => any) | void) => void
+  next: Next<V>
```
```suggestion
  next: NavigationGuardNext<V>
```

posva (changes requested):
I'm renaming the type to match vue-router-next
One thing that is problematic is that only `beforeRouteEnter`'s `next` allows a callback. So I think we should export a type without the callback and make a specific type for `beforeRouteEnter` where we change `next` to include `(vm: V) => any`

beary (commented):
Looks good :smiley:

beary (commented):
:+1:

beary:
I am not familiar with the operation of pr, if it interferes with people, I am sorry.
Or if the commit record is too confusing, you can close this pull request and create a new one.

posva:
No worries for the commit history @beary, we squash commits at the end! Let me know if this is things are not clear and I will finish this up

beary:
@posva 
I read your changes, I think it is very good, And there is nothing I need to add!

posva:
@beary There is a change left: https://github.com/vuejs/vue-router/pull/2497#pullrequestreview-408928568

beary (commented):
Reviewed

beary (commented):
I submitted it for review, but it is still in a reviewable state

posva (changes requested):
One thing that is problematic is that only `beforeRouteEnter`'s `next` allows a callback. So I think we should export a type without the callback and make a specific type for `beforeRouteEnter` where we change `next` to include `(vm: V) => any`

posva:
@beary because the review is the same

beary (commented):
Reviewed

posva:
what? see https://github.com/vuejs/vue-router/pull/2497#pullrequestreview-408976579
if you don't see where to add, that's fine, just let me know and I will pick it but let's not go in circles and waste each other time

beary:
@posva 
Sorry I couldn't add it recently

PR #2498, opened 2018-11-25
[Docs] Use ternary operator in example

parkourben99:


kazupon (commented):
Thanks!

If it's possible, we would be grateful if you could update the other langs (`zh`, `ja`, and more). 🙏

posva:
I'm actually not that convinced for the example to use a ternary here. Even though  I would use it in my projects, I think it's easier to learn using a plain if/else @chrisvfritz WDYT?

parkourben99:
Shouldn't we be showing best practice examples?

chrisvfritz:
I agree with @posva that this just makes the examples more difficult to read for many people. For those who are very familiar and comfortable with ternaries, they'll already know how to convert this into a ternary. For people who aren't, this more verbose version works just as well and still might be preferred to the ternary for readability.

Issue #2493, opened 2018-11-22
[DOCS] Clearer usage of prop functions with named views


tesla3327:
### What problem does this feature solve?
I spent a few hours this morning trying to get prop functions with named views to work.

At first I thought that `props` itself was supposed to be a function that returned an object for all named views:
```js
const routes = [
  {
    path: '/organizations/:id/insights/:widget',
    components: {
      default: SeeMore,
      filter: Filter,
      actionBar: ActionBarBack,
    },
    props: route => ({
      actionBar: {  
        name: route.params.widget,
      },
      filter: {
        hideAutocomplete: true,
      }
    }),
  },
]
```

After reading and stepping through the `vue-router` code, I eventually figured out the correct implementation is:
```js
const routes = [
  {
    path: '/organizations/:id/insights/:widget',
    components: {
      default: SeeMore,
      filter: Filter,
      actionBar: ActionBarBack,
    },
    props: {
      actionBar: route => ({  
        name: route.params.widget,
      }),
      filter: {
        hideAutocomplete: true,
      }
    },
  },
]
```

The docs are not very clear (at least to me) that this is the correct implementation. Not sure if others have had this problem, or if it's just me 🤷‍♂️

### What does the proposed API look like?
My suggestion is to add a section to `Passing Props to Route Components` that explains how to use this with `Named Views`, along with an example.

Could also add a link to this section in the `Named Views` section of the docs.

posva:
This is indeed missing in docs. It should go into https://router.vuejs.org/guide/essentials/passing-props.html if anyone wants to make a contribution
And, as you said, a little note to it in the Named Views section would also be nice.

karanisverma:
@posva I would like to contribute here!

suprabhasupi:
@posva I would like to pick this up

posva:
go ahead

suprabhasupi:
> go ahead

Thanks @posva! 
I have created a PR for the this issue. 
I have tried to be as simple and direct as possible(As this is my first contribution I might be missing some information to add), Please have a look and let me know if there is any changes required. 🙂

y-nk:
I wished that this PR was merged sooner. Hopefully this ticket was easy to find.

mikeher:
I also just spent an extraordinary amount of time trying to find a solution until I stumbled upon this ancient issue. Thank you for your efforts, this solved my problem.

douglasg14b:
[...]

PR #2499, opened 2018-11-26
Adding name mode section in passing props to route component

suprabhasupi:
Fixes (#2493)

### DIFF
--- a/docs/guide/essentials/passing-props.md
+++ b/docs/guide/essentials/passing-props.md
@@ -72,4 +72,22 @@ The URL `/search?q=vue` would pass `{query: 'vue'}` as props to the `SearchUser`
 
 Try to keep the `props` function stateless, as it's only evaluated on route changes. Use a wrapper component if you need state to define the props, that way vue can react to state changes.
 
+## Name mode
+
+You can pass props to component using their component name.
+
+``` js
+const router = new VueRouter({
+  routes: [
+    {
+      path: '/user-profile',
+      component: {location: Location, userInfo: UserInfo},
+      props: { userInfo: route => ({ isNewUser: false }) }
+    }
+  ]
+})
+```
+
+The URL `/user-profile` would pass `{ isNewUser: false }` to `userInfo` Component.
+
 For advanced usage, check out the [example](https://github.com/vuejs/vue-router/blob/dev/examples/route-props/app.js).


suprabhasupi:
@posva Is there any issue with this PR. Why its is not getting merged? Please let me if there are changes to be done.

posva:
No, no problems. It's not getting merged yet because I haven't checked :)

autumnwoodberry (commented):
This behavior is not really one of the "modes" and the example in the documentation has a comment/example that explains it already:

```
// for routes with named views, you have to define the `props` option for each named view:
{
  path: '/user/:id',
  components: { default: User, sidebar: Sidebar },
  props: { default: true, sidebar: false }
}
```

Perhaps we could remove this comment/example from the code at the top of the page and move it to a new section at the bottom titled "Usage with named views". This way it would be explained as an explicit feature which would be harder to miss. You could even mention that the rules still apply for using the different modes. Maybe something like this:

    ## Usage with named views

    A route with [named views](named-views.html) must have an explicitly defined `props` option for each named view.

    In the following example, the `header` and `default` views will recieve the `id` prop, but the `footer` view will not.

    ```js
      routes: [
        {
          path: '/user/:id',
          components: {
            header: Header,
            default: User,
            footer: Footer
          },
          props: {
            header: true,
            default: true,
          }
        }
      ]
    });
    ```

    The `props` option for each named view can still be defined using any of the available modes:

    ```js
      routes: [
        {
          path: '/user/:id',
          components: {
            header: Header,
            default: User,
            footer: Footer
          },
          props: {
            header: { sticky: true },
            default: route => route.query,
            footer: true,
          }
        }
      ]
    })
    ```

docs/guide/essentials/passing-props.md
```diff
@@ -72,4 +72,22 @@ The URL `/search?q=vue` would pass `{query: 'vue'}` as props to the `SearchUser`
...
+## Name mode
+
+You can pass props to component using their component name.
+
+``` js
+const router = new VueRouter({
+  routes: [
+    {
+      path: '/user-profile',
+      component: {location: Location, userInfo: UserInfo},
```
tmorehouse, line 84: Should this be `components`  instead of `component`?

posva:
I think having this documented at https://router.vuejs.org/guide/essentials/passing-props.html as it currently is makes more sense

PR #2506, opened 2018-11-27
fix: expect 0 to be defined

### LINKED ISSUE #2505: Catch all route with prefix throw `Expected "0" to be defined`
### Version
3.0.2

### Reproduction link
[https://codesandbox.io/s/lykxn57kr7](https://codesandbox.io/s/lykxn57kr7)

Prefix routes are fine when they don't have a prefix but with a `/sth/` it becomes problematic.

This causes problems especially for `nuxt-i18n` together with a catch-all-route (as it uses lang prefixes)

### Steps to reproduce

1. Visit the CSB/Website
2. See warning in the console.

Info: This did not happen in 3.0.1
Related: #2496, #2503 

### What is expected?
No warning

Warning

clarkdo:
Try to fix #2505, delegate `params['pathMatch']` to `params['0']` for `pathToRegexp` package matching.

### DIFF
--- a/src/util/params.js
+++ b/src/util/params.js
@@ -17,11 +17,18 @@ export function fillParams (
     const filler =
       regexpCompileCache[path] ||
       (regexpCompileCache[path] = Regexp.compile(path))
+    if (params && params.pathMatch) {
+      params[0] = params.pathMatch
+    }
     return filler(params || {}, { pretty: true })
   } catch (e) {
     if (process.env.NODE_ENV !== 'production') {
       warn(false, `missing param for ${routeMsg}: ${e.message}`)
     }
     return ''
+  } finally {
+    if (params && params[0]) {
+      delete params[0]
+    }
   }
 }


rchl:
This fix feels hacky but I think for all practical purposes, it should be fine.

Maybe I would create a copy of params instead of mutating it to eliminate need to remove property and make it a bit cleaner.

Also, I would add tests to test the bug and the fix - here is my modified create-matcher.spec.js with relevant tests added - https://gist.github.com/rchl/2418f5c40e66f9f27a34f89f83960b9a

Please prioritize this @posva  as it's a breaking bug in specific setup (with nuxt-i18n) and in development ENV, it breaks nuxt-link/router-link (in production just cases warning in the client).

posva:
I will come to it when I can :/
But it's supposed to just be a warning that shouldn't be displayed. Why would it break. And yes, a tes is required

rchl:
Sorry, this is just a warning. I've confused this bug with https://github.com/nuxt-community/nuxt-i18n/issues/73

posva:
I'll merge this and add the test and comments locally, thanks!

posva:
hmm, doesn't seem to be fixed yet. I need to debug further. It's now on a branch `fix/2505` in case you want to take a look. I won't be able to until tomorrow

PR #2512, opened 2018-12-02
no need set init data;just use async data

miewx:
use like this
```
export default {
    beforeRouteEnter:(from, to, next)->
        li = (await $.get '/api/li.txt').split('\n').filter(Boolean)
        @data = ->
            {
                n:0
                now:1
                li
                src : [
                    li[0]
                    li[1]
                ]
            }
        next()

```


posva:
Hello, do you have a reproduction of the problem you are fixing?

posva:

PR #2518, opened 2018-12-04
style(test): add eslint configuration for unit tests

rchl:
While trying to familiarize myself with code base and tests, got annoyed by my editor highlighting "errors" in test files so decided to add config for unit tests.


.eslintrc
```diff
@@ -6,5 +6,18 @@
...
+  ],
+  "overrides": [
+    {
+      "files": [
+        "**/*.spec.js"
+      ],
+      "plugins": [
+        "jasmine"
+      ],
+      "env": {
```
posva: the part specific to tests can actually go into an `.eslintrc` file inside the test folder

posva (changes requested):
Thanks!

PR #2531, opened 2018-12-13
Prevent exceptions when a consumer calls `isSameRoute` with falsey `a`

zspencer:
In certain cases, I've encountered an exception where `isSameRoute` is
called by an upstream consumer where a is undefined. This is despite the
fact that the type declaration explicitely prohibits such a thing.

Unfortunately, due to my own incompetence at reading the way the vue
error handling stack works; I've been unable to isolate the *caller*.

So, I've added a test case where `a` is a falsey value; in the hope that
this will prevent an exception from being raised, which was triggering
an infinite loop in my particular case; which, again unfortunately, I couldn't figure
out higher up the stack.

Since this is possibly a bug in the calling library, I'm not really
sure it's worth fixing here; but if it is fixed here it'd save me a few
more hours or days of debugging.

Further, I've been unable to create an isolated use case; as my app relies on a number of plugins and middleware with NuxtJS and reproducing it outside of the app I'm working with (Sadly) seems like another huge time sink.

If you'd like, I can try to convince my client to invest the time to resolving the error in a more appropriate manner, otherwise I'll likely wind up needing to rely on a forked version of vue-router :/.

### DIFF
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -73,7 +73,7 @@ function getFullPath (
 export function isSameRoute (a: Route, b: ?Route): boolean {
   if (b === START) {
     return a === b
-  } else if (!b) {
+  } else if (!a || !b) {
     return false
   } else if (a.path && b.path) {
     return (
--- a/test/unit/specs/route.spec.js
+++ b/test/unit/specs/route.spec.js
@@ -66,6 +66,16 @@ describe('Route utils', () => {
       expect(isSameRoute(a, b)).toBe(true)
       expect(isSameRoute(a, c)).toBe(false)
     })
+
+    it('handles cases where the a route is falsey', () => {
+      const a = undefined
+      const b = {
+        name: 'a',
+        hash: '#hi',
+        query: { arr: ['1', '2'], foo: 'bar' }
+      }
+      expect(isSameRoute(a, b)).toBe(false)
+    })
   })
 
   describe('isIncludedRoute', () => {


zspencer:
I am not sure why the e2e test is failing on CI, as it is working on my local machine. Not that that is useful information; but is this test perhaps flakey?

Also, I'm happy to write a real issue for this so I can structure my commit messages appropriately, but will be unable to provide an end-to-end reproduction.

posva:
Okay, I will need an actual (boiled down) repro of the problem as this fix may not be enough to fix the actual problem. I know it's hard but if I cannot do much with just this 😓

zspencer:
I mean, it fixes the actual problem in my clients code-base. The question is can I convince my client to invest the $X that it will take to create the reproduction step?

And right now, I can't. That's my own failing, for sure. It's for a prototype that will likely be replaced in the next year; so keeping my fork up to-date will be significantly cheaper.

zspencer:
Would video of the exception be helpful?

posva:
If you can't reproduce it now, it's fine. Come back once you have a boiled down repro then 🙂 
No, a video won't help 😛

zspencer:
Totally understandable, if anyone reading this thread is willing to pair with me for 1~2 hours over video chat to create an isolated reproduction of this issue; I'll happily pay $400 to them and/or put that into the Vue open-collective.

The particular parts I am struggling with are the way that the call in the trace is happening within nextTick; which appears to be obliterating the stacktrace into whatever plugin or part of the NUXT stack is causing this issue.

If that's of interest to anyone reading this, [DM me on twitter](http:///www.twitter.com/zspencer) or [dev.to](https://dev.to/zspencer) with a brief description of why you're qualified to help and I'll set up a screenshare.

Issue #974, opened 2016-12-02
Force navigation when clicking on a router-link even if the route doesn't change


DevanB:
Hey there. This is more of a question than a bug, but could turn into a feature request, if there is no way to currently handle this.

tl;dr; - Is there some way for me to use a global hook and run some code regardless if the current page is the page that will be navigated too?

I'm trying to call some methods (Vuex commits actually) in the vue-router `beforeEach` or `afterEach` hooks. These methods do things like closing the navigation, which need to be ran when clicking every link in the application. Everything works like a charm, until the user clicks a link to the page they are currently on. It seems (as it should) that the page doesn't actually change, but also that none of the hooks are called.

Thanks!

posva:
It makes sense not to call the hooks because there's no navigation since it's the same route.
@DevanB What code do you need to run on that case?

Edit: The easiest workaround to trigger a new navigation is to include a fake query or hash so the route is indeed different and the navigation isn't aborted, e.g. `$router.push({ query: { ...$route.query, t: Date.now() }})`

DevanB:
I'm trying to run a Vuex commit to close the navigation either before or after the route transition. 

It does make sense for the hooks not to run because no navigation is happening, but I assume that is overridable somehow. If not, I guess I have to dump `<router-link>` tags and go to straight anchor tags to bind a method, call my Vuex commit and then manually push the navigation (if it's not already on the page). All of that seems lengthy just to run some arbitrary code if it wasn't possible to somehow write it in one place (a hook for instance) and have it run each time. Even if I had to do the check if it's the same route that's fine, because it's less code to declare it in one place than binding a method every link in my entire app and manually navigating.

posva:
You can also create a custom component that binds the right vuex action to the click event on top of using a router-link element

DevanB:
Didn't think about that! I will give it a shot. Thanks for the suggestion! Keep up the great work! Maybe one day I'll have the pleasure of being a collaborator on Vue; really enjoying working with it.

DevanB:
Just an update for any people that may want to solve this in a similar manner. I solved it by creating, importing and using this component, instead of `router-link`:

`RouterLink.vue`
```
<script>
  import { mapActions } from 'vuex'
  export default {
    methods: {
      ...mapActions([ 'closeIsiAndNavigation' ])
    },
    name: 'RouterLink',
    props: {
      to: {
        type: String,
        required: true
      },
      tag: {
        type: String,
        required: false,
        default: 'a'
[...]

PR #2532, opened 2018-12-13
public method "reload"

ibyteyou:
test page: /reload
my reason: i needs in method to trig router-events (eaches), because in my project, static information about page title (document.title and title-markup in header with i18n) gets of $route information (?id or /:id, (supply №${id})), but now i need extra values such as "company_name"...

1. this.$router.meta.company_name = 'Some company'
2. this.$router.reload()
3. ???
4. all titles in page updated

in test-page "examples/reload" there are other useful uses (transition update, re-confirm, counts)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -25,6 +25,7 @@
     <li><a href="auth-flow">Auth Flow</a></li>
     <li><a href="discrete-components">Discrete Components</a></li>
     <li><a href="nested-router">Nested Routers</a></li>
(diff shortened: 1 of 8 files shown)


posva:
Thanks for your contribution. However this is quite different from what we want to add 🙂 
There is https://github.com/vuejs/vue-router/issues/974 opened but I haven't even confirmed with others that we want to add that feature nor came up with the exact API yet although it would probably be another option passed alongside params and query

PR #2538, opened 2018-12-19
fix(match): Optional params kept when opening nested route (fix #2110)

### LINKED ISSUE #2110: Optional parameters are dropped in named nested routes
### Version
3.0.1

### Reproduction link
~~[https://codepen.io/wrrrn/pen/GxqmOb](https://codepen.io/wrrrn/pen/GxqmOb)~~
http://jsfiddle.net/7a31maev/1/

### Steps to reproduce
I've linked to what the current router code is. The issue is simple, so I hope I'm missing something obvious.

### What is expected?
When the :paramOptional is set, and child route is pushed to, the path will look as such - 
'/search/paramOne/paramOptional/paramTwo/show-modal'

When navigated to, the child removes the paramOptional. The resulting url is - 
'/search/paramOne/paramTwo/show-modal'

ElkeCodes:
When navigating to a nested route, we do not discard the optional params for the parents to ensure that params are not lost.

### DIFF
--- a/examples/global.css
+++ b/examples/global.css
@@ -7,7 +7,7 @@ html, body {
   padding: 0 20px;
 }
 
-ul {
+ul, ol {
   line-height: 1.5em;
   padding-left: 1.5em;
 }
--- a/examples/nested-routes/app.js
+++ b/examples/nested-routes/app.js
@@ -44,6 +44,17 @@ const Zap = {
   template: '<div><h3>zap</h3><pre>{{ $route.params.zapId }}</pre></div>'
 }
 
+const Fox = {
+  template: `
+    <div>
+      <h3>fox</h3>
+      <pre>{{ $route.params.paramOptional }}</pre>
+      <router-link class="optional-param-nested" :to="{ name: 'foxy' }">Child with optional param in parent</router-link>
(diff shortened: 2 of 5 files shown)


posva (approved):
Thanks for this and sorry for the delay!
I had to look back a few other commits that were conflicting with this change

posva:
I'm realizing after more changes that this isn't enough as relative locations (`{ params: { param1: '1' }}`) won't work because they still get all params merged from the current location...
Fixing this require a way bigger refactoring the I anticipated

macco3k:
Does this explain why in the following definition the `meterId` parameter is dropped when typing the route in the browser?

```typescript
{
  name: "overview/meters",
  path: "/overview/:id/meters/:type/:meterId?",
  component: ListsPage,
  meta: {
    requiresAuth: true,
    breadcrumbs: true,
    access: ["site", "group"],
  },
  props: (route) => ({
    viewSource: "meters",
    options: {
      energyType: route.params.type,
      expandedMeterKeys: route.params.meterId ? [route.params.meterId] : [],
    },
  }),
},
```

So what's the status on this one? It is quite annoying to have to define two (almost identical) routes just to be able to use optional parameters.

PR #2553, opened 2018-12-25
docs: add .net mvc example server config for history mode

mzhukovs:
This should be helpful for some folks from the .NET world to get routing properly setup in their [Core] MVC app. Currently, there is no sample under the server config section on [this page](https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations) in the docs.

### DIFF
--- a/docs/guide/essentials/history-mode.md
+++ b/docs/guide/essentials/history-mode.md
@@ -122,6 +122,31 @@ Add this to your `firebase.json`:
 }
 ```
 
+#### ASP.NET Core MVC
+
+Add this to your `Startup.cs`:
+
+```csharp
+public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+{
+    ...
+    app.Use(async (context, next) =>
+    {
+        await next();
+
+        var path = context.Request.Path.Value;
+
+        if (context.Response.StatusCode == (int)HttpStatusCode.NotFound && !Path.HasExtension(path) && !path.StartsWith("/api"))
+        {
+            context.Request.Path = "/Home";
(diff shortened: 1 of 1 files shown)


lloydjatkinson:
I think it would be important to differentiate in the docs between the configuration for a normal multi-page app and a SPA with REST API backend.

For example this is my configuration for an ASP.NET Core MVC REST API with a Vue SPA.

```cs
    public class Startup
    {
        public Startup(IConfiguration configuration)
        {
            Configuration = configuration;
        }

        public IConfiguration Configuration { get; }

        // This method gets called by the runtime. Use this method to add services to the container.
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_2);

            // In production, the Vue files will be served from this directory
            services.AddSpaStaticFiles(configuration =>
            {
                configuration.RootPath = "ClientApp/client-app/dist";
            });
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }
            else
            {
                // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
                app.UseHsts();
            }

            app.UseHttpsRedirection();
            app.UseStaticFiles();
            app.UseSpaStaticFiles();
            app.UseMvc();

            app.UseSpa(spa =>
            {
                spa.Options.SourcePath = "ClientApp/client-app";
            });
        }
    }
}
```

PR #2577, opened 2019-01-11
docs(zh): link to the correct website

tuxinghuan:


posva:
Thanks! I saw there were more remaining links that were wrong so I fixed them all

Issue #1381, opened 2017-04-27
Named route throws Cannot assign to read only property error


parallelo3301:
### Version
2.5.2

### Reproduction link
[https://github.com/fandaa/vue-router-bug/commits/master](https://github.com/fandaa/vue-router-bug/commits/master)

### Steps to reproduce
1. Clone repo
2. `npm i` and `npm run dev`
3. Navigate to `http://localhost:8080/x` and click on `go to y`

### What is expected?
Browser navigated to route `/y`.

### What is actually happening?
Getting `TypeError: Cannot assign to read only property 'path' of object '#<Object>'` as error.

---
You can try to start at  `http://localhost:8080/y` and try to navigate to `/x` by clicking on links with `go to x using` as a prefix.

vue-router seems to not cooperate with vue@2.3.0, it works just fine within my project using vue@2.2.6.

The problem is only with named routes, you can try `http://localhost:8080/a`, where you can navigate to `/b` and vice-versa without any difficulty.

parallelo3301:
By replacing this line [https://github.com/vuejs/vue-router/blob/dev/dist/vue-router.esm.js#L1302](https://github.com/vuejs/vue-router/blob/dev/dist/vue-router.esm.js#L1302) (source:
 [https://github.com/vuejs/vue-router/blob/dev/src/create-matcher.js#L56](https://github.com/vuejs/vue-router/blob/dev/src/create-matcher.js#L56)) to:
```
location = Object.assign({}, location, {
  path: fillParams(record.path, location.params, `named route "${name}"`),
})
``` 
it no longer throws an error. I don't have enough time to trace it more deeply now (why is .path not writeable), but I hope this is at least good starting point.

posva:
hey @fandaa can you simplify the example in a js fiddle, please: http://jsfiddle.net/posva/9r6xhqbp/

parallelo3301:
Hi @posva, I tried to simplify it at [WebpackBin](https://www.webpackbin.com/bins/-KiuknCJJWQsuv0ctGlA).

If you uncomment line 20 at `main.js` (`name: 'a'`), it will start throwing an error (if you are on /b and try to navigate to /a).
Also, what I found - if you at the same time comment line 47 (`const matched = router.getMatchedComponents(to)`) it will now work. So calling `router.getMatchedComponents(to)` causes the issue.

[...]

PR #2583, opened 2019-01-17
fix(create-matcher): "cannot assign to read only property path"

SetSplin:
In create-matcher.js in match function location was marked as constant. That course as error described here (https://github.com/vuejs/vue-router/issues/1381)

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -54,8 +54,20 @@ export function createMatcher (
       }
 
       if (record) {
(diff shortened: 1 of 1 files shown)


posva:
Thanks but that was already fixed as you can see in the repro (http://jsfiddle.net/polasekf/sg5Lju5z/1/). If you found another bug please, open a new issue with a boiled down repro

PR #2588, opened 2019-01-21
docs: add descriptions for parseQuery / stringifyQuery

pfeiferbit:
This clears up what the implementor of these functions has to expect.

---

Stumbled across this while trying to use the [qs](https://www.npmjs.com/package/qs) package for `parseQuery` and `stringifyQuery` since I work with a PHP backend that relies on `foo[bar][baz]=qux` syntax with a strict RFC-3986 encoding.

Wording could probably be better since English ain't my primary language.

Greetings

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -251,6 +251,8 @@ Since it's just a component, it works with `<transition>` and `<keep-alive>`. Wh
 - type: `Function`
 
   Provide custom query string parse / stringify functions. Overrides the default.
+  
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Thanks for this! I think it would make more sense to add an example with qs (which is more likely to be what people want to use) and add your notes as comments in the code.

posva:
Closing due to inactivity. If anybody wants to improve the documentation in that aspect, please, open a new PR :)

PR #2591, opened 2019-01-22
Warn about root paths without a leading slash (fix #2550)

### LINKED ISSUE #2550: warn about root paths without a leading slash
### Version
2.5.3

### Reproduction link
[https://codesandbox.io/s/6zvmx5lorr](https://codesandbox.io/s/6zvmx5lorr)

### Example
```js
import Vue from "vue";
import Router from "vue-router";

import Home from "../src/views/Home.vue";
import Products from "../src/views/Products.vue";

Vue.use(Router);

const routes = [
  { name: "home", path: "/", component: Home },
  {
    name: "products",
    path: "products", // <-- the problem is here
    component: Products
  }
];

export default new Router({
  routes
});
```

### Steps to reproduce
1. Open the reproduction
1. Click on the products link

### What is expected?
I spent some time trying to understand what the problem was, so it would be nice to see a warning message telling that the route is invalid because doesn't have the initial slash.

The router goes to a route that doesn't exist.

Sayegh7:
close #2550

This PR creates the following warning message when a root route exists without a leading slash:
screen shot 2019-01-22 at 3 12 15 am

This facilitates the debugging process as the current behavior does not generate any warnings or errors.

### DIFF
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -37,10 +37,10 @@ export function createRouteMap (
   if (process.env.NODE_ENV === 'development') {
     // warn if routes do not include leading slashes
     const found = pathList
-      .find(path => path.charAt(0) !== '*' && path.charAt(0) !== '/')
+      .filter(path => path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
 
-    if (found) {
-      warn(false, `Non-nested routes must include a leading slash character. Replace "${found}" with "/${found}".`)
+    if (found.length > 0) {
+      warn(false, `Non-nested routes must include a leading slash character. Replace "${found[0]}" with "/${found[0]}".`)
     }
   }
 
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -37,10 +37,15 @@ export function createRouteMap (
   if (process.env.NODE_ENV === 'development') {
     // warn if routes do not include leading slashes
     const found = pathList
+    // check for missing leading slash
       .filter(path => path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
+    // split the path to get the root path part only
+      .map(path => path.split('/')[0])
+    // remove duplicates caused by split child paths
+      .filter((path, index, pathList) => pathList.indexOf(path) === index)
 
     if (found.length > 0) {
-      warn(false, `Non-nested routes must include a leading slash character. Replace "${found[0]}" with "/${found[0]}".`)
+      warn(false, `Non-nested routes must include a leading slash character. Fix the following routes: ${found.join('\n')}.`)
     }
   }
 
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -34,14 +34,14 @@ export function createRouteMap (
     }
   }
 
-  // warn if routes do not include leading slashes
-  const pathsMissingSlashes = pathList
-    .filter(path => !!path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
-    .map(path => path.split('/')[0])
-    .filter((path, index, pathList) => pathList.indexOf(path) === index)
+  if (process.env.NODE_ENV === 'development') {
+    // warn if routes do not include leading slashes
+    const found = pathList
+      .find(path => path.charAt(0) !== '*' && path.charAt(0) !== '/')
 
-  if (pathsMissingSlashes.length > 0) {
-    warn(false, `The following routes require a leading slash in their paths: ${pathsMissingSlashes.join(', ')}`)
+    if (found) {
+      warn(false, `Non-nested routes must include a leading slash character. Replace "${found}" with "/${found}".`)
+    }
   }
 
   return {
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -39,13 +39,10 @@ export function createRouteMap (
     const found = pathList
     // check for missing leading slash
       .filter(path => path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
-    // split the path to get the root path part only
-      .map(path => path.split('/')[0])
-    // remove duplicates caused by split child paths
-      .filter((path, index, pathList) => pathList.indexOf(path) === index)
 
     if (found.length > 0) {
-      warn(false, `Non-nested routes must include a leading slash character. Fix the following routes: ${found.join('\n')}.`)
+      const pathNames = found.map(path => `- ${path}`).join('\n')
+      warn(false, `Non-nested routes must include a leading slash character. Fix the following routes: \n${pathNames}`)
     }
   }
 
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -34,6 +34,16 @@ export function createRouteMap (
     }
   }
 
+  // warn if routes do not include leading slashes
+  const pathsMissingSlashes = pathList
+    .filter(path => !!path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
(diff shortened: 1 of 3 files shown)


posva (changes requested):
Thanks! We, however, require a test case for this (showing warning vs not showing). The warning should only appear in development mode and it should display more information like the path that may be incorrect. You can check other warnings for inspiration

Sayegh7:
@posva Added test cases, moved the change to `create-route-map`, and included the problematic routes in the warning message

src/create-route-map.js
```diff
@@ -34,6 +34,16 @@ export function createRouteMap (
     }
   }
 
+  // warn if routes do not include leading slashes
+  const pathsMissingSlashes = pathList
```
posva: can you wrap this whole thing so it only gets executed in development?

src/create-route-map.js
```diff
@@ -34,6 +34,16 @@ export function createRouteMap (
...
+    .filter(path => !!path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
```
posva: I think a single `find` is simpler. No need for the map and I don't really understand the last `filter` call
We can also remove the `!!path` check as it's a string

src/create-route-map.js
```diff
@@ -34,6 +34,16 @@ export function createRouteMap (
...
+    .map(path => path.split('/')[0])
+    .filter((path, index, pathList) => pathList.indexOf(path) === index)
+
+  if (pathsMissingSlashes.length > 0) {
+    warn(false, `The following routes require a leading slash in their paths: ${pathsMissingSlashes.join(', ')}`)
```
posva: Printing the one route we found is enough.

```suggestion
    warn(false, `Non-nested routes must include a leading slash character. Replace "${found)" with "/${found}".`)
```

test/unit/specs/create-map.spec.js
```diff
@@ -97,6 +97,48 @@ describe('Creating Route Map', function () {
     expect(console.warn.calls.argsFor(0)[0]).toMatch('vue-router] Duplicate param keys in route with path: "/foo/:id/bar/:id"')
   })
 
+  it('in development, warn if a path is missing a leading slash', function () {
+    process.env.NODE_ENV = 'development'
+    maps = createRouteMap([
+      { path: '/', name: 'home', component: Home },
+      { path: 'bar', name: 'bar', component: Bar },
```
posva: for this test, bar is enough. We don't even need '/'

test/unit/specs/create-map.spec.js
```diff
@@ -97,6 +97,48 @@ describe('Creating Route Map', function () {
...
+          { path: 'bar/:id', component: Bar }
+        ]
+      },
+      { path: '*', name: 'any', component: Baz }
+    ])
+    expect(console.warn).toHaveBeenCalledTimes(1)
+    expect(console.warn.calls.argsFor(0)[0]).toEqual('[vue-router] The following routes require a leading slash in their paths: bar, foo')
+  })
+
+  it('in development, it has not logged a missing leading slash warning when all paths have slashes', function () {
```
posva:
```suggestion
  it('in development, it does not log the missing leading slash when routes are valid', function () {
```

test/unit/specs/create-map.spec.js
```diff
@@ -97,6 +97,48 @@ describe('Creating Route Map', function () {
...
+        children: [
+          { path: 'bar/:id', component: Bar }
+        ]
+      },
+      { path: '*', name: 'any', component: Baz }
+    ])
+    expect(console.warn).not.toHaveBeenCalled()
+  })
+
+  it('in production, it has not logged a missing leading slash warning', function () {
```
posva:
```suggestion
  it('in production, it does not log the missing leading slash warning', function () {
```

test/unit/specs/create-map.spec.js
```diff
@@ -97,6 +97,48 @@ describe('Creating Route Map', function () {
...
+    process.env.NODE_ENV = 'production'
+    maps = createRouteMap([
+      { path: '/', name: 'home', component: Home },
+      { path: 'bar', name: 'bar', component: Bar },
```
posva: only bar is enough here (1 route)

posva (changes requested):

src/create-route-map.js
Sayegh7: `find` is much simpler but unfortunately I do not think it's supported by this version of nightwatch which causes e2e tests to fail.

posva: you are totally right! Thank you for catching that, it would fail on IE as well. Then let's keep the `filter` and display the list of failing routes since the information is there already. If the routes could be displayed one per line, that would be perfect as it would be more readable (I don't remember how it was before)

Sayegh7: It used to be just comma separated, but one per line sounds better

src/create-route-map.js
```diff
@@ -34,6 +34,21 @@ export function createRouteMap (
...
+    const found = pathList
+    // check for missing leading slash
+      .filter(path => path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
+    // split the path to get the root path part only
+      .map(path => path.split('/')[0])
+    // remove duplicates caused by split child paths
+      .filter((path, index, pathList) => pathList.indexOf(path) === index)
+
+    if (found.length > 0) {
+      warn(false, `Non-nested routes must include a leading slash character. Fix the following routes: ${found.join('\n')}.`)
```
posva: the list should be more like

```
[...] routes:
- bar
- foo
```

test/unit/specs/create-map.spec.js
```diff
@@ -97,6 +97,34 @@ describe('Creating Route Map', function () {
...
+      { path: '/bar', name: 'bar', component: Bar }
+    ])
+    expect(console.warn).not.toHaveBeenCalled()
+  })
+
+  it('in production, it does not log the missing leading slash warning', function () {
+    process.env.NODE_ENV = 'production'
+    maps = createRouteMap([
+      { path: '/', name: 'home', component: Home },
+      { path: 'bar', name: 'bar', component: Bar },
```
posva: only `bar` is enough in this test

src/create-route-map.js
```diff
@@ -34,6 +34,21 @@ export function createRouteMap (
...
 
+  if (process.env.NODE_ENV === 'development') {
+    // warn if routes do not include leading slashes
+    const found = pathList
+    // check for missing leading slash
+      .filter(path => path && path.charAt(0) !== '*' && path.charAt(0) !== '/')
+    // split the path to get the root path part only
+      .map(path => path.split('/')[0])
+    // remove duplicates caused by split child paths
+      .filter((path, index, pathList) => pathList.indexOf(path) === index)
```
posva: I don't think this method is good as it will remove remaining path from `/some/longer/url` when it doesn't have children. So I think there should be some kind of mixing with the `pathMap` to make it display the same path that is displayed with the record, otherwise displaying the whole path seems simpler and easier to spot, even if it could lead to multiple paths when only one record is affected

posva (changes requested):
We are also missing the warning for `addRoutes`. Could you add it as well?

Sayegh7:
@posva Are you referring to the `addRoutes` function used by the matcher? If so, I believe that this function already calls `createRouteMap` which would trigger the warning message.

posva (approved):
True! I forgot it does an in place thing with the same function

Thanks!

PR #2620, opened 2019-02-21
docs: simplify installing from GitHub

pRizz:
Simplify the docs for installing dev from GutHub

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -38,8 +38,7 @@ You will have to clone directly from GitHub and build `vue-router` yourself if
 you want to use the latest dev build.
 
 ``` bash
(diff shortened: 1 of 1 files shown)


posva:
Thank you but that wouldn't be enough as it would require the user to import from `src`

pRizz:
@posva the script still has the `npm run build` step.

`npm install vuejs/vue-router` installs `vue-router` from this GitHub repo into `node_modules`
`cd node_modules/vue-router` is the same
`npm run build` then we build in that directory

posva:
oh, I missed that cd part. I will still have to try

posva:
Thank you for the proposal but I think it's easier to keep a more traditional way of installing a local version

PR #2640, opened 2019-03-05
fix(router): prevent memory leaks by removing app references from $router.apps once app destroyed (Fixes #2639)

### LINKED ISSUE #2639: [bug] Memory leak with this.$router.apps array and secondary apps on page being destroyed
### Version
3.0.2 (and earlier)

### Reproduction link
[https://bootstrap-vue.js.org/play/](https://bootstrap-vue.js.org/play/)

See next steps to easily reproduce the issue.

### Steps to reproduce
Just go to https://bootstrap-vue.js.org/play/

And enter the following code in the HTML Template area:

```js
<div>
  <p>This App UID: {{ _uid }}</p>
  <p>$route.apps length: {{ $router.apps.length }}</p>
  <pre>{{
    $router.apps.map(
      a => {
      	return { uid: a._uid, destroyed: a._isDestroyed }
      }
    ).reverse()
  }}</pre>
</div>
```

and just enter `{ }` into the JavaScript area (or leave it as is)

Hit space / delete the space every second or so in either the HTML entry area, or the javascript entry area (there is a 0.75 second debouncer on the app creation/destroy on keypress).

Each time, the rendered app will be destroyed and recreated.

The app shows a mapped version of `$router.apps` array (in the results section)

You will see the array grow and grow and grow, with all these `destroyed` instances sticking around.

It looks like vue-router isn't cleaning out `$router.apps` array during the destroy hook of the app. (`hook:destroyed').

### What is expected?
That the apps that have been `$destroyed` would be removed from the `$router.apps` array

The destroyed app references remain in the `$router.apps` array, causing memory usage to grow/leak.

---
I've noticed a memory leak with `this.$router.apps` array.  It might be related (but not sure) to other memory leaks reported.

On a docs site we have, we have a playground that allows you to create and test mini apps in real time.

As the user edits the test app, it destroys (app.$destroy()), and then re compiles and mounts the new version.

Each time the playground app is created, it gets appended to the `this.$router.apps` array, while the destroyed app still remains in the array (albeit it has a destroyed instance, but there is still quite a bit of stuff in the reference).

The playground is running under Nuxt.js, and each test in the playground is created as a new app (with a reference to the main $router instance).

We are adding in a temporary fix that will scan the `$router.apps` array, and slice out the destroyed app reference.

tmorehouse:
Prevent memory leaks by removing app(s) from `$router.apps` array once app has been destroyed.

Fixes #2639

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -91,7 +91,7 @@ export default class VueRouter {
 
     // main app already initialized.
     if (this.app) {
-     app.$once('hook:destroyed', () {
+     app.$once('hook:destroyed', () => {
        // Clean out app from this.apps array once destroyed
        const index = this.apps.indexOf(app)
        if (index > -1) this.apps.splice(index, 1)
--- a/src/index.js
+++ b/src/index.js
@@ -89,13 +89,16 @@ export default class VueRouter {
 
     this.apps.push(app)
 
+    // Clean out app from this.apps array once destroyed
+    app.$once('hook:destroyed', () => {
+      const index = this.apps.indexOf(app)
+      if (index > -1) this.apps.splice(index, 1)
+      // Ensure we still have a main app
+      if (this.apps[0]) this.app = this.apps[0]
+    })
+
     // main app already initialized.
     if (this.app) {
-      app.$once('hook:destroyed', () => {
-        // Clean out app from this.apps array once destroyed
-        const index = this.apps.indexOf(app)
-        if (index > -1) this.apps.splice(index, 1)
-      })
       return
     }
 
--- a/src/index.js
+++ b/src/index.js
@@ -91,7 +91,12 @@ export default class VueRouter {
 
     // main app already initialized.
     if (this.app) {
-      return
+     app.$once('hook:destroyed', () {
+       // Clean out app from this.apps array once destroyed
+       const index = this.apps.indexOf(app)
+       if (index > -1) this.apps.splice(index, 1)
+     })
+     return
     }
 
     this.app = app
--- a/src/index.js
+++ b/src/index.js
@@ -91,12 +91,12 @@ export default class VueRouter {
 
     // main app already initialized.
     if (this.app) {
-     app.$once('hook:destroyed', () => {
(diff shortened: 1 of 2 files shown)


tmorehouse:
Not quite sure why the netlify deploy is failing... Maybe the removal needs to happen in a `$nextTick`?

src/index.js
```diff
@@ -91,6 +91,11 @@ export default class VueRouter {
 
     // main app already initialized.
     if (this.app) {
+      app.$once('hook:destroyed', () => {
```
posva: if you add it inside the if, creating an app and then destroying the main app after it would still result in a memory leak, right?

Could you add some tests to api.spec.js or another new file checking the apps array is emptied correctly, please?

posva (changes requested):

src/index.js
tmorehouse: The main app, would typically be destroyed last (I would think)

The reason for placing it inside of the if would be that it would t destroy the main app, only apps that came along after the main app was instantiated.

tmorehouse: I'll see what I can do a bout adding a few tests.

tmorehouse: Another question would be, what if the main app is destroyed  y the other secondary apps are still alive? Should the main app be removed as well, and should the next app in the apps array (with the lowest index number) be copied to `this.app` ?

tmorehouse:
@posva I have added test suites, and handled situation where the main (initial) app is destroyed.

tmorehouse:
Just wondering if there should be additional test(s) for a nested router (i.e. a second `router` created)?

posva:
What do you mean by nested router?

tmorehouse:
@posva I see in the e2e tests there are tests for nested routers (a router within a router):

https://github.com/vuejs/vue-router/blob/dev/test/e2e/specs/nested-router.js

https://github.com/vuejs/vue-router/tree/dev/examples/nested-router

Although the multiple routers share the same Router class instance, so probably additional testing isn't necessary.

posva:
Closing in favor of #2706

PR #2646, opened 2019-03-12
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort, we noticed that vue-router could not be built reproducibly.

This is due to it embedding the current build year into the file headers. This has been replaced with using [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) where available. This was originally filed in Debian as [#924378](https://bugs.debian.org/924378).

### DIFF
--- a/build/configs.js
+++ b/build/configs.js
@@ -5,10 +5,11 @@ const cjs = require('rollup-plugin-commonjs')
 const node = require('rollup-plugin-node-resolve')
 const replace = require('rollup-plugin-replace')
 const version = process.env.VERSION || require('../package.json').version
+const now = new Date(process.env.SOURCE_DATE_EPOCH ? (process.env.SOURCE_DATE_EPOCH * 1000) : new Date().getTime());
 const banner =
 `/*!
   * vue-router v${version}
-  * (c) ${new Date().getFullYear()} Evan You
+  * (c) ${now.getFullYear()} Evan You
   * @license MIT
   */`
 


posva:
hey, I don't get this, in what situation current code isn't working?

lamby:
> hey, I don't get this, in what situation current code isn't working?

It builds differently depending on which year you happen to be building `vue-router` in. :)

posva:
yeah, that's the point, to write current year. In what situation is `new Date().getFullYear()` not working?

lamby:
It's not that it "doesn't work", it is that it generates a different result depending on the year. This makes the build non-reproducible which is important for a variety of reasons - see the links in my original post! :)

posva:
I appreciate your concern but you are still not explaining things to me...
We **want** the year to be different based on when we run the script

lamby:
I am not going into great depth here, but I would not want to bore you by explaining things here that you might have read elsewhere already… for example https://reproducible-builds.org/docs/buy-in/ etc.

posva:
It's the first time I've heard about reproducible builds and I have read just a bit with the links you sent me but I really don't see what attack this is preventing. I will rather see the opposite as it is using an env variable that doesn't seem to be standard in every unix os (it doesn't exist in osx). Maybe there is a problem for other languages or libraries but I don't see what is being prevented here, I just see the code becoming more complicated for no benefit. Don't worry about copy pasting things that I may know, I won't feel bothered or anything, I get your intentions seem good

lamby:
Unfortunately my time is really stretched so you can imagine that I cannot spend 1 hour on each pull request explaining the background and justify the reasons behind reproducible builds. Perhaps you could spend a few moments reviewing some slides/presentations [on our Resources](https://reproducible-builds.org/resources/) page. Again, really sorry I cannot provide a custom justification to every single project...

posva:
It's okay, I understand

lamby:
> It's okay, I understand

Cool; fancy applying the patch in that case? :)

PR #2664, opened 2019-03-22
feat(core): support add routes as children to an existing route

Hquestion:
VueRouter.addRoutes method support add routes as children to an existing route


Sapphire2k:
That's pretty cool!!

posva:
Added at ca80c44

PR #2665, opened 2019-03-22
docs: fix typo

Scrum:


posva:
this was correct

Scrum:
> this was correct

@posva not, i get error
```log
Uncaught TypeError: Cannot read property '$router' of null
```

PR #2689, opened 2019-03-30
fix(router-view): register instance in init hook (fix #2561)

### LINKED ISSUE #2561: beforeRouteEnter hook's cb is not called (same component across diffrent routes and keep-alive applied )
### Version
3.0.2

### Reproduction link
[https://jsfiddle.net/9vctongd/1/](https://jsfiddle.net/9vctongd/1/)

### Steps to reproduce
1. open the JSFiddle provided, open the console
2. click link GO TO FOO1
3. click link GO TO BAR
4. click link GO TO FOO2

### What is expected?
console prints:

foo enter

foo enter cb

foo enter

foo enter cb

console prints:

foo enter

foo enter cb

foo enter

---
also results in an infinite function call (function poll in base.js)

zrh122:
close #2561

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -25,6 +25,7 @@
     <li><a href="auth-flow">Auth Flow</a></li>
     <li><a href="discrete-components">Discrete Components</a></li>
     <li><a href="nested-router">Nested Routers</a></li>
+    <li><a href="navigation-guard-callback">Navigation Guard Callback</a></li>
   </ul>
 </body>
 </html>
--- a/examples/navigation-guard-callback/app.js
+++ b/examples/navigation-guard-callback/app.js
@@ -1,0 +1,74 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+Vue.use(VueRouter)
+
+const Foo = {
+  template: '<div class="current">foo</div>',
+  beforeRouteEnter (to, from, next) {
+    // in-component beforeRouteEnter hook
+    next(vm => {
+      // print log
+      app.logText = 'beforeRouteEnter callback emited'
+    })
+  }
+}
+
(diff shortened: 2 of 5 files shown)


posva:
Hey could you add a test for this?

An e2e test should do it

zrh122:
Yes, i will.

zrh122:
@posva , e2e test added.

posva (commented):
Thanks, I'm merging this locally after applying some modifications (splitting the e2e test to keep the navigation guard one in an existing e2e test and the keep alive in another

posva:
Merged locally!

zrh122:
@posva  the e2e test for #2561 may be not correct. I removed the code ([register instance in init hook](https://github.com/vuejs/vue-router/blob/5ad453995bd1d9259e058d4b1ad6428e825255da/src/components/view.js#L77)), and executed `npm run test:e2e`, all the test case was passed. It should not be passed as i expect.

mine:
```javascript
// routes
[
  // foo1
  { path: '/foo1', component: Foo },

  // foo2, same component as foo1
  { path: '/foo2', component: Foo },

  // bar
  { path: '/bar', component: Bar }
]
```
click order: `/foo1` -> `/bar` -> `/foo2`

yours:
```javascript
// routes
[  
  {
    path: '/with-guard1',
    name: 'with-guard1',
    component: WithGuard
  },
  {
    path: '/with-guard2',
    name: 'with-guard2',
    component: WithGuard
  }
]
```
click order: `/with-guard1` -> `/with-guard2` -> `/with-guard1`

maybe you can add another route, and the click order can be:

 `/with-guard1` -> `/another` -> `/with-guard2`

posva:
Right! Thanks for checking, I'm going to fix the test

PR #2706, opened 2019-04-11
fix(router): prevent memory leaks by removing app references

### LINKED ISSUE #2639: [bug] Memory leak with this.$router.apps array and secondary apps on page being destroyed

posva:
Fix #2639

Hey, @tmorehouse, I refactored the tests and reverted the code added to the if because I couldn't see why we would override the main app. Could you take a look please?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -90,21 +90,23 @@ export default class VueRouter {
     this.apps.push(app)
 
     // set up app destroyed handler
+    // https://github.com/vuejs/vue-router/issues/2639
     app.$once('hook:destroyed', () => {
       // clean out app from this.apps array once destroyed
       const index = this.apps.indexOf(app)
       if (index > -1) this.apps.splice(index, 1)
-      // ensure we still have a main app, unless this is the last app left
-      if (this.apps.length > 0) this.app = this.apps[0]
+      // ensure we still have a main app
+      if (this.app === app) this.app = this.apps[0] || null
+      // no more apps, the router can be released
+      if (!this.app) {
+        // TODO: destroy router, listeners?
+        // Maybe the user wants to reuse the router though
+      }
     })
 
     // main app previously initialized
+    // return as we don't need to set up new history listener
     if (this.app) {
-      if (this.app !== this.apps[0]) {
-        // main app had been destroyed, so replace it with this new app
-        this.app = this.apps[0]
-      }
-      // return as we don't need to set up new history listener
       return
     }
 
--- a/src/index.js
+++ b/src/index.js
@@ -91,7 +91,7 @@ export default class VueRouter {
 
     // main app already initialized.
     if (this.app) {
-     app.$once('hook:destroyed', () {
+     app.$once('hook:destroyed', () => {
        // Clean out app from this.apps array once destroyed
        const index = this.apps.indexOf(app)
        if (index > -1) this.apps.splice(index, 1)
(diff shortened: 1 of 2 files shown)


tmorehouse:
I'll take a look here in a bit.

src/index.js
```diff
@@ -89,7 +89,19 @@ export default class VueRouter {
...
+      // ensure we still have a main app or null if no apps
+      // we do not release the router so it can be reused
+      if (this.app === app) this.app = this.apps[0] || null
+    })
+
+    // main app previously initialized
+    // return as we don't need to set up new history listener
     if (this.app) {
       return
     }
```
tmorehouse, line 107: With `this.app` being set to null (if all apps are cleared out), this early return doesn't happen, and the code below it will be executed again.

Will that be an issue? IF so, maybe a flag needs to be set so that the following code is not executed a second time.

posva: but this is called when a `router` is injected in a created component, it's only executed once per root instance

tmorehouse: Doesn't `init` get called for each app instance, as it is how the apps get added to `this.apps` array:

```js
  init (app: any /* Vue component instance */) {
    process.env.NODE_ENV !== 'production' && assert(
      install.installed,
      `not installed. Make sure to call \`Vue.use(VueRouter)\` ` +
      `before creating root instance.`
    )

    this.apps.push(app)
```

posva: yes, but once per root instance

tmorehouse: Ah, correct... so it should be good.

Issue #2728, opened 2019-04-22
beforeRouterEnter next function callback does not follow the same lifecycle pattern that Vue does


itmayziii:
### Version
3.0.1

### Reproduction link
[https://codesandbox.io/embed/n48n7jpq6j](https://codesandbox.io/embed/n48n7jpq6j)

### Steps to reproduce
1. Add a beforeRouterEnter function that has a callback to a component
2. Make the same component from step 1 use a mixin that also uses beforeRouteEnter with a callback
3. The callback in the component will be called before the callback from the mixin which is inconsistent with how Vue handles the rest of the lifecycle methods. 

Note that the mixin will actually be called before the component (expected behavior) but the callback in the mixin will not be (unexpected.)

### What is expected?
The mixin beforeRouteEnter callback should be called before the components.

### What is actually happening?
The components beforeRouterEnter callback is called before the mixins.

posva:
Yeah, it's inconsistent. Why would this be a problem though? making one of the callbacks depends on another one coming from a mixin is very implicit

itmayziii:
In our case the mixin is handling common logic between different components (which I believe is the main use case for them.)  This common logic does some setup in `beforeRouteEnter` and does some teardown in `beforeRouteLeave`. I would use the `created` lifecycle method but there is no guarantee the router will be going to a new component entirely, it may re-use the current one. In these cases where it re-uses the current component it's not possible to do common setup tasks (that rely on the vue instance) since the component is called first.

If it is acknowledged that this is indeed something Vue router wants to fix I would gladly volunteer to look at the issue.

posva:
okay, I don't think it's a good idea to keep such things in a mixin, but I would make it more consistent to have it in the right order and it will be great if you are willing to help on this! Please, do!

PR #2738, opened 2019-04-24
fix: make callback of next in beforeRouterEnter more consistent

zrh122:
Improve #2728. Make callback of `next` to be called as same order as `beforeRouterEnter`.

### DIFF
--- a/examples/navigation-guards/app.js
+++ b/examples/navigation-guards/app.js
@@ -108,9 +108,18 @@ const NestedParent = {
   }
 }
 
+const GuardMixin = {
+  beforeRouteEnter (to, from, next) {
+    next(vm => {
+      vm.$parent.logs.push('mixin')
+    })
+  }
+}
+
 const NestedChild = {
   props: ['n'],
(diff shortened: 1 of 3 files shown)


posva (changes requested):
Hi, thanks! Could you add a test for this? An e2e test should do it. You could check that a string grows to a specific value (to check the order is right) but if you have a better idea, that's good too!

zrh122:
@posva , I will add a test for it. I intend to add the test in [navigation guards](https://github.com/vuejs/vue-router/blob/dev/test/e2e/specs/navigation-guards.js), but I find that the test code in this file never execute because it just return when find use `PhantomJS`, and currently only use `PhantomJS`.

posva:
I see, I will take a look next week then and keep you posted. Thanks for #2749 BTW!

posva:
As you said, it does fix #2761 too, I added a test for that. I will add another test for the other case too

zrh122:
> As you said, it does fix #2761 too, I added a test for that. I will add another test for the other case too

Thx.

PR #2750, opened 2019-04-28
docs: switch v-if with class for consistency

SachsKaylee:
A small nit that bothered me: v-if and class are in a different order in one of the divs.


posva (approved):
Thanks

PR #2755, opened 2019-05-02
docs: .htaccess and lazy loaded routes (fix #2753)

### LINKED ISSUE #2753: apache .htaccess not working with lazy-loaded routes
### Version
3.0.3

### Reproduction link
[https://beta.pow.cool/members](https://beta.pow.cool/members)

### Steps to reproduce
- Create a new app
- Add a lazy loaded route named 'myRoute'
- Build it
- Check that you have a `myRoute.js` file in your build folder
- Deploy it to an apache server after following the `.htaccess` recommandation in the vue-router docs
- Check that your browser shows the content of the `myRoute.js` file instead of serving the `index.html` file when browsing `http://mysite.com/myRoute`

### What is expected?
The browser shoud display the index.html at the given route.

The browser displays the js content of the javascript file.

---
It works as expected with the following conf (taken from create-react-app):
```
Options -MultiViews
RewriteEngine On
RewriteCond %{REQUEST_FILENAME} !-f
RewriteRule ^ index.html [QSA,L]
```

xseignard:
On apache we need to add more configuration to handle lazy loaded
routes. Updated docs in all available languages.

See `Multiviews` paragraph here: https://httpd.apache.org/docs/2.4/en/content-negotiation.html#negotiation

### DIFF
--- a/docs/fr/guide/essentials/history-mode.md
+++ b/docs/fr/guide/essentials/history-mode.md
@@ -22,8 +22,10 @@ Ne vous inquiétez pas. Pour résoudre ce problème, il vous suffit d'ajouter un
 ### Apache
 
 ```apache
-<IfModule mod_rewrite.c>
+<IfModule mod_negotiation.c>
   Options -MultiViews
+</IfModule>
+<IfModule mod_rewrite.c>
   RewriteEngine On
   RewriteBase /
   RewriteRule ^index\.html$ - [L]
--- a/docs/fr/guide/essentials/history-mode.md
+++ b/docs/fr/guide/essentials/history-mode.md
@@ -23,6 +23,7 @@ Ne vous inquiétez pas. Pour résoudre ce problème, il vous suffit d'ajouter un
 
 ```apache
 <IfModule mod_rewrite.c>
+  Options -MultiViews
   RewriteEngine On
   RewriteBase /
   RewriteRule ^index\.html$ - [L]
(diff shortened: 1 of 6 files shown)


posva:
Thanks for this! I don't use Apache so I will wait for someone else review before merging

phanan:
This looks good, but better yet, we should check for `mod_negotiation`'s availability as well:

```
<IfModule mod_negotiation.c>
    Options -MultiViews
</IfModule>
# The rest of the config
```

xseignard:
Hi @phanan you're right. I'll update the PR.
In the meantime CRA don't bother that much, see the recommended Apache conf of their: https://facebook.github.io/create-react-app/docs/deployment#serving-apps-with-client-side-routing

posva:
I think the other if was meant to be kept, wdyt @phanan ?

hunsalz:
Hey guys, I would appreciate if the PR from @phanan gets accepted. I had the same issue with a user of us and I had to search a while before I found this. I can imagine that a lot of users gave up frustrated before having a solution. Please take a look at this again and push it live! Thx in advance

stsmdt:
As mentioned by @hunsalz we simply used the current example configuration for Apache from the documentation and had issues with the lazy loaded routes then.

Adding `Options -MultiViews` to the configuration solved the problem then. This should be included in the documentation.

PR #2768, opened 2019-05-13
fix(flow,types): fix props type

zrh122:
fix props type.


WofWca:
#2365 has the same purpose.

posva:
Sorry for the delay on this. I added the test case you added to types and it's already passing. Let's pass on this for the moment

PR #2771, opened 2019-05-17
(fix #2607): Fix router.back in abstract mode when 2 consecutive same routes appear in history stack

### LINKED ISSUE #2607: abstract mode's router.back() fails if there are 2 consecutive same routes in history stack
### Version
3.0.2

### Reproduction link
[https://jsfiddle.net/Rocka/bhu5vkn8/3/](https://jsfiddle.net/Rocka/bhu5vkn8/3/)

### Steps to reproduce
1. Click `10. artist`
2. Hit button `togglePlayer`
3. Click `10. artist`
4. Hit button `back` 2 times

### What is expected?
route back to `/search`

nothing happend, route and route-view do not change

---
'cause I didn't figure out how to determine wether the route is going forward or backword, I decided to use `abstract` mode and traverse the router stack to find out which direction the route is going, then hit into this bug. The demo works fine if change router's mode to `hash` or `history`.

Domino9697:
This PR provides a fix to the issue #2607.

Reproduction link

https://jsfiddle.net/Rocka/bhu5vkn8/3/

In abstract mode, if you try to navigate through your history with the same routes appearing consecutively in your stack history, the history index remains unchanged.

In History or Hash mode, this issue does not appear because history navigation is handled by the browser and not by vue-router. In all modes, we do not want to re-render a page if the current route and the next route are identical. However, in the case of abstract mode, we still need to update the history index whenever the current and next routes are the same since AbstractHistory handles its own history stack.

### DIFF
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -37,6 +37,10 @@ export class AbstractHistory extends History {
     this.confirmTransition(route, () => {
       this.index = targetIndex
       this.updateRoute(route)
+    }, (err) => {
+      if (err === 'SAME_ROUTE') {
+        this.index = targetIndex
+      }
     })
   }
 
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -2,10 +2,11 @@
 
 import type Router from '../index'
 import { History } from './base'
+import { NavigationDuplicated } from './errors'
 
 export class AbstractHistory extends History {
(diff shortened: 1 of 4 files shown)


posva:
I rebased against dev and adapted the code to use an Error object

Domino9697:
Thank you @posva !

PR #2772, opened 2019-05-20
(fix #2724): Fix active links when parent link redirects to child

### LINKED ISSUE #2724: Link to default child is not highlighted when another child is active
### Version
3.0.3

### Reproduction link
[https://jsfiddle.net/3svnktq2/1/](https://jsfiddle.net/3svnktq2/1/)

### Steps to reproduce
The setup is that 'parent' is a menu item. When you go to that item, you are redirected to the child state with default parameters. Within that child state you can modify your parameters which should result in keeping the menu item highlighted.

1) Click on the parent link
-> Redirected to child state with type parameter 'type1' and parent is highlighted as active state
2) Click on 'go to child with type2'

### What is expected?
You stay in the same child state with a different parameter, keeping the parent link highlighted as the active state.

'Parent' is no longer highlighted as the active state. It is only highlighted when the type parameter is 'type1'.

---
This happens since version 3.0.3 of vue-router. If you change the loaded version of vue-router to 3.0.2, the behaviour is as expected. More history on this issue: https://github.com/vuejs/vue-router/issues/2717.

Domino9697:
This PR provides a fix to the issue #2724 .

Reproduction link

https://jsfiddle.net/posva/fj7ywzxv/

Whenever a parent node has a redirect option, the router.resolve method will return the redirected route when given the parent location as argument. This creates a problem when comparing a link route with the current route as the link route will be the route of the redirected component and not the route of the parent component...


posva:
Hey, sorry for the delay and thanks for this! I changed a bit the test and added more. The `redirect` option should be on a child. it doesn't really make sense to add it alongside a component option since a static redirect route cannot be reached.

Waiting for tests to pass

Domino9697:
Thank you !

Issue #1187, opened 2017-02-22
Need a way to set what scroll position is saved


brad426:
Similar to the way you can define a `scrollBehavior` function on a router instance, I need a way to define how the `savedPosition` value gets computed/populated. I need to override the default behavior which stores the window's scroll position, and instead store the scroll position of another element.

My application uses a structure where the `document`/`window` stays static, and content scrolls inside of a container. Example: http://codepen.io/brad426/pen/pezRge

LinusBorg:
Hi Brad.

just thinking out loud:

I think this won't be so easy to do since those two things - the scrollposition the router saves right now, and the scrollpostion of an arbitrary element - have not much in common, technically. One works with the `popstate` event and `push()` method  of `window.history`, while the other requires to read and write to a certain DOM elements property.

Also, this functionality is not necessarily tied to router components, but could be useful to all sorts of components, which makes me think that it would be better to do this in a generic plugin/component ...

posva:
The `scrollBehavior` could be an object instead of a function but as @LinusBorg I'm wondering if keeping this behaviour in a component isn't more flexible and reusable

brad426:
Cool, thanks for your feedback.
I agree that this functionality would be useful to all sorts of components, but I think a lot of people might also find a "global" override useful.

Regarding my [example app structure](http://codepen.io/brad426/pen/pezRge), I'd like to keep the router's functionality exactly as is, i.e. have it call `saveScrollPosition` whenever a `popstate` event is triggered, the only change would be to allow a custom handler for `saveScrollPosition` . Example:

``` javascript
const router = new VueRouter({
  mode: 'history',
  routes,
  saveScrollPosition() {
     let el = document.querySelector('.page-content')
     return { x: el.scrollLeft, y: el.scrollTop }
  }
})
```

erguotou520:
@LinusBorg see my issue's program #1249 , is that possible to realize?

blake-newman:
IMO, we should set a scrollEl property in main router config. Thus allowing us to use a specific scroll container for both scrollBehavior and scroll position. 

What's everyone's opinion on this?

Dadibom:
You don't need special functions for this, before/afterEach would be a much better place to put these kinds of things

unamix:
It would be nice if we could somehow define from which element we want to save the scroll position. 
In the meantime i solved it this way:

```
<keep-alive>
  <router-view></router-view>
</keep-alive>
```

```
const router = new VueRouter({
[...]

PR #2780, opened 2019-05-22
feat(scroll) - WIP - allow a different scroll element other than window

### LINKED ISSUE #2777: scrollBehavior should allow configuring scrolling element
A lot of times - the scrolling element in the document is not the window, making scrollBehavior not work (and the saved position too probably). 

Should allow setting the scrolling element in the application (not sure if as a global, or per scrollBehavior function or both).

##### Global Option:

When initiating Vue router - allow setting scrollingElement property that will replace window.

##### Per Scroll Behavior Option:

When returning the scroll object should allow to set scrollingElement that will override the window (or the global config).

savedPosition should always be related to the global config (or window).

phouri:
Still a WIP - just to make sure I am on the right track before fixing flow / e2e / unit tests.

This works (for me locally), let me know what you think.

Fixes #2777

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -28,7 +28,7 @@ declare type NavigationGuard = (
 declare type AfterNavigationHook = (to: Route, from: Route) => any
 
 type Position = { x: number, y: number };
-type PositionResult = Position | { selector: string, offset?: Position } | void;
+type PositionResult = Position | { selector: string, offset?: Position, scrollElement?: string } | void;
 
 declare type RouterOptions = {
   routes?: Array<RouteConfig>;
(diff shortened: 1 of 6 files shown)


phouri:
First time I am using Flow - so probably did some stuff wrong, let me know if you see anything obvious.

If this is fine - I will proceed with fixing flow and creating new tests.

posva:
Thanks for working on this
ref https://github.com/vuejs/vue-router/issues/1187

I will come back to you when I can to review the api, this is the kind of thing that would nicely fit an RFC :)

phouri:
Will start working on an RFC right away :)

phouri:
https://github.com/vuejs/rfcs/pull/38

phouri:
Any news?

scoutrul:
Hello! I just need this functionality.
Thank you.

DamianGlowala:
@posva, can we expect this pull request to be brought into life and reconsidered? Very, very useful feature I cannot find a workaround to at the moment :/

PR #2781, opened 2019-05-22
fix(match): remove if statement that is no need

liyangworld:
At the line 39, already checked the "record", so there is no need to check "record" again.


posva (approved):
True! Thank you

PR #2793, opened 2019-06-01
False positive incorrect type. What can be passed as param on `this.$router.push` unnecessarily limited to string

parml:
**BEFORE:**
  The implementation assumes that only a sting can be passed as a prop
```
this.$router.push({
        name: 'route',
        params: {
            propName: 'a string', // <= OK
            propNameNumber: 5, // <= WRONG
            propNameArray: [5, 'string', {id: 4, color: MyEnum.RED], // <= WRONG
        }
});
```

Actually Vue.js does allow for types that the offending .d.ts disallows.

**AFTER:**
  False positive cases accepted as correct.

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -109,7 +109,7 @@ export interface Location {
   path?: string;
   hash?: string;
   query?: Dictionary<string | (string | null)[] | null | undefined>;
-  params?: Dictionary<string>;
+  params?: Dictionary<any>;
   append?: boolean;
   replace?: boolean;
 }


posva:
Hi, thanks for this but the types are right, for the moment there is no casting, so one should always provide strings because values will come as strings if the user visits directly from the url

parml:
> Hi, thanks for this but the types are right

It is a fact that Vue Router does accept passing numbers, arrays, objects as props. 
As such, Vue Router treats the Router component as a regular component. 

The addressed type file imposes this restriction: a route component should only be allowed to be passed strings. As stated above this restriction does not come from Vue Router. 

Only one of the following can hold:
- Vue Router is wrong and the types are right
- Vue Router is correct and the types are wrong.

I don't understand what you mean by _there is no casting_, however I have a perfectly working passing of non-strings and a typescript error. 
If I understand correctly, a type system will issue an error on a piece of code if that piece of code will not work. Here it issues an error on a piece of code that works perfectly well.

PR #2798, opened 2019-06-04
chore(release): add a missing word in message

briwa:
Hello, just came across this code and I noticed that it's missing a word. Not sure if this is intentional but just putting it here. Thanks!


posva (approved):
Ah, yeah it was missing, thanks!

PR #2802, opened 2019-06-05
fix(normalizeLocation): add a copy for params with named locations (fix #2800)

### LINKED ISSUE #2800: Dynamic segments for nested items not updating correctly
### Version
3.0.3

### Reproduction link
https://jsfiddle.net/zrh122/3v4cj0qk/

### Steps to reproduce

- Click on Item 1
- Click on Item 2
- Click on type link

### What is expected?

We end up at /item/2/logs

We end up at /item/1/logs

---

This behavior happens when we don't explicitly set a parent param for a child route. Setting $route.params.id before rendering the <router-link> fixes the behavior, but this shouldn't be necessary as we already navigated to /item/2, but vue-router didn't update the dynamic segment.

zrh122:
Fix #2800

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -17,6 +17,7 @@
     <li><a href="redirect">Redirect</a></li>
     <li><a href="route-props">Route Props</a></li>
     <li><a href="route-alias">Route Alias</a></li>
+    <li><a href="route-params">Route Params</a></li>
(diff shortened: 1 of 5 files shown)


posva (approved):
Thanks!

PR #2807, opened 2019-06-11
test: fix running e2e in Firefox

zrh122:
1. make `npm run test:e2e:ff` working as expected, [related info](https://github.com/nightwatchjs/nightwatch/issues/2128).
2. add `-e chrome` as default env to nightwatch (`npm run test:e2e` with chromeOptions config applied).
3. remove useless code and comments.


posva:
Hey, thank you but there a part from the removal of `args.push` I don't think the rest is necessary. I'm adding scripts just as a playground right now anyway, I will run tests on CI on the long run like the one for ie9

Issue #2827, opened 2019-06-26
maybe getMatchedComponents type is missed


kahirokunn:
### Version
3.0.3

### Reproduction link
[https://github.com/vuejs/vue-router/blob/732c8258258608c2e29268cbda2a4ed957d884d4/types/router.d.ts#L31](https://github.com/vuejs/vue-router/blob/732c8258258608c2e29268cbda2a4ed957d884d4/types/router.d.ts#L31)

### Steps to reproduce
Maybe this `getMatchedComponents` type is incorrect.
https://github.com/vuejs/vue-router/blob/732c8258258608c2e29268cbda2a4ed957d884d4/types/router.d.ts#L31

### What is expected?
I expected correct `getMatchedComponents` type

### What is actually happening?
I try to read vue-router code.
But code is too long.
So I don't have a much confidence.
I'm sorry if the point is wrong.

I read this source code.

https://github.com/vuejs/vue-router/blob/732c8258258608c2e29268cbda2a4ed957d884d4/src/util/resolve-components.js#L13

there has this sentence.

> we want to halt the navigation until the incoming component has been resolved.

And I read some code.
I guess `getMatchedComponents` function never return asyncComponent.

https://github.com/vuejs/vue-router/blob/732c8258258608c2e29268cbda2a4ed957d884d4/src/index.js#L173

So I think this `getMatchedComponents` type is incorrect.
https://github.com/vuejs/vue-router/blob/732c8258258608c2e29268cbda2a4ed957d884d4/types/router.d.ts#L31

kahirokunn:
Is anyone here? (´；ω；｀)

posva:
See https://github.com/vuejs/vue-router/pull/2828#issuecomment-507022544

PR #2828, opened 2019-06-26
maybe getMatchedComponents type is missed #2827

kahirokunn:
This pull request for solved this issue.
https://github.com/vuejs/vue-router/issues/2827

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -28,7 +28,7 @@ export declare class VueRouter {
   go (n: number): void;
   back (): void;
   forward (): void;
-  getMatchedComponents (to?: RawLocation | Route): Component[];
+  getMatchedComponents (to?: RawLocation | Route): (ComponentOptions<Vue> | typeof Vue)[];
   onReady (cb: Function, errorCb?: ErrorHandler): void;
   onError (cb: ErrorHandler): void;
   addRoutes (routes: RouteConfig[]): void;
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -28,7 +28,7 @@ export declare class VueRouter {
   go (n: number): void;
   back (): void;
   forward (): void;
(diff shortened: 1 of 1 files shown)


types/router.d.ts
```diff
@@ -28,7 +28,7 @@ export declare class VueRouter {
   go (n: number): void;
   back (): void;
   forward (): void;
-  getMatchedComponents (to?: RawLocation | Route): Component[];
+  getMatchedComponents (to?: RawLocation | Route): (ComponentOptions<Vue> | typeof Vue)[];
```
```suggestion
  getMatchedComponents (to?: RawLocation | Route): ComponentOptions<Vue>[];
```

it will never be Vue, so we can remove that as well

posva (changes requested):

types/router.d.ts
kahirokunn: Thx review!

is `Vue.extend({})` never come here?
I thought ExtentedVue will come here.
like this.

```vue:component
<template>
  <div></div>
</template>

<script lang="ts">
import Vue from "vue";
export default Vue.extend({});
</script>
```

```js:router
import Vue from 'vue'
import Router from 'vue-router'

Vue.use(Router)

export const createRouter = () => {
  return new Router({
    mode: 'history',
    routes: [
      { path: '/', component: () => import('@/pages/Index.vue') },
    ]
  })
}
```

kahirokunn: I just try pushed code as said.

posva: I totally overlooked the `Vue.extend` version of pages. The current implementation is correct, as we cannot transform any component option, so we leave them as they are.
Sorry about that

posva:
The current implementation is correct, as we cannot transform any component option, so we leave them as they are, it can be an async function, a Vue.extend (a constructor) or a plain object
Sorry noticing it late!

PR #2838, opened 2019-06-30
Handle Promise rejection in guards (fix #2833)

### LINKED ISSUE #2833: Handle Promise rejection in guards
### Version
3.0.3

### Reproduction link
[https://jsfiddle.net/orzwg537/](https://jsfiddle.net/orzwg537/)

### Steps to reproduce
Run the fiddle and look in console

### What is expected?
a message saying error caught

There is an Unhandled Promise rejection instead

---
Currently, errors thrown in guards are caught by router.onError but Promises rejections (or error thrown in an `async` function) are not caught. This should be handled in all the other guards as well. To handle them right now it is possible by passing the error to `next`:

```
try {
  await fetchdata()
} catch (error) {
  next(error)
}
```

oleksiikhr:
#2833

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
   "presets": ["env", "flow-vue"],
-  "plugins": ["syntax-dynamic-import"]
+  "plugins": ["syntax-dynamic-import", "transform-runtime"]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
   "presets": ["env", "flow-vue"],
-  "plugins": ["syntax-dynamic-import", "transform-runtime"]
+  "plugins": ["syntax-dynamic-import"]
 }
--- a/examples/error-handling/app.js
+++ b/examples/error-handling/app.js
@@ -3,42 +3,90 @@ import VueRouter from 'vue-router'
 
 Vue.use(VueRouter)
 
+// eslint-disable-next-line no-unused-vars
+let asyncMode = false
+
+const makeError = (msg) => {
+  if (asyncMode) {
+    return new Promise((resolve, reject) => {
+      reject(new Error(msg))
+    })
+  }
+
+  throw new Error(msg)
+}
+
 // Components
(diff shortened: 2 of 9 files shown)


posva (changes requested):
Hi, thanks for the PR!
However, we don't use babel to transpile the lib, so this won't work (it is used with tests). Could you find a version without adding the babel plugin dependency?
Can you also add the tests for the others navigation guards, so they are all covered? beforeEnter, beforeRouteEnter, beforeRouteUpdate
Also, could you please add back the template for the PR? It helps us to organise things

Thanks

oleksiikhr:
Very interesting, solved without a library :)

> Also, could you please add back the template for the PR? It helps us to organise things

What is a template?

posva:
Hey! I realized you closed the issue. Did something happen? Maybe it was closed by accident?

Eduardo San Martin Morote

oleksiikhr:
I studied this code, found 2 places (`hook`) where **Promise** can be accepted (`beforeEach`, `afterEach`):

https://github.com/vuejs/vue-router/pull/2838/files#diff-feafce5e816936212eccd6808c5421a6R123
https://github.com/vuejs/vue-router/pull/2838/files#diff-feafce5e816936212eccd6808c5421a6R185

And if in the first case the execution of the code, the sequence itself does not change, in the second case it is not so (R185).

If we take into account that everything happens through callbacks, then in the function `confirmTransition`, I also need to transfer to `onComplete` the callback also the `this._abort` function. In the **abstract.js**, **base.js** files, receive and transmit the `updateRoute` function.

In general, I do not like it, I understand that most likely I need to wrap parts of the code in **Promise**, do consistent error handling and not break the original logic.

Since this PR has errors in logic, I decided to close it, so that suddenly due to inactivity, they would not begin to accept it :)

posva:
Thanks for the explanation. I wanted to make sure it wasn't an accident :)
With current codebase it's a bit harder to implement indeed

PR #2841, opened 2019-07-07
docs: update active class explanation

debs-obrien:

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -101,7 +101,10 @@ In this case the `<a>` will be the actual link (and will get the correct `href`)
 - type: `string`
 - default: `"router-link-active"`
 
-  Configure the active CSS class applied when the link is active. Note the default value can also be configured globally via the `linkActiveClass` router constructor option.
+  The CSS class is applied when the link is active. You can then style this in your CSS using the class name `router-link-active`.  Note the default value can also be configured globally via the `linkActiveClass` property in your router file meaning your can name this class anything you want and your new class name will be applied when the link is active.
+ ``` js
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -101,7 +101,10 @@ In this case the `<a>` will be the actual link (and will get the correct `href`)
 - type: `string`
 - default: `"router-link-active"`
 
-  Configure the active CSS class applied when the link is active. Note the default value can also be configured globally via the `linkActiveClass` router constructor option.
+  The CSS class is applied when the link is active. You can then style this in your CSS using the class name `router-link-active`.  Note the default value can also be configured globally via the `linkActiveClass` property in your router file meaning your can name this class anything you want and your new class name will be applied when the link is active.
+ ``` js
```
posva, line 105: I would rather go back to the previous version and show an example the result.

```vue
<router-link>
<!-- renders -->
<a ...>
<router-link active-class="is-active">
<!-- renders -->
<a ...>
```

Additionally link the `linkActiveClass` to the page on the docs because adding it here only makes it more confusing

posva (changes requested):

docs/api/README.md
debs-obrien: @posva sorry, this confuses me even more. my example is globally changing the active class in the router. Your example is manually changing it in a specific link. And when you say link this to the page on the docs well i thought this was the page of the docs that defines this as i didnt find it anywhere else or am i missing something?

posva: This is the router-link api section, it can link to the Router config section. That's also why it's more confusing to show an example about `linkActiveClass`

Issue #1976, opened 2018-01-09
Expose RouterLink and RouterView components


eddyerburgh:
### What problem does this feature solve?
Currently it's not possible to find RouterLink and RouterView when using vue-test-utils. You can find the rendered content, but if you want to check that a RouterLink receives the correct to property, you need to check the <a> tag renders an href:
```js
wrapper.find('a').element.href
```
We're considering adding a RouterLink stub to vue-test-utils, so you can do this:
```js
wrapper.find(RouterLink).props().to
```
But adding a stub basically means reimplimenting the RouterLink, since we don't actually need to stub it, we just need to have access to the component as a selector.
Instead, would we be able to expose RouterLink and RouterView from Vue Router?
  

### What does the proposed API look like?
We could import the components with the same name they are registered with:

```js
import { RouterLink, RouterView } from 'vue-router'
```

JounQin:
Why not use `Vue.component` to attach them?

```js
const RouterView = Vue.component('router-view');
const RouterLink = Vue.component('router-link');
```

eddyerburgh:
@JounQin I didn't know you could use Vue component like that.

The problem is you need Vue Router installed on the Vue instance, which adds `$route` and `$router` as read-only values. So instead, you'd have to do something like this:

```js
const localVue = createLocalVue()
localVue.use(VueRouter)
const RouterLink = localVue.component('router-link');

const wrapper = shallow(Comp, { localVue })
wrapper.find(RouterLink).props().to
```

Which I suppose isn't too bad 🤔

LinusBorg:
@eddyerburgh I "not too bad" good enough to close this? :-P

eddyerburgh:
I think we still need the change for two reasons:

1) If you install Vue Router you can't mock `$route` or `$router` in a test, so you wouldn't be able to test a RouterLink component and mock the param value of `$route`.
2) We'd need to make users aware that they need to follow this pattern (installing VueRouter then calling `Vue.component` to access to RouterLink component). If we expose RouterLink and RouterView, users will be able to follow the recommended pattern for finding stubbed components

posva:
Looks good to me and it's very easy to add. What do you think? @yyx990803

yyx990803:
I think that's good to have. Open to PR.

eddyerburgh:
This would be a breaking change for the common js or browser build.

We can export them in vue-test-utils instead

ArmorDarks:
[...]

PR #2853, opened 2019-07-17
Add named exports RouterLink, RouterView to es modules

graup:
#1976

### DIFF
--- a/build/configs.js
+++ b/build/configs.js
@@ -49,9 +49,12 @@ module.exports = [
 ].map(genConfig)
 
 function genConfig (opts) {
+  const input = opts.format === 'es'
+    ? resolve('src/index-es.js')
+    : resolve('src/index.js')
   const config = {
     input: {
-      input: resolve('src/index.js'),
+      input,
       plugins: [
         flow(),
         node(),
--- a/src/index-es.js
+++ b/src/index-es.js
@@ -1,0 +1,6 @@
+import VueRouter from './index.js'
+export default VueRouter
+
+/* Additional named exports for ES build only */
+export { default as RouterView } from './components/view'
+export { default as RouterLink } from './components/link'


posva:
Thank you for this but I think adding the exports only for the es module is not a good idea as people don't always know which version they are using so it would be confusing not to have the export because they aren't using the right version. At the same time, if we add it to every module, it's a breaking change for common js users.

So, as said in other comments, to get the components at runtime, you call

```js
const Link = Vue.component('RouterLink')
const View = Vue.component('RouterView')
```

That's how libs like Vuetify and bootstrap-vue do as well. The router components **do not work if the plugin is not installed** anyway, so importing them in an environment where it's not sure they are installed would produce errors about `$route` and `$router` not being accessible whereas with the solution above, you can emit a custom warning for your users:

```js
const Link = Vue.component('RouterLink')
if (process.env.NODE_ENV === 'development && !Link) {
  console.warn('You should install Vue Router [link]')
}
```

graup:
@posva thanks for taking the time to respond. I agree this solution is kind of an ugly workaround.

However, in my case the `const Link = Vue.component('RouterLink')` solution doesn't work, because I'm using several (typescript) packages. My components have explicit `Vue` imports and thus the global `Vue` class may be different, so it cannot resolve components registered in another context. Do you maybe have another idea how to solve this? It is the `<router-link>` that bites me but my real problem may be out of scope of what vue-router can do. In that case I'll gladly take it to StackOverflow.

graup:
@posva alternatively, would you be open to a PR doing what was outlined in [this comment](https://github.com/vuejs/vue-router/issues/1976#issuecomment-374020304)?

graup:
Looks like my issue is really caused by this: https://github.com/vuejs/vue/issues/8278

A change in vue-router wouldn't solve this anyway, so you can consider this case closed.

posva:
Yeah, the problem about having multiple Vue imports is out of scope of vue router

> alternatively, would you be open to a PR doing what was outlined in this comment?

No, because the comment doesn't give any substantial argument. It's not required to import the router in every component test. You can easily stub the components without giving any implementation or even use global options to stub them everywhere/install globally with vue-test utils

Issue #1769, opened 2017-09-27
router.push & router.replace to return promise if callbacks are not provided


charles-mathieus-jomedia:
### What problem does this feature solve?
It would help preventing callback hell... Would be more consistent with vue js api... 

### What does the proposed API look like?
router.push('/new/url').then(() => doSomeCoolStuff())

belmarca:
This seems to work fine if for example your component is expecting an event.

```
methods: {
  asyncRouterPush (route) {
    return new Promise(resolve => {
      resolve(this.$router.push({ path: route }))
    })
  },
  doSomething (myVar) {
    this.asyncRouterPush(myVar)
      .then(() => {
        this.$root.$emit('someEvent')
      })
  }
}
```

adamreisnz:
Any progress on this? Seems a bit backwards in 2019 that we still have to use a callback API, and can't simply `await router.push()`.

adamreisnz:
Here's a simple helper function in the mean time...

```js
/**
 * Async push method
 */
router.pushAsync = function(route) {
  return new Promise((resolve, reject) => {
    router.push(route, resolve, reject);
  });
};
```

Mighty683:
I will try implement this issue.

Mighty683:
The problem is `UnhandledPromiseRejectionWarning` if we dont implement `catch` for previous implementations.

`router.push` should throw error on route abort? Or should resolve with `undefined` on route abort?

https://github.com/vuejs/vue/blob/dev/src/core/util/next-tick.js should I implement this similar to `nextTick`?

posva:
Yes, it's similar to nextTick. It must work when Promises are not supported. The behavior only changes when no callback is passed and there is support for Promises

PR #2862, opened 2019-07-30
feat: return a promise with push and replace

Mighty683:
Closing(#1769)

Changing behaviour of `push` `replace` to return promise when no callback is provided.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -153,8 +153,8 @@ export default class VueRouter {
   push (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     // $flow-disable-line
     if (!onComplete && !onAbort && typeof Promise !== 'undefined') {
-      return new Promise(resolve => {
-        this.history.push(location, resolve)
+      return new Promise((resolve, reject) => {
+        this.history.push(location, resolve, reject)
       })
     } else {
       this.history.push(location, onComplete, onAbort)
@@ -164,8 +164,8 @@ export default class VueRouter {
   replace (location: RawLocation, onComplete?: Function, onAbort?: Function) {
     // $flow-disable-line
     if (!onComplete && !onAbort && typeof Promise !== 'undefined') {
-      return new Promise(resolve => {
-        this.history.replace(location, resolve)
+      return new Promise((resolve, reject) => {
+        this.history.replace(location, resolve, reject)
       })
     } else {
       this.history.replace(location, onComplete, onAbort)
--- a/src/index.js
+++ b/src/index.js
@@ -151,11 +151,25 @@ export default class VueRouter {
   }
 
   push (location: RawLocation, onComplete?: Function, onAbort?: Function) {
-    this.history.push(location, onComplete, onAbort)
+    // $flow-disable-line
+    if (!onComplete && !onAbort && typeof Promise !== 'undefined') {
+      return new Promise(resolve => {
+        this.history.push(location, resolve)
+      })
+    } else {
+      this.history.push(location, onComplete, onAbort)
+    }
   }
 
   replace (location: RawLocation, onComplete?: Function, onAbort?: Function) {
-    this.history.replace(location, onComplete, onAbort)
+    // $flow-disable-line
+    if (!onComplete && !onAbort && typeof Promise !== 'undefined') {
+      return new Promise(resolve => {
+        this.history.replace(location, resolve)
+      })
+    } else {
+      this.history.replace(location, onComplete, onAbort)
+    }
   }
 
   go (n: number) {
--- a/test/unit/specs/api.spec.js
+++ b/test/unit/specs/api.spec.js
@@ -152,7 +152,7 @@ describe('router.push/replace', () => {
     })
   })
   describe('callbacks', () => {
-    it('push with callback return undefined', done => {
+    it('push does not return a Promise when a callback is passed', done => {
       expect(router.push('/foo', done)).toEqual(undefined)
     })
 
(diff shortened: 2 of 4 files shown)


types/router.d.ts
```diff
@@ -23,8 +23,8 @@ export declare class VueRouter {
   beforeEach (guard: NavigationGuard): Function;
   beforeResolve (guard: NavigationGuard): Function;
   afterEach (hook: (to: Route, from: Route) => any): Function;
-  push (location: RawLocation, onComplete?: Function, onAbort?: ErrorHandler): void;
-  replace (location: RawLocation, onComplete?: Function, onAbort?: ErrorHandler): void;
+  push (location: RawLocation, onComplete?: Function, onAbort?: ErrorHandler): Promise<Route> | void;
```
posva: we should be able to keep both functions with overloading

```ts
push (location: RawLocation, onComplete?: Function, onAbort?: ErrorHandler): void
push (location: RawLocation): Promise<Route>
```

src/index.js
```diff
@@ -151,11 +151,25 @@ export default class VueRouter {
   }
 
   push (location: RawLocation, onComplete?: Function, onAbort?: Function) {
-    this.history.push(location, onComplete, onAbort)
+    // $flow-disable-line
+    if (!onComplete && !onAbort && typeof Promise !== 'undefined') {
+      return new Promise(resolve => {
+        this.history.push(location, resolve)
```
posva: we must handle the reject as well

src/index.js
```diff
@@ -151,11 +151,25 @@ export default class VueRouter {
...
+      this.history.push(location, onComplete, onAbort)
+    }
   }
 
   replace (location: RawLocation, onComplete?: Function, onAbort?: Function) {
-    this.history.replace(location, onComplete, onAbort)
+    // $flow-disable-line
+    if (!onComplete && !onAbort && typeof Promise !== 'undefined') {
+      return new Promise(resolve => {
+        this.history.replace(location, resolve)
```
posva: same, here we must handle the reject

test/unit/specs/api.spec.js
```diff
@@ -151,38 +151,87 @@ describe('router.push/replace callbacks', () => {
...
 
-  it('replace abort', done => {
-    router.replace('/foo', spy1, spy2)
-    router.replace('/bar', () => {
-      expect(calls).toEqual([1, 1, 2, 2])
-      expect(spy1).not.toHaveBeenCalled()
-      expect(spy2).toHaveBeenCalled()
-      done()
+  describe('promises', () => {
+    it('push with callback return undefined', done => {
```
posva: This test should be moved to the group of callbacks

```suggestion
    it('does not return a Promise when a callback is passed', done => {
```

test/unit/specs/api.spec.js
```diff
@@ -151,38 +151,87 @@ describe('router.push/replace callbacks', () => {
...
+        .then(spy1)
+        .finally(() => {
+          expect(calls).toEqual([1, 2, 3, 4])
+          expect(spy1).toHaveBeenCalledWith(router.currentRoute)
+          done()
+        })
+    })
+
+    it('push abort', done => {
+      router.push('/foo')
```
posva: this test should be updated with a catch

test/unit/specs/api.spec.js
```diff
@@ -151,38 +151,87 @@ describe('router.push/replace callbacks', () => {
...
+      router.push('/bar').finally(() => {
+        expect(calls).toEqual([1, 1, 2, 2])
+        expect(spy1).not.toHaveBeenCalled()
+        done()
+      })
+    })
+
+    it('replace with callback return undefined', done => {
```
posva: this should be moved to the other group

test/unit/specs/api.spec.js
```diff
@@ -151,38 +151,87 @@ describe('router.push/replace callbacks', () => {
...
+        .then(spy1)
+        .finally(() => {
+          expect(calls).toEqual([1, 2, 3, 4])
+          expect(spy1).toHaveBeenCalledWith(router.currentRoute)
+          done()
+        })
+    })
+
+    it('replace abort', done => {
+      router.replace('/foo')
```
posva: this test should also be updated with a `catch`

posva (changes requested):

src/index.js
Mighty683: Adding rejection will cause `UnhandledPromiseRejectionWarning` rise for cases when no `catch` is provided. Will this be a problem with backward compability?

types/router.d.ts
Mighty683: Done

test/unit/specs/api.spec.js
Mighty683: Done

src/index.js
posva: no, it's not a problem. The error appears only if there is a rejection and no `catch` is provided, so we are good

Mighty683: Updated with promise rejection. Also added error handling tests.

test/unit/specs/api.spec.js
Mighty683: Done

test/unit/specs/api.spec.js
Mighty683: Added

test/unit/specs/api.spec.js
Mighty683: Added

src/index.js
Mighty683: Done

posva:
I fixed the problems with e2e tests, you can rebase against dev

posva (approved):
Thanks!

PR #2867, opened 2019-08-02
fix(ssr): Memory leak in poll method (fix #2606).

### LINKED ISSUE #2606: SSR memory leak in poll method
### Version
3.0.2

### Reproduction link
[https://jsfiddle.net/ronald_d_rogers/5mxdpkLj/2/](https://jsfiddle.net/ronald_d_rogers/5mxdpkLj/2/)

### Steps to reproduce
Make `router-view` appearance conditional based on some variable (or simply have it not appear):
```html
<div id="app">
  <div v-if="condition">
    <router-view />
  </div>
</div>
```

Ensure that the component that matches the route has a `beforeRouteEnter` guard, where a function is passed to the `next(...)` method (e.g. `next(vm => {}`):
```es6
export default {
  beforeRouteEnter(to, from, next) {
    // Poll happens only if you pass in a function to next(...)
    next(vm => {})
  }
}
```

Visit that route.
```sh
curl http://localhost:8080/route
```

A poll will happen infinitely until `router-view` is created.

In a typical SSR application where an app is created per request, it will never be created. 
The end result is a memory leak with an event loop filled with poll's every 16ms.

In the reproduction JS Fiddle, a leak occurs every time you press the `SSR Request` button.
If you set `dontLeak` to true, the leak stops.

The leak can be verified in the JS Fiddle by clicking on the `SSR Request` button a few times (say 20 times), opening Chrome Developer Tools, going to the Performance tab, and in the recording looking at the contents of "Frame".

capture

Here is an example repo as well:
https://github.com/ronald-d-rogers/vue-router-ssr-memory-leak

### What is expected?
A new poll does not recurse infinitely for every SSR request.

A new poll recurses infinitely for every SSR request.

ronald-d-rogers:
Fix #2606.

This fix depends on the following being merged first to work: https://github.com/vuejs/vue/pull/9479

Fixes memory leak in `poll` method when a Vue app is created per request, which is typical in SSR.

In particular, the leak happens when:

1. The `router-view` is conditioned to appear upon a variable that never becomes true.
```html
<div id="app">
  <div v-if="condition">
    <router-view />
  </div>
</div>
```

2. A `beforeRouteEnter` guard is defined in the view in the following way:
```es6
export default {
  beforeRouteEnter(to, from, next) {
    // Poll happens only if you pass in a function to next(...)
    next(vm => {})
  }
}
```

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -177,8 +177,14 @@ export class History {
     }
 
     runQueue(queue, iterator, () => {
+      const app = this.router.app
       const postEnterCbs = []
-      const isValid = () => this.current === route
+      const isValid = () => {
+        if (app && app._isBeingDestroyed) {
+          return false
+        }
+        return this.current === route
+      }
       // wait until async components are resolved before
       // extracting in-component enter guards
       const enterGuards = extractEnterGuards(activated, postEnterCbs, isValid)
@@ -189,8 +195,8 @@ export class History {
         }
         this.pending = null
         onComplete(route)
-        if (this.router.app) {
-          this.router.app.$nextTick(() => {
+        if (app) {
+          app.$nextTick(() => {
             postEnterCbs.forEach(cb => {
               cb()
             })


ronald-d-rogers:
Note I haven't had time to test this again after all of the changes that have been made to both projects. Will do this tomorrow if possible.

ronald-d-rogers:
There still appears to be a leak. I've gone ahead and updated https://github.com/ronald-d-rogers/vue-router-ssr-memory-leak with the latest versions of `vue` and `vue-router`, and included instructions on testing the fix and an explanation how it works. Hope this helps.

posva:
Thanks for the detailed report @ronald-d-rogers. I will try to give it a look next week

ronald-d-rogers:
@posva Note that I don't think this a great fix. I think the solution is probably to find a way to get rid of the `poll`, which was just added to fix a bug where the `router-view` hasn't yet appeared because the previous route is transitioning out, because we don't have a `vm` to pass to the `beforeRouteEnter` hook's `next` callback if the hook was defined with one (`next(vm => { ... })`), at the time the callbacks are called.

One way to get around it may be by putting the callback on the matching `RouteRecord`, adding some method like `registerComponentInstance(key, vm)` and `registerBeforeRouterEnterCb(key, cb)` to it, and only firing when both have been registered `if (this.hasCb(key) && this.hasInstance(key)) { /* fire */ }`.

But I can't tell if that would work here or not. Maybe I will give it a go today or tomorrow.

ryouaki:
This issues will block on poll function. So you should let poll exit at right time。

posva:
Let's close this in favor of https://github.com/vuejs/vue-router/pull/2875

PR #2877, opened 2019-08-08
parameters should be placed before hash(fix #2876)

### LINKED ISSUE #2876: the url pattern is not conform to the URI generic syntax in the newest version 3.1.0
### Version
3.1.0

### Reproduction link
[https://github.com/yangshuan/vue-router-test-3.1.1](https://github.com/yangshuan/vue-router-test-3.1.1)

### Steps to reproduce
Steps are very simple:
1. Use the vue-cli to generate a new empty project;
2. Start the project;
3. Open a browser (e.g. Chrome), and open the application index page;
4. Everything seems normal, but if you use url with parameters, e.g. http://localhost:8080/?a=1, the page will goto http://localhost:8080/#/?a=1, the uri is not right, it should be 'http://localhost:8080/?a=1#/'

### What is expected?
The URI should always be conform to the URI generic syntax 'scheme:[//authority]path[?query][#fragment]', otherwise most of the lib or code which get the parameters from the URI will fail.

this url is not conform to the URI generic syntax which is 'scheme:[//authority]path[?query][#fragment]';
The problem the 3.1.0 version brings out is that, when we want to get the parameters from the URL by the default URI generic syntax, we fail all the time, since the uri is not right.

yangshuan:
fix #2876

The URI should always be conform to the URI generic syntax 
scheme:[//authority]path[?query][#fragment]
Otherwise most of the lib or code which get the parameters from the URI will fail.

### DIFF
--- a/src/history/hash.js
+++ b/src/history/hash.js
@@ -142,7 +142,7 @@ function getUrl (path) {
   const query = searchPos > -1 ? base.slice(searchPos) : ''
   base = query ? base.slice(0, searchPos) : base
 
-  return `${base}#${path + query}`
+  return `${base}${query}#${path}`
 }
 
 function pushHash (path) {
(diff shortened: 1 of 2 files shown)


posva:
Thank you for this but we can also revert the commit that introduced this change and remove a bit more code. I will do that based on what we discuss at https://github.com/vuejs/vue-router/issues/2876

PR #2888, opened 2019-08-17
docs(guide): add note to props function for named views

rebz:
Added a note to the end of the Functions Mode. A couple of us were stuck on using prop functions for Named Views, it's simple yet easily overlooked. Thought this might help other devs who run into the same issue.


posva:
Thank you but there is already a note about named props. I think we could improve it by adding a tip pointing out the code. That way it should appear in search

BTW there is https://github.com/vuejs/vue-router/pull/2499

PR #2905, opened 2019-08-23
feat: add a default component for nested routes

### LINKED ISSUE #2105: Allow nested routes without parent component
Allows the routes to be defined with a prefix that doesn't need to be repeated in each route.

Currently this doesn't work because the parent route doesn't have a component specified. This feature isn't about nested routes but just about nested paths.

```js
{ path: '/prefix': children: [
  { path: 'one', component: ... },
  { path: 'two', component: ... },
]}
```
Creates `/prefix/one` and `/prefix/two`.

deckar01:
Fixes #2105

Allows nested child routes to render inside routes without a component. Previously this prevented the child route's component from rendering without any warning. Now the parent's component will default to a `PassThrough` component that provides the `<router-view>` the child needs to render in.

### DIFF
--- a/examples/nested-routes/app.js
+++ b/examples/nested-routes/app.js
@@ -80,7 +80,14 @@ const router = new VueRouter({
 
         { path: 'quy/:quyId', component: Quy },
 
-        { name: 'zap', path: 'zap/:zapId?', component: Zap }
+        { name: 'zap', path: 'zap/:zapId?', component: Zap },
+
+        {
+          path: 'wrap',
+          children: [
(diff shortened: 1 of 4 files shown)


deckar01:
This needs docs and more test coverage, but I thought I should get some feedback before investing any more time into this.

posva:
I appreciate the initiative but this is not something we want to add, at least yet. That's also why the issue was closed.

A note about the implementation you made: the component should be functional and there should be a render function instead of a template

PR #2926, opened 2019-09-13
fix: publish should include all files under src

Astray-git:
Currently the npm package only includes 4 js files under src, all folders are missing.


posva (approved):
Thanks!

PR #2930, opened 2019-09-17
fix(router-view): fix deeply nested keep-alive router-views displaying

### LINKED ISSUE #2923: keep-alive deeply nested router-views display the last visited component component
### Version
3.1.3

### Reproduction link
[https://jsfiddle.net/lxcuso4/cu6keqw2/28](https://jsfiddle.net/lxcuso4/cu6keqw2/28)

### Steps to reproduce
Click on childA
Click on index
Click on childB

### What is expected?
we want to show B page

but actually A page dispaly

---
This behavior happens when we use keep-alive and have three layers of nested routing

zrh122:
Fix #2923.

### DIFF
--- a/examples/keepalive-view/app.js
+++ b/examples/keepalive-view/app.js
@@ -27,6 +27,8 @@ const IndexChild2 = { template: '<div>index child2</div>' }
 
 const Home = { template: '<div>home</div>' }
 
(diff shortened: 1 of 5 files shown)


posva:
Thanks for this @zrh122 !

PR #2954, opened 2019-10-03
Adds note on subfolders to history mode samples

### LINKED ISSUE #2863: Improve history mode server configuration documentation to include non-root publicPath
The documentation for [HTML5 History Mode - Example Server Configurations](https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations) assumes all apps are running at the document root (ie `/`). Apps using a non-root [`publicPath`](https://cli.vuejs.org/config/#publicpath) are going to struggle with these config examples.

Update the examples to include `<publicPath>` placeholders, eg

nginx
```
location <publicPath> {
  try_files $uri $uri/ <publicPath>/index.html;
}
```

AVGP:
fixes #2863

### DIFF
--- a/docs/guide/essentials/history-mode.md
+++ b/docs/guide/essentials/history-mode.md
@@ -19,7 +19,7 @@ Not to worry: To fix the issue, all you need to do is add a simple catch-all fal
 
 ## Example Server Configurations
 
-**Note**: The following examples assume you are serving your app from the root folder. If you deploy to a subfolder, you should use [the `publicPath` option of Vue CLI](https://cli.vuejs.org/config/#publicpath) and the related [`base` property of the router](https://router.vuejs.org/api/#base). You also need to adjust the examples below to use the subfolder instead of the root folder (e.g. replacing `RewriteBase /` with `RewriteBase /name-of-your-subfolder/`)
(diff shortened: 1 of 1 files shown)


docs/guide/essentials/history-mode.md
```diff
@@ -19,6 +19,8 @@ Not to worry: To fix the issue, all you need to do is add a simple catch-all fal
 
 ## Example Server Configurations
 
+**Note**: The following examples assume you are serving your app from the root folder. If you deploy to a subfolder, see [the `publicPath` option of Vue CLI](https://cli.vuejs.org/config/#publicpath) and the related [`base` property of the router](https://router.vuejs.org/api/#base).
```
posva: I think we should give a bit more information like: you should use the publicPath and the related... and adapt the rewrite rules to point to `/base` instead of `/`

AVGP: Fair point, I was wondering about the right level of verbosity. Does the new wording work?

posva (approved):
Thanks a lot! Before merging I want to know for sure that the change of the Apache rule is accurate. @phanan maybe you can confirm?

phanan:
@posva I have not much experience with vue-router deployed in a subdirectory, but the Apache config looks legit :) The additional text misses a dot though.

posva:
Thanks for the quick feedback!

PR #2955, opened 2019-10-04
Not only string can be passed as a param in Location

yooouuri:
When I do the following:

```
this.$router.push({ name: 'somepage', params: { someParam: false } });
```

The following error occurs:

image

When I console.log `this.$route.params` in my somepage route, the `someParam` variable with `false` as value exists.

So update the type to allow more than a string.


posva:
Thank you but for the moment, only strings are valid because numbers should be casted to strings. Before changing the TS types we would need to introduce casting for params and query values for numbers to strings. Booleans are not accepted because the casting might be unexpected

Issue #2957, opened 2019-10-07
Allow reloading of the current route when clicked


boycce:
### What problem does this feature solve?
I think it's important for `vue-router` to try mimic native browser for such as when dealing with links being clicked.  Adding an option to allow the current route being reloaded when clicked would allow any API requests to be re-fetched that located under the navigational guards (beforeEach, beforeRouteEnter, etc)

### What does the proposed API look like?
Maybe include a router instance option to allow for this behaviour

posva:
Duplicate of https://github.com/vuejs/vue-router/issues/974

boycce:
Yes and also a duplicate of #2430 
Many developers are still left with inconclusive and convoluted answers when trying to provide their users with a native browser behaviour while using navigational guards. Imagine the frustration if browsers removed this feature.

posva:
If you know it was a duplicate, please don't open a new issue...

boycce:
Apologies, I didn't realise that #974 was still open, I just saw that it was opened 3 years ago and assumed it was closed.

Issue #2430, opened 2018-10-13
Is there any way to force navigation when clicking on a router-link even if the route doesn't change?


trandaison:
### What problem does this feature solve?
Reload current page's state or refresh the page.

For example, I have a `<HeaderBar>` which contains a `Home` button, just like the Facebook one.
If I'm in another page (rather than Home), when I hit this button, it will navigate to Home. But when I'm in Home page, hit this button it does nothing.

It's expected to reload the page, and scroll to the top, right?

Please don't tell me to write out an action to refresh the state and scroll to top. This header bar is a shared component, and the Home button can be hit at any where.

### What does the proposed API look like?
Some thing like 
```vue
<router-link :to="{ name: 'Home', force: true }">Home</router-link>
```
or
```js
this.$router.replace({ name: 'Home', force: true });
```

posva:
You can reload the page, but you should handle the _refresh_ in a click handler if you want to have that custom behaviour

Duplicate of https://github.com/vuejs/vue-router/issues/974

trandaison:
@posva 
If so, Navigation Guards (`beforeRouteLeave`, `beforeRouteUpdate`) will not be triggered.

How do you deal with that?

posva:
yeah, they are not supposed to be triggered as you are in the same route with no param change or whatsoever... You need to refactor the code in methods and call them in both places. Cheers

trandaison:
**[Solved]**
This is my solution, if any body needs.

I create a base component, and use it instead of `<router-link>`.
If you hit the link when you're in that page already, it will create a query name `_`, so the `fullPath` will be change and the page will be "reload".

You can change the query name to whatever you want, and use the timestamp or some unique string to make the `fullPath` changes.

```vue
<template lang="html">
  <router-link v-bind="$attrs" :to="to" @click.native="clickHandler">
    <slot></slot>
  </router-link>
</template>

<script>
export default {
  name: 'BaseLink',

  props: {
    to: {
      type: [Object, String],
      default: '/',
    },
  },

  data() {
    return {
      currentRoute: null,
    };
  },

  created() {
    this.currentRoute = this.$route.name;
  },

  methods: {
    clickHandler($event) {
      if (!($event.ctrlKey || $event.metaKey) && this.currentRoute === this.to.name) {
        const query = { ...this.$route.query, _: new Date().getTime() };
        const newRoute = { ...this.$route, query };
        this.$router.replace(newRoute);
      }
    },
  },
};
</script>
```

[...]

PR #2960, opened 2019-10-09
feat(history): allow same route navigation

boycce:
To get the component to build a fresh instance you need to add a key attribute i.e `<router-view :key="$route.id"></router-view>`. Regardless the vue-router hooks still fire when reloading.

Issue: #974 
Related: #2957 #2430

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -7,7 +7,7 @@ import VueRouter from 'vue-router'
 Vue.use(VueRouter)
 
 // 2. Define route components
-const Home = { template: '<div>home</div>' }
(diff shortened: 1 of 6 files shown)


pedzed:
While this is not yet implemented, how could you force a refresh when the `query` changes in `<router-link>`?

n123456789k:
@pedzed you can do a watch on the route

pedzed:
@narwy Can you give an example? Or refer me to some reading material?

Thanks.

posva:
Thank you but this doesn't implement #974 

Removing the duplicated navigation would be a breaking change, so the change is v4 only (per https://github.com/vuejs/rfcs/pull/150)

PR #2961, opened 2019-10-10
DOCS: An example with the server written in golang is added (essentials/history-mode)

AndrusGerman:
#### Reason for the proposal:
Because different servers already support **golang**. I think it's a good idea to add golang as an example.

#### why gorilla/mux:
Since it complies with the net / http standards of golang. It has no conflicts with the standard library
> The code has been verified and works correctly


AndrusGerman:
This implementation is not complete I will improve it and send it again

PR #2962, opened 2019-10-10
docs: example is added with a server written in Go

AndrusGerman:
#### Reason for the proposal:
Because different servers already support **golang**. I think it's a good idea to add golang as an example.

#### why gorilla/mux:
Since it complies with the net / http standards of golang. It has no conflicts with the standard library
> The code (**Golang**) has been verified and works correctly.

#### What exactly did you modify?
These changes only include modifications to the **'history-mode.md**' files. Adding the example of a server in **Go**

#### Did you implement the example in all languages?
Yes

### DIFF
--- a/docs-gitbook/de/essentials/history-mode.md
+++ b/docs-gitbook/de/essentials/history-mode.md
@@ -55,12 +55,14 @@ import (
 )
 var httpPort = "80"
 var indexFile = "index.html"
+var folder = "./"
+
 func serverHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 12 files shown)


anirudhmurali:
This is exactly what I wanted! I tried using this: https://github.com/ddo/go-vue-handler, but your solution works perfectly with gorilla/mux. Thanks a lot for this, wish this gets merged soon.

AndrusGerman:
Checking my pull request list, I see that this has not been confirmed for more than 3 years, I'm willing to update it, but surely the documentation will change with respect to the original pull request.
Everything is fine?

PR #3000, opened 2019-10-17
feat: add destroy method, remove event listener on destroy

bailnl:
I have a problem in the process of implementing the micro-frontend. When switching between multiple applications, because the Router instance was not destroyed, the event listener was not removed. 

I consulted the API documentation and did not find the destroy API。

so~ I implemented One, in the destroy method will remove the existing event listener


posva:
Would https://github.com/vuejs/vue-router/pull/3172 fix your problems?

posva:
Closing since #3172 was merged and was supposed to address the same problems

Issue #3008, opened 2019-10-24
Better error message when the scrollBehavior selector is invalid


theprojectsomething:
### What problem does this feature solve?
Currently if you pass a hash id with unescaped CSS special characters `/[!"#$%&'()*+,\-./:;<=>?@[\\]^{|}~]/` to a scrollBehavior selector, execution will fail: 
```
'XXX' is not a valid selector
```
Because of this url-styled hashes (e.g. `"#one/two"` or `"#main/secondary"`), both valid ids in HTML5, cannot be passed directly from the `to.hash` or `from.hash` parameters.
This is because behind the scenes `document.querySelector`, which relies on CSS character compliance, is used in all but a very small number of circumstances (i.e. `/^#\d/`). This feature proposes broadening the use case for `document.getElementById` to all selectors that have the appearance of an id, specifically ungrouped selectors starting with a hash (i.e. `/^#[^, ]+$/`)

### What does the proposed API look like?
API doesn't change, except the following would work:
```
scrollBehavior(to, from, savedPosition) {
  return {
    selector: to.hash,
  }
}
// where:
window.location = '/one/two#three/four';
```

posva:
After further research, I'm realizing that trying to be smart and detect valid id selectors is a bad idea because there are cases where it won't be possible. Adding support for hashes starting with a number was a mistake and has been removed in v4. So instead I think we should do the following

- Deprecate `selector` in favour of `el` (`Element | string` like `new Vue({ el })` and `app.mount(el)`)
- Show a deprecation notice when the selector matches `/^#\d` pointing the user to use the `el` option instead
- Show a warning if `el` is a string, starts with a `#` and fails with `querySelector` pointing out to use `document.getElementById(<theid>)` or to escape the selector and point to a resource.
- A version will be updated at https://github.com/vuejs/vue-router-next/blob/master/src/scrollBehavior.ts#L69-L99

theprojectsomething:
Hi @posva thanks for coming back. Couple of notes ...

It's been a while, but I think the gist/context of the original PR was that `scrollToPosition` was being used internally by the router to replicate/implement scrolling to the location/href hash on route change. I'm not sure if that's still the case (or maybe I'm wrong - was it ever the case??).

Given that the hash is meant to map to an `id` (or technically also a `[name]`) within the document, I would have expected this concept to be a first class citizen within the logic. Using `querySelector` breaks this, unless there is intervention by the developer (which will be required for everything but hardcoded anchors).

[...]

PR #3009, opened 2019-10-24
Allow scrollBehavior selectors to match ids with unescaped CSS special characters (feature request #3008)

theprojectsomething:
As per the title and the original feature request #3008 this PR means a wider range of selectors are passed to `document.getElementById`, resulting in more straightforward use of `scrollBehavior` with e.g. `to.hash`

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -121,14 +121,16 @@ function isNumber (v: any): boolean {
   return typeof v === 'number'
 }
 
-const hashStartsWithNumberRE = /^#\d/
+const selectorLooksLikeIdRE = /^#[^, ]+$/
 
 function scrollToPosition (shouldScroll, position) {
   const isObject = typeof shouldScroll === 'object'
   if (isObject && typeof shouldScroll.selector === 'string') {
+    // getElementById is used only where a selector looks like a single [id] ... this allows for use of CSS
+    // 'special characters' that would otherwise fail in querySelector without escaping, e.g. "#one/two"
     // getElementById would still fail if the selector contains a more complicated query like #main[data-attr]
     // but at the same time, it doesn't make much sense to select an element with an id and an extra selector
-    const el = hashStartsWithNumberRE.test(shouldScroll.selector) // $flow-disable-line
+    const el = selectorLooksLikeIdRE.test(shouldScroll.selector) // $flow-disable-line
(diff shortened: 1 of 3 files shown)


src/util/scroll.js
```diff
@@ -121,14 +121,16 @@ function isNumber (v: any): boolean {
   return typeof v === 'number'
 }
 
-const hashStartsWithNumberRE = /^#\d/
+const selectorLooksLikeIdRE = /^#[^, ]+$/
```
posva, line 124: why any character but `,` and ` `?

theprojectsomething: @posva good question - logic is that spaces aren't allowed in an HTML5 id and both a space or a comma would suggest a complex selector, e.g:
```
<div id="one">  <!-- getElementById('one') -->
  <div class="two"> <!-- querySelector('#one .two') --> </div>
  <div id="one/two"><!-- getElementById('one/two') --></div></div>
<div id="two"><!-- querySelector('#two, #one') --> </div>
```
If special characters or ids starting with a number are used in a complex selector (i.e. one that starts with a `#` but contains a comma or space) they would need to be properly escaped as if they were being used in CSS / querySelector. This is currently the case and continues to apply in the PR

theprojectsomething: This also means the onus isn't on the router to attempt to escape characters that aren't intended to be escaped, instead it simply checks for a simple id query and passes everything else to querySelector with all the caveats. The idea being that you can pass the hash fragment from a url or `to.hash` without parsing it. The only gotcha being that it can't contain a `,` (which is technically allowed in an id)

theprojectsomething: @posva a `,` or ` ` suggests a complex selector (either nested or an OR). Technically an id selector could contain an `,` but the above provides the best balance of simplicity vs. coverage without breaking changes. Consider:
```
// Simple selector (passed to getElementById):
#one
#1
#one/two

// Complex selector (passed to querySelector): 
#one .two
#one, .two
#one\,two // selects id="one,two"
```
Did you have any suggestions?

theprojectsomething: Hi @posva any update on this PR?

posva:
Thanks for this but I think we need to go in a different direction: https://github.com/vuejs/vue-router/issues/3008#issuecomment-634094481

Issue #2925, opened 2019-09-09
Missing param due to case-sensitive token regex


pimlie:
### Version
3.1.3

### Reproduction link
[https://jsfiddle.net/j4s8ka1v/](https://jsfiddle.net/j4s8ka1v/)

### Steps to reproduce
Unfortunately the above reproduction link does **not** reproduce the issue, not sure why the _tokensToFunction_ method isnt used in the jsfiddle. But I've traced the issue to `tokensToFunction` from `path-to-regexp`. The issue seems to have been fixed in v3.1.0 due to [this pr](https://github.com/pillarjs/path-to-regexp/pull/191). You can check eg [`dist/vue-router.esm.js`](https://github.com/vuejs/vue-router/blob/dev/dist/vue-router.esm.js#L619) to see the RegExp does not set the `i` modifier

That makes this report more or less a fyi / request to upgrade `path-to-regexp` 😃 

Locally the error is caused by a route similar to
```
  routes: [{
      path: "/:locale(en|pt-br)?/:source(.*)?",
      component: myComponent,
    }],
```

which fails when you browse to e.g. `/pt-BR/path`:

![image](https://user-images.githubusercontent.com/1067403/64534777-5af11100-d316-11e9-9225-f34616eee0fe.png)

### What is expected?
That the token / param is matched case-insensitively

### What is actually happening?
It doesnt, causing a missing param error

posva:
Thanks, unfortunately upgrading it comes with some breaking changes, one of them is changing `'*'` routes to `'(.*)'` which is widely used. The upgrade will happen in next major version though.
So this is something people will have to workaround by providing different possible versions (case sensitive and case insensitive) until next major version.

pimlie:
Thanks, could you please explain what you think the workaround is?

Unfortunately I dont fully understand why this issue should be closed, this issue exists and directly contradicts with the docs which says that all routes [are matched case insensitively by default](https://router.vuejs.org/api/#routes). Wouldnt it be better for visibility to keep this open with a target label like `breaking-change` and/or `target: v4`?

boringame:
This problem also bothers me, which is why I submitted the PR.
Here's what I did while waiting for the merge: generate expressions with upper and lower case letters.

```typescript
function escapeIgnoreCase(
    text: string
): string {
    const holderFlag = "((holderFlag))";
    const holderFlagEscape = RegExp.escape(holderFlag);
[...]

PR #3032, opened 2019-11-18
chore: upgrade path-to-regexp to v1.8.0 (fix #2924)

### LINKED ISSUE #2924: Keep path-to-regexp as a dependency for non-browser builds (or expose its exports)
For vue-router v2.0.0 `path-to-regexp` was moved from a dependency to a [dev dependency](https://github.com/vuejs/vue-router/commit/76d83b6eb4af507763c5ee6f61a52ed9a7804572) (thanks @pi0 for finding the commit).

Unfortunately the commit doesnt list a reason why this was (probably mostly security/political?), but I think it could make sense to keep `path-to-regexp` as an external dependency for non-browser builds.

One reason is that Nuxt.js also uses e.g. the [`tokensToFunction`](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/utils.js#L452) method of `path-to-regexp`. This means that currently a Nuxt.js app will include two copies of that function, one from vue-router and one from Nuxt.js itself which increases the bundle size needlessly.

Another solution for the above could be that `vue-router` would expose the path-to-regexp methods like `tokensToFunction` method so Nuxt.js can use those, but using dependencies has more benefits ofc like quicker delivery of patches for `path-to-regexp`.

Revert: 76d83b6eb4af507763c5ee6f61a52ed9a7804572

pimlie:
See issue https://github.com/vuejs/vue-router/issues/2925 and the commit messages.

The server host configuration commit could be omitted but I'm running Node on a remote docker cluster which means I need to set the host to `0.0.0.0`. To be able to run the e2e tests I actually also needed to add `'no-sandbox', 'headless', 'disable-gpu', 'disable-dev-shm-usage'` to `chromeOptions.args` in `test/e2e/nightwatch.config.js`. Otherwise Chrome wont be able to start in a docker container.

### DIFF
--- a/examples/server.js
+++ b/examples/server.js
@@ -27,7 +27,8 @@ fs.readdirSync(__dirname).forEach(file => {
 
 app.use(express.static(__dirname))
 
+const host = process.env.HOST || 'localhost'
 const port = process.env.PORT || 8080
-module.exports = app.listen(port, () => {
-  console.log(`Server listening on http://localhost:${port}, Ctrl+C to stop`)
+module.exports = app.listen(port, host, () => {
+  console.log(`Server listening on http://${host}:${port}, Ctrl+C to stop`)
 })
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "lint-staged": "^8.2.0",
     "nightwatch": "^1.1.13",
     "nightwatch-helpers": "^1.0.0",
-    "path-to-regexp": "^1.7.0",
+    "path-to-regexp": "^1.8.0",
     "rollup": "^1.20.1",
     "rollup-plugin-buble": "^0.19.8",
     "rollup-plugin-commonjs": "^10.0.2",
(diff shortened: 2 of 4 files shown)


posva (changes requested):
Thanks! Could you move the test to a new one and also add a test that checks that case sensitive when the option is specified and that  params do not match?

pimlie:
> that checks that case sensitive when the option is specified

Hmm, I dont seem to have access to _RouteConfig.caseSensitive_ in _normalizeLocation_ as that only receives a _Route_ type. Would it be ok to add the normalized [`pathToRegexpOptions`](https://github.com/vuejs/vue-router/blob/dev/src/create-route-map.js#L80) to the _RouteRecord_ type?

posva:
Ah right, it will need a test using the Router instance instance instead let's cover it that way for the moment

pimlie:
Unfortunately I am not sure how I can use the Router instance to add a test for 2925. Could you elaborate a bit on what I should do?

posva:
I forgot about this one! I locally added a new test. Thank you!

PR #3045, opened 2019-11-27
fix(hash): Properly handle encoded parameters in hash (close #2725)

### LINKED ISSUE #2725: Properly encode/decode the percentage character (% -> %25) across browsers and upon initial navigation

```js
[
  {
    path: '*',
    component: Other,
  },
]
```

## Query / TODO

---

`index.html`
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>Document</title>
  </head>

  <body>

    <div id="app">
      <pre>{{ $route.path }}</pre>
      <router-view></router-view>
    </div>

    <script>
      var Other = {
      }
      var router = new VueRouter({
        mode: 'history',
        routes: [
          {
            path: '*',
            component: Other,
          },
        ],
      })

      new Vue({
        router: router,
        el: '#app',
      })
    </script>
  </body>
</html>
```

leameow:
Fix an issue where page browsing to a route and then reloading leads to inconsistent parameters and fullPath values.

### DIFF
--- a/src/history/hash.js
+++ b/src/history/hash.js
@@ -107,27 +107,15 @@ function ensureSlash (): boolean {
 }
 
 export function getHash (): string {
-  // We can't use window.location.hash here because it's not
-  // consistent across browsers - Firefox will pre-decode it!
+  // We need to get the raw hash to keep consistent behavior with routes
+  // `fullPath` using encoded path. The params are decoded later in the
+  // matcher.
   let href = window.location.href
   const index = href.indexOf('#')
   // empty path
   if (index < 0) return ''
 
   href = href.slice(index + 1)
-  // decode the hash but not the search or hash
-  // as search(query) is already decoded
-  // https://github.com/vuejs/vue-router/issues/2708
-  const searchIndex = href.indexOf('?')
-  if (searchIndex < 0) {
-    const hashIndex = href.indexOf('#')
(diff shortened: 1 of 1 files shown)


posva:
hello, sorry for the delay. Do you have a reproduction for this one?

leameow:
It relates to https://github.com/vuejs/vue-router/issues/2725, specifically, when I click a link to `#/%252520` my URL get transformed to `#/%20` and when I reload the page I get `#/` which is not correct and consistent. This fix might not be the proper fix, but we have been using this patch and didn't have issues so far.

posva:
@isra17 can you provide an actual repro (you can use this: https://jsfiddle.net/posva/9r6xhqbp/). Also what browsers are you seeing this at? Only hash mode?

leameow:
Can't repro injsfiddle due to URL not being visiable/modifiable. Both chrome and firefox at least. I didn't test on other mode than hash mode.

You can close the issue, I don't have my environnement with this bug anymore and I don't have the bandwith right now to dig further on this.

Cheers!

posva:
I revisited the problem and there are a few other issues. The actual problem of double decoding affects both history and hash mode. Thanks for the case of `/%252520`, it's a good one to check double encoding!

Issue #3027, opened 2019-11-10
Add onComplete & onAbort to router-link


XAMPPRocky:
### What problem does this feature solve?
In 2.2.0 Vue added `onComplete` and `onAbort` callbacks to `router.push`, however there seems to be no equivalvent when using `<router-link>`. I've seen suggestions that can you use `beforeRouteEnter` as an alteranative in some cases, however I have a case where I have a `<header>` containing `<router-link>`s that remains on the page after a user clicks on the router link, I want to be able to change a property of the header after the user has clicked the link. So `beforeRouteEnter` would not work for my case. Ideally I'd like to be able to supply `onComplete` and `onAbort` to the `router-link` directly.

### What does the proposed API look like?
Personally I have no preference over how this should look, I'm not experienced enough with vue router to say what is the most idiomatic, however I would expect to be something like the following.
```html
<router-link to="…" on-complete="callback" on-abort="callback"></router-link>
<!-- or -->
<router-link to="…" complete="callback" abort="callback"></router-link>
```

kelunik:
Implemented by #3114.

posva:
I completely forgot about this issue. This kind of thing should go through the thinking process of an RFC in our RFC repo at vuejs/rfcs. One of the things I see is using events instead of callbacks, which is more adapted to Vue. Another thing would be giving more specific examples about what is being done onAbort/onComplete that is specific to that router-link. To me this is the kind of thing that would go in an `afterEach` hook because it looks global.

Right now this can be achieved with an event listener that calls `$router.push` and building the href with `$router.resolve` (which is a workaround).

kelunik:
@posva: Could you please clarify how the workaround would look like? Currently the best solution without this PR seems to be forking the `RouterLink` for me.

bponomarenko:
We have one use case, which might be connected to this issue, but correct me if I'm wrong.
What I'm trying to achieve in an application, is to simply display loading indicator on the app level while navigation is in progress. In most cases this can be covered by doing:

```js
const state = Vue.observable({ navigating: false });

router.beforeEach((to, from, next) {
  state.navigating = true;
  next();
});

router.afterEach(() => {
  state.navigating = false;
});
```

By registering these guards before any other guards, I can make sure that loading indicator will be shown until all other global, in-component and router guards are resolved, async components are loaded, etc.
[...]

PR #3047, opened 2019-11-30
Create router errors

lmichelin:
This PR improves the way router errors are thrown.

4 router errors types have been created:

- `NavigationDuplicated`
- `NavigationCancelled`
- `NavigationRedirected`
- `NavigationAborted`

Before:

```
Uncaught (in promise) undefined
```

After:

```
Uncaught (in promise) Error: NavigationAborted: Navigating to location ("/foo?abort=y") has been aborted
    at eval (base.js:157)
    at eval (app.js:33)
    at iterator (base.js:153)
    at step (async.js:13)
    at step (async.js:17)
    at runQueue (async.js:21)
    at HashHistory.confirmTransition (base.js:179)
    at HashHistory.transitionTo (base.js:74)
```

### DIFF
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -2,8 +2,8 @@
 
 import type Router from '../index'
 import { History } from './base'
-import { NavigationDuplicated } from './errors'
-import { isExtendedError } from '../util/warn'
+import { isRouterError } from '../util/warn'
+import { NavigationFailureType } from './errors'
 
 export class AbstractHistory extends History {
   index: number
@@ -51,7 +51,7 @@ export class AbstractHistory extends History {
         this.updateRoute(route)
       },
       err => {
-        if (isExtendedError(NavigationDuplicated, err)) {
+        if (isRouterError(err, NavigationFailureType.NAVIGATION_DUPLICATED)) {
           this.index = targetIndex
         }
       }
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -4,14 +4,20 @@ import { _Vue } from '../install'
 import type Router from '../index'
 import { inBrowser } from '../util/dom'
 import { runQueue } from '../util/async'
-import { warn, isError, isExtendedError } from '../util/warn'
+import { warn, isError, isRouterError } from '../util/warn'
 import { START, isSameRoute } from '../util/route'
 import {
   flatten,
   flatMapComponents,
   resolveAsyncComponents
 } from '../util/resolve-components'
-import { NavigationDuplicated } from './errors'
+import {
+  createNavigationDuplicatedError,
+  createNavigationCancelledError,
+  createNavigationRedirectedError,
+  createNavigationAbortedError,
+  NavigationFailureType
+} from './errors'
 
 export class History {
   router: Router
@@ -104,7 +110,7 @@ export class History {
       // When the user navigates through history through back/forward buttons
       // we do not want to throw the error. We only throw it if directly calling
       // push/replace. That's why it's not included in isError
-      if (!isExtendedError(NavigationDuplicated, err) && isError(err)) {
+      if (!isRouterError(err, NavigationFailureType.NAVIGATION_DUPLICATED) && isError(err)) {
         if (this.errorCbs.length) {
           this.errorCbs.forEach(cb => {
             cb(err)
@@ -122,7 +128,7 @@ export class History {
       route.matched.length === current.matched.length
     ) {
       this.ensureURL()
-      return abort(new NavigationDuplicated(route))
+      return abort(createNavigationDuplicatedError(current, route))
     }
 
     const { updated, deactivated, activated } = resolveQueue(
@@ -146,12 +152,15 @@ export class History {
     this.pending = route
     const iterator = (hook: NavigationGuard, next) => {
       if (this.pending !== route) {
-        return abort()
+        return abort(createNavigationCancelledError(current, route))
       }
       try {
         hook(route, current, (to: any) => {
-          if (to === false || isError(to)) {
+          if (to === false) {
             // next(false) -> abort navigation, ensure current URL
+            this.ensureURL(true)
+            abort(createNavigationAbortedError(current, route))
+          } else if (isError(to)) {
             this.ensureURL(true)
             abort(to)
           } else if (
@@ -160,7 +169,7 @@ export class History {
               (typeof to.path === 'string' || typeof to.name === 'string'))
           ) {
             // next('/') or next({ path: '/' }) -> redirect
-            abort()
+            abort(createNavigationRedirectedError(current, route))
             if (typeof to === 'object' && to.replace) {
               this.replace(to)
             } else {
--- a/src/history/errors.js
+++ b/src/history/errors.js
@@ -1,22 +1,67 @@
-export class NavigationDuplicated extends Error {
-  constructor (normalizedLocation) {
-    super()
-    this.name = this._name = 'NavigationDuplicated'
-    // passing the message to super() doesn't seem to work in the transpiled version
-    this.message = `Navigating to current location ("${
-      normalizedLocation.fullPath
-    }") is not allowed`
-    // add a stack property so services like Sentry can correctly display it
-    Object.defineProperty(this, 'stack', {
-      value: new Error().stack,
-      writable: true,
-      configurable: true
-    })
-    // we could also have used
-    // Error.captureStackTrace(this, this.constructor)
-    // but it only exists on node and chrome
-  }
+export const NavigationFailureType = {
+  redirected: 1,
+  aborted: 2,
+  cancelled: 3,
+  duplicated: 4
 }
 
-// support IE9
-NavigationDuplicated._name = 'NavigationDuplicated'
+export function createNavigationRedirectedError (from, to) {
+  return createRouterError(
+    from,
+    to,
+    NavigationFailureType.redirected,
+    `Redirected from "${from.fullPath}" to "${stringifyRoute(to)}" via a navigation guard.`
+  )
+}
+
+export function createNavigationDuplicatedError (from, to) {
+  return createRouterError(
+    from,
+    to,
+    NavigationFailureType.duplicated,
+    `Avoided redundant navigation to current location: "${from.fullPath}".`
+  )
+}
+
+export function createNavigationCancelledError (from, to) {
+  return createRouterError(
+    from,
+    to,
+    NavigationFailureType.cancelled,
+    `Navigation cancelled from "${from.fullPath}" to "${to.fullPath}" with a new navigation.`
+  )
+}
+
+export function createNavigationAbortedError (from, to) {
+  return createRouterError(
+    from,
+    to,
+    NavigationFailureType.aborted,
+    `Navigation aborted from "${from.fullPath}" to "${to.fullPath}" via a navigation guard.`
+  )
+}
+
+function createRouterError (from, to, type, message) {
+  const error = new Error(message)
+  error._isRouter = true
+  error.from = from
+  error.to = to
+  error.type = type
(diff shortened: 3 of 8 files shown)


lmichelin:
@posva I wrote some unit tests, the PR is ready to be reviewed 😉

ariabuckles:
Bump on this? This would be really helpful for tracking down where promise rejections are coming from <3

bponomarenko:
It seems this PR can also solve #3027. +1 for it to be merged and released.

quentinus95:
@lmichelin could it be possible to export the exceptions so they can be imported (e.g. using `import { NavigationDuplicated } from "vue-router"`)? It would allow catching the exceptions and use `instanceof` (especially when coding with TypeScript)  to take decisions based on the thrown exceptions.

posva:
@quentinus95 `instanceof` with extended Errors doesn't work on some browsers like IE.
This PR is on hold until https://github.com/vuejs/vue-router-next/pull/123 is merged to make sure exported errors share the same naming and reduce migration cost

posva (changes requested):
Hi @lmichelin 

Now that errors are stable on vue-router-next, I think we could try to align the user API part to make migration easier.
Namely, we need:

- a `type`, `from` property
- an object (to simulate the enumeration) with the same properties as [`NavigationFailureType`](https://github.com/vuejs/vue-router-next/blob/master/src/errors.ts#L26-L30)
- an exported TS interface so users can type errors

We can also remove:

- the `name` and `_name` properties
- all the code to hack around Error extension and use plain errors (https://github.com/vuejs/vue-router-next/blob/master/src/errors.ts#L77-L90)

It's okay to have a few differences like the lack of `from` and `to` properties. Also, let me know if you don't have the time so I can pick it up. If you only manage to add a few things but not all, that's, of course, fine as well!

Link to RFC: https://github.com/vuejs/rfcs/pull/150
Link to vue-router-next errors file: https://github.com/vuejs/vue-router-next/blob/master/src/errors.ts

package.json
```diff
@@ -62,7 +62,7 @@
     "babel-preset-flow-vue": "^1.0.0",
     "browserstack-local": "^1.4.0",
     "buble": "^0.19.8",
-    "chromedriver": "^79.0.0",
+    "chromedriver": "^81.0.0",
```
posva: Let's keep this change out of the PR. If needed I will upgrade this on dev and you can then merge or rebase against it

src/history/abstract.js
```diff
@@ -51,7 +50,7 @@ export class AbstractHistory extends History {
         this.updateRoute(route)
       },
       err => {
-        if (isExtendedError(NavigationDuplicated, err)) {
+        if (isRouterError(err)) {
```
posva: We probably want to keep the check of duplicated navigation

src/history/base.js
```diff
@@ -104,7 +109,7 @@ export class History {
       // When the user navigates through history through back/forward buttons
       // we do not want to throw the error. We only throw it if directly calling
       // push/replace. That's why it's not included in isError
-      if (!isExtendedError(NavigationDuplicated, err) && isError(err)) {
+      if (!isRouterError(err) && isError(err)) {
```
posva: same as the comment above, I think we should keep the check of the error type

src/history/errors.js
```diff
@@ -1,22 +1,45 @@
...
+  newStack.splice(1, 2) // remove 2 last useless calls
+  error.stack = newStack.join('\n')
+  return error
+}
+
+export const ErrorTypes = {
+  NAVIGATION_GUARD_REDIRECT: 'NavigationRedirected',
+  NAVIGATION_ABORTED: 'NavigationAborted',
+  NAVIGATION_CANCELLED: 'NavigationCancelled',
+  NAVIGATION_DUPLICATED: 'NavigationDuplicated'
```
posva: let's keep the same names as the public API: https://github.com/vuejs/vue-router-next/blob/master/src/errors.ts#L26-L30 (`NavigationFailureType`)

src/history/errors.js
```diff
@@ -1,22 +1,45 @@
...
-// support IE9
-NavigationDuplicated._name = 'NavigationDuplicated'
+export function createNavigationRedirectedError (from, to) {
+  return createRouterError(from, to, ErrorTypes.NAVIGATION_GUARD_REDIRECT, `Navigating to location ("${
+    to.fullPath
+  }") has been cancelled by a redirection`)
+}
+
+export function createNavigationDuplicatedError (from, to) {
+  return createRouterError(from, to, ErrorTypes.NAVIGATION_DUPLICATED, `Navigating to location ("${
```
posva: Let's take the same error messages as in https://github.com/vuejs/vue-router-next/blob/master/src/errors.ts#L47-L72

src/history/errors.js
```diff
@@ -1,22 +1,45 @@
...
+    to.fullPath
+  }") has already been initiated`)
+}
+
+export function createNavigationCancelledError (from, to) {
+  console.log(to)
```
posva: this is probably a leftover

posva (changes requested):

package.json
lmichelin: I did this change to be able to run e2e tests. Without this update, e2e tests can't run: https://circleci.com/gh/vuejs/vue-router/5121 😉

posva: oh wow, they are still green, that's so wrong lol
I updated on `dev` the chromedriver but there are a few tests failing I need to fix. After that you can merge dev into this (don't worry about the git commit history, we squash at the end)

posva: It's good now!

src/util/warn.js
```diff
@@ -16,10 +16,9 @@ export function isError (err: any): boolean {
...
-  return (
-    err instanceof constructor ||
-    // _name is to support IE9 too
-    (err && (err.name === constructor.name || err._name === constructor._name))
-  )
+export function isRouterError (err: any, errorType: ?string): boolean {
+  const _isRouterError = isError(err) && err._isRouter
+
+  if (!errorType) return _isRouterError
+  return _isRouterError && err.type === errorType
```
posva: I think we can make this simpler:

```suggestion
  return isError(err) && err._isRouter && (errorType == null || err.type === errorType)
```

src/history/base.js
```diff
@@ -104,7 +110,7 @@ export class History {
       // When the user navigates through history through back/forward buttons
       // we do not want to throw the error. We only throw it if directly calling
       // push/replace. That's why it's not included in isError
-      if (!isExtendedError(NavigationDuplicated, err) && isError(err)) {
+      if (!isRouterError(err, NavigationFailureType.NAVIGATION_DUPLICATED) && isError(err)) {
```
posva, line 113: We can remove the `isError` check since `isRouterError` already checks that
```suggestion
      if (!isRouterError(err, NavigationFailureType.NAVIGATION_DUPLICATED)) {
```

src/history/errors.js
```diff
@@ -1,22 +1,67 @@
...
-    })
-    // we could also have used
-    // Error.captureStackTrace(this, this.constructor)
-    // but it only exists on node and chrome
-  }
+export const NavigationFailureType = {
+  NAVIGATION_GUARD_REDIRECT: 'NavigationRedirected',
+  NAVIGATION_ABORTED: 'NavigationAborted',
+  NAVIGATION_CANCELLED: 'NavigationCancelled',
+  NAVIGATION_DUPLICATED: 'NavigationDuplicated'
```
posva: The values do not matter much since they shouldn't be used directly but I think it's better to force that behavior by using numbers (like an enum), to prevent the mistake of people using a simple string comparison: `error.type === 'NavigationRedirected'` instead of `error.type === NavigationFailureType.redirected`

We should also keep the same keys as in https://github.com/vuejs/vue-router-next/blob/master/src/errors.ts#L27-L29

```suggestion
  redirected: 1,
  aborted: 2,
  cancelled: 3,
  duplicated: 4
```

posva (changes requested):
I realised you have a sample for an e2e test but it doesn't seem like you need it since you added a unit test already

lmichelin:
> I realised you have a sample for an e2e test but it doesn't seem like you need it since you added a unit test already

`examples/router-errors/app.js` is an example to show how to throw the errors, it's not made for e2e purposes, but just to show simple examples for these errors.

src/history/base.js
lmichelin: I don't think it's equivalent: If I understood correctly we want to check if `err` is an error but not a navigation duplicated error. When `err` is `null` (like here https://github.com/vuejs/vue-router/blob/588220c475b29e038bc6c0badddac7992bf27445/src/history/base.js#L188) the result of the condition will not be the same.

posva: You are right,  misread the line

posva (approved):
Thanks a lot for your help with this @lmichelin and sorry for the delay!

src/history/base.js
schmidtk, line 113: This should be renamed to `NavigationFailureType.duplicated`.

src/history/abstract.js
```diff
@@ -51,7 +51,7 @@ export class AbstractHistory extends History {
         this.updateRoute(route)
       },
       err => {
-        if (isExtendedError(NavigationDuplicated, err)) {
+        if (isRouterError(err, NavigationFailureType.NAVIGATION_DUPLICATED)) {
```
schmidtk, line 54: Here as well.

schmidtk (commented):
Good morning @lmichelin and @posva. I discovered this PR while handling navigation duplicated errors in my application and have been tracking the changes in case it requires migration when we update `vue-router` in the future.

I noted two places in the changes where `NavigationFailureType.NAVIGATION_DUPLICATED` references were not updated after the keys were changed in `NavigationFailureType`. I'm admittedly new to `vue-router` internals, so apologies if I misunderstood something here. Thanks for your time.

lmichelin:
You are right, I completely forgot these references 🙈

lmichelin:
I just opened a new PR to fix this: https://github.com/vuejs/vue-router/pull/3204

alecgibson:
I think this was a breaking change? We don't get the `name` property on the `Error` any more, which is useful when trying to handle particular errors (eg ignoring `NavigationDuplicated`). At the very least it may be nice to expose `isRouterError` as a utility method? And also to expose the failure types, especially because it's an opaque enum.

posva:
The errors were not documented and considered private. Checking navigation failures docs are on the way!

alecgibson:
I'd say that checking an `Error.name` field is pretty [standard practice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name), especially if it's not been prefixed with an underscore. I don't mind a change, but I'm just saying this could be considered breaking.

snoozbuster:
+1 for this being a breaking change - once again I am back here to figure out why suddenly there are errors in my app where there weren't errors before (and the fact that it was merged and published without docs on how to do proper error checking also seems like a mistake)

snoozbuster:
@posva - are these type constants exported from the package, so that I can inspect navigation errors in my application and handle them appropriately? I don't see anything which adds that, is it in another PR?

afv:
I was checking `err.name === 'NavigationDuplicated'` before but after these changes the name is just `'Error'`.

As a workaround I'm now importing `NavigationFailureType` at `main.js` and adding it to `window`:

```
import { NavigationFailureType } from 'vue-router/src/history/errors';

window.NavigationFailureType = NavigationFailureType;
```

On components I can now check `err.type === window.NavigationFailureType.duplicated`.

posva:
Please check https://github.com/vuejs/vue-router/pull/3220 for the proposed documentation for errors. Your feedback could be valuable there

PR #3056, opened 2019-12-07
fix(router-link): Navigation works when the tag value is rendered as …

holazz:
Navigation works when the tag value is rendered as a custom component


posva:
If you want a custom component to apply the `click` listener, rebind it inside the component or use the v-slot api

PR #3060, opened 2019-12-09
Remove broken gitbook links

Opinion:
The folders "en" and "zh-cn" don't exist within "docs-gitbook" so I removed them from LANGS.md

If they have some different use, please let me know :)

Cheers! :beer:


posva:
Thank you, this folder is kept for historical reasons but it's not used anymore

PR #3073, opened 2019-12-21
feat(core): add aria-current to active links (close #2116)

### LINKED ISSUE #2116: Extend <router-link> to set an aria-current attribute for active links, instead of an active class
This feature enables developers to mark up the active router link in an accessible manner for users of assistive technologies.

Similar to the `active-class` property on `<router-link>` introduce an `aria-current` string property, defaulting to a `null` value. When set, use the `aria-current` DOM attribute in addition to the active class.

Criteria for whether the DOM attribute should be set are the same  as for `active-class`. The attribute should be set on the DOM node itself -- or instead on the special nested `<a>` element that router-link can upgrade, should one exist.

The DOM attribute's value should be taken from the `aria-current` router-link property. Optionally, this could be validated against the list of valid options as per the ARIA 1.1 specification and console warnings could be provided on attempts to set invalid values.

smhigley:
Hi! First contributor + pretty new to Vue 👋 😄 

This PR adds `aria-current="page"` to active links, and resolves #2116. I made the following two choices:

1. Use `aria-current="page"` over other token values
The original issue mentions allowing custom token values, or future `<router-link>` extension possibilities. That still seems useful, but `aria-current="page"` seems like the right choice at least 90% of the time. The other tokens are either much less common (e.g. `step`) or unlikely to be used with a router at all (e.g. `date`)

2. Use the `activeClass` logic over only applying it to the exact active route
This is more hand-wavy, but seems like the better choice -- URL params shouldn't necessarily affect `aria-current`, and it's not the worst thing to have parent routes also receive `aria-current` if `exact` is not applied.

I'm hoping this is a small enough change that it can go in before the harder work of figuring out how to extend `<router-link>` is done, and it seems like a sensible default even with customization. I've run into some issues using vue-router b/c of this, so I'd personally love to see `aria-current` added :). I'm happy to answer any questions on the accessibility side, and let me know if I missed anything PR-wise. Thanks!

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -8,6 +8,7 @@ import { warn } from '../util/warn'
 // work around weird flow bug
 const toTypes: Array<Function> = [String, Object]
 const eventTypes: Array<Function> = [String, Array]
+let ariaCurrentTypes: 'page' | 'step' | 'location' | 'date' | 'time'
 
 const noop = () => {}
 
@@ -27,7 +28,10 @@ export default {
     replace: Boolean,
     activeClass: String,
     exactActiveClass: String,
-    ariaCurrentValue: 'page' | 'step' | 'location' | 'date' | 'time',
+    ariaCurrentValue: {
+      type: ariaCurrentTypes,
+      default: 'page'
+    },
     event: {
       type: eventTypes,
       default: 'click'
@@ -68,8 +72,7 @@ export default {
       ? classes[exactActiveClass]
       : isIncludedRoute(current, compareTarget)
 
-    const ariaCurrentType = this.ariaCurrentValue || 'page'
-    const ariaCurrentValue = classes[exactActiveClass] ? ariaCurrentType : null
+    const ariaCurrentValue = classes[exactActiveClass] ? this.ariaCurrentValue : null
 
     const handler = e => {
       if (guardEvent(e)) {
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -67,6 +67,8 @@ export default {
       ? classes[exactActiveClass]
       : isIncludedRoute(current, compareTarget)
 
+    const ariaCurrentValue = classes[activeClass] ? 'page' : null
+
     const handler = e => {
       if (guardEvent(e)) {
         if (this.replace) {
@@ -117,7 +119,7 @@ export default {
 
     if (this.tag === 'a') {
       data.on = on
-      data.attrs = { href }
+      data.attrs = { href, 'aria-current': ariaCurrentValue }
     } else {
       // find the first <a> child and apply listener and href
       const a = findAnchor(this.$slots.default)
@@ -145,6 +147,7 @@ export default {
 
         const aAttrs = (a.data.attrs = extend({}, a.data.attrs))
         aAttrs.href = href
+        aAttrs['aria-current'] = ariaCurrentValue
       } else {
         // doesn't have <a> child, apply listener to self
         data.on = on
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -8,7 +8,6 @@ import { warn } from '../util/warn'
 // work around weird flow bug
 const toTypes: Array<Function> = [String, Object]
 const eventTypes: Array<Function> = [String, Array]
-let ariaCurrentTypes: 'page' | 'step' | 'location' | 'date' | 'time'
 
 const noop = () => {}
 
@@ -29,7 +28,7 @@ export default {
     activeClass: String,
     exactActiveClass: String,
     ariaCurrentValue: {
-      type: ariaCurrentTypes,
+      type: String,
       default: 'page'
     },
     event: {
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -67,7 +67,7 @@ export default {
       ? classes[exactActiveClass]
       : isIncludedRoute(current, compareTarget)
 
-    const ariaCurrentValue = classes[activeClass] ? 'page' : null
+    const ariaCurrentValue = classes[exactActiveClass] ? 'page' : null
 
     const handler = e => {
       if (guardEvent(e)) {
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -27,6 +27,7 @@ export default {
     replace: Boolean,
     activeClass: String,
     exactActiveClass: String,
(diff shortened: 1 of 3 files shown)


phiter:
Oh I didn't see that this already existed.

Let's hope this gets merged soon.

posva:
hey @smhigley thanks for the well detailed PR!

I think this makes sense and is safe to add as a default for vue router but I was wondering about point 2: applying it to active instead of exact-active. I think the latter makes more sense as a default because it is more likely to end up in one single element marked as `aria-current="page"` which is what is recommended on the spec (https://www.w3.org/TR/wai-aria-1.1/#aria-current). Do you know if this may not be true anymore or if there is something else to consider that isn't present on the spec?

It's not directly related but I also wanted to note that active and exact behavior is very likely to change on v4 (https://github.com/vuejs/rfcs/pull/136) and I was going to use exact active to set it. If you have any input for that, please go ahead 🙂 !

phiter:
@posva it should be in exact active.

smhigley:
@posva I updated the PR to add `aria-current` to only exact active links :).

I originally did non-exact active because the tutorials and examples I happened to see seemed to be adding active styles based on the active class, and I was trying to choose whichever would add most parity between visual active styles and `aria-current`. I don't have a very good feel of which is more helpful in practice, though, since I'm fairly new to Vue. I'm happy to defer to the one you think fits better :)

test/e2e/specs/active-links.js
```diff
@@ -70,12 +70,14 @@ module.exports = {
         browser.assert
           .cssClassPresent(`li:nth-child(${i}) a`, 'router-link-exact-active')
           .assert.cssClassPresent(`li:nth-child(${i}) a`, 'router-link-active')
+          .assert.attributeEquals(`li:nth-child(${i}) a`, 'aria-current', 'page')
```
posva, line 74: let's add another test after line 33 with the assert for text that checks on the link `/active-links/users` that the aria attribute is not present

src/components/link.js
```diff
@@ -67,6 +67,8 @@ export default {
       ? classes[exactActiveClass]
       : isIncludedRoute(current, compareTarget)
 
+    const ariaCurrentValue = classes[exactActiveClass] ? 'page' : null
```
posva: I think it makes sense to allow customising the value of `page`. Can we add an optional prop `ariaCurrentValue` that defaults to `page`

posva (changes requested):

src/components/link.js
phiter: @posva that could make sense but the other options for `aria-current` aren't really related to links.

posva: `step` (and a few others) could, so allowing to customise it makes sense

phiter: yeah I suppose. If you have a stepper component that uses vue-router and changes the url, that would make sense.

smhigley: is `ariaCurrentValue` fine as the prop name? Happy to add as long as it's uncontroversial 😄

posva: Seems fine to me. It's similar to `true-value` and `false-value` and it's very explicit. So unless anybody has a better suggestion we should go with it

src/components/link.js
```diff
@@ -27,6 +27,7 @@ export default {
     replace: Boolean,
     activeClass: String,
     exactActiveClass: String,
+    ariaCurrentValue: 'page' | 'step' | 'location' | 'date' | 'time',
```
posva: Let's use the default in props and pass a correct type

```suggestion
    ariaCurrentValue: {
      type: String,
      default: 'page'
},
```

src/components/link.js
```diff
@@ -67,7 +68,8 @@ export default {
       ? classes[exactActiveClass]
       : isIncludedRoute(current, compareTarget)
 
-    const ariaCurrentValue = classes[exactActiveClass] ? 'page' : null
+    const ariaCurrentType = this.ariaCurrentValue || 'page'
+    const ariaCurrentValue = classes[exactActiveClass] ? ariaCurrentType : null
```
posva:
```suggestion
    const ariaCurrentValue = classes[exactActiveClass] ? this.ariaCurrentValue : null
```

test/e2e/specs/active-links.js
```diff
@@ -71,6 +71,9 @@ module.exports = {
           .cssClassPresent(`li:nth-child(${i}) a`, 'router-link-exact-active')
           .assert.cssClassPresent(`li:nth-child(${i}) a`, 'router-link-active')
           .assert.attributeEquals(`li:nth-child(${i}) a`, 'aria-current', 'page')
+        if (i < 19) {
+          browser.assert.not.attributeEquals(`li:nth-child(${i + 1}) a`, 'aria-current', 'page')
```
posva: Let's remove this test and instead add the one I said above

posva (changes requested):

test/e2e/specs/active-links.js
smhigley: I'm not 100% sure what you were suggesting above, but I tried adding a test at line 33 instead. I had thought "line 33" above was a typo since the comment was on line 73 -- if this change isn't what you were thinking and you did want a different test after line 73, I might need clarification :)

src/components/link.js
smhigley: Added the default, and it looks like the same hack you used for `toTypes` and `eventTypes` allows adding a literal type here (which seems preferred to me, admittedly coming from the Typescript side of things)

test/e2e/specs/active-links.js
posva: Ah I see, sorry it wasn't clear. What you added is exactly what I had in mind!

src/components/link.js
posva: Unfortunately, it won't work 😅 
The codebase is used a quite outdated version of flow. For typing props in TS, there is a `PropsType` that would allow doing what you want. Here we do need to provide the `String` constructor for it to work with Vue

smhigley: Ah, that's unfortunate :P.

In any case, updated to `String` now!

posva (approved):
Thank you!

smhigley:
Thank *you*! 😄

PR #3081, opened 2019-12-26
no warn if `pathMatch` is empty (fix #3072)

### LINKED ISSUE #3072: Named catch-all route warns if `pathMatch` is empty
### Version
3.1.3

### Reproduction link
[http://jsfiddle.net/yza5k02v/](http://jsfiddle.net/yza5k02v/)

### Steps to reproduce
- click on /nope
- click on button

### What is expected?
no warn

there is a warn

---
maybe the param should be marked as optional to prevent this warning from ever happening

yinm:
Try to fix https://github.com/vuejs/vue-router/issues/3072
if pathMatch is empty, skip warning.

### DIFF
--- a/src/util/params.js
+++ b/src/util/params.js
@@ -25,7 +25,8 @@ export function fillParams (
     return filler(params, { pretty: true })
   } catch (e) {
     if (process.env.NODE_ENV !== 'production') {
-      warn(false, `missing param for ${routeMsg}: ${e.message}`)
+      // Fix #3072 no warn if `pathMatch` is string
+      warn(typeof params.pathMatch === 'string', `missing param for ${routeMsg}: ${e.message}`)
     }
     return ''
   } finally {
--- a/test/unit/specs/create-matcher.spec.js
+++ b/test/unit/specs/create-matcher.spec.js
@@ -71,6 +71,15 @@ describe('Creating Matcher', function () {
     expect(path).toEqual('/error/some')
   })
 
+  it('matches named catch-all route with empty pathMath param without warning', function () {
+    process.env.NODE_ENV = 'development'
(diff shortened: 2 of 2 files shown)


src/util/params.js
```diff
@@ -24,7 +24,10 @@ export function fillParams (
 
     return filler(params, { pretty: true })
   } catch (e) {
-    if (process.env.NODE_ENV !== 'production') {
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      params.pathMatch !== '' // Fix #3072 no warn if `pathMatch` is empty
```
posva: let's check if typeof is not string on the warn method (replacing the `false` value)

posva (changes requested):

src/util/params.js
yinm: fixed at https://github.com/vuejs/vue-router/pull/3081/commits/590adbbf1e78662e58fc26ee656ea600261db1c4.
I also simplify source code at https://github.com/vuejs/vue-router/pull/3081/commits/8a213d9f49f0f22b858672879b5077130a3a3db4.

src/util/params.js
```diff
@@ -24,9 +24,8 @@ export function fillParams (
 
     return filler(params, { pretty: true })
   } catch (e) {
-    if (process.env.NODE_ENV !== 'production') {
```
posva, line 27: the if was necessary though

posva (changes requested):

src/util/params.js
yinm: oops, I revert it

posva (approved):
Thanks!

PR #3112, opened 2020-01-31
:pencil: docs(router): Update history mode apache config, fix SPA

anselal:
Make Single Page Applications load the correct endpoint on refresh and not just fallback to index.html


anselal:
What caused the test to fail?

posva:
Thank you but this is not related to the server configs of that page in documentation. It's already specified in other documentation pages

PR #3117, opened 2020-02-06
fix(TS): update type of Route.name

rijkvanzanten:
On very first load, `name` is `null`, not `string` or `undefined`. This means that either the passed route is wrong, or the type is incorrect.

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -134,7 +134,7 @@ export interface Location {
 
 export interface Route {
   path: string
-  name?: string
+  name?: string | null
   hash: string
   query: Dictionary<string | (string | null)[]>
   params: Dictionary<string>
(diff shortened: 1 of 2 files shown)


posva (approved):
Yeah, only the initial location has a name of `null`

posva (changes requested):
I forgot, can you fix the failing test?

rijkvanzanten:
Jumped into it too hastily before.. it's fixed now @posva 🙂

posva:
Thanks!

zhuwenqi001:
after route allows name to be null, there will be such a problem in my project: 
`router.push` and `next` method error. Do other interfaces also need to be modified? Such as Routeconfig and Location

LinusBorg:
Appearantly so. Please open a new issue about this.

posva:
@zhuwenqi001 see https://github.com/vuejs/vue-router/issues/3139
In your case you can do `tab.name!` because you know that the name exists. I'm thinking that allowing `null` for `Location` shouldn't introduce any problem, so we could maybe allow it for convenience

PR #3125, opened 2020-02-15
fix(router-link): active class with encoded params

### LINKED ISSUE #3103: Router link active class not applied with non ascii params
### Version
3.1.3

### Reproduction link
[https://codesandbox.io/s/romantic-browser-xvj0m](https://codesandbox.io/s/romantic-browser-xvj0m)

### Steps to reproduce

* Refresh page at https://xvj0m.csb.app/user/t%C3%A9st : the active class is not applied.

### What is expected?
Active class on the router link

No active class is applied

---

Finally, the active class is applied when you click on router links, even with non ASCII char. The only problem is at the first load of the page.

oleksiikhr:
I will try to fix this :). Closes https://github.com/vuejs/vue-router/issues/3103

The problem is url generation. We have 2 routes:

```
<router-link to="/special/tést1">/special/tést1</router-link>
<router-link :to="{ name: 'special', params: {word: 'tést2'} }">/special/tést2</router-link>
```

In the first case, the route is transmitted as it is. In the second - it is decoded through the library: `path-to-regexp` (in new versions this can be changed by parameter)

Before:

After:

But, if it works in **History Mode**. Then for **Hash Mode** you need to edit separately, because there the same problem exists.

[Somewhere here](https://github.com/vuejs/vue-router/blob/dev/src/history/hash.js#L128)


posva (changes requested):
Thank you for the PR but this the fix shouldn't change how URLs are generated. The encoded version should be the one generated. The problem is the other way around: when the page is reloaded, the `fullPath` property is decoded (/user/tést) instead of being  encoded (/user/t%C3%A9st) like it usually is when doing `$router.push({ name: 'special', params: { test: 'tést' }})`

Regarding the difference between named routes and string-based push like `$router.push('/url')` and `$router.push({ path: '/url' })`: when doing string-based push, the user is responsible for encoding the urls but when doing others, the router should encode params for you. That's why the two router-link you added in the e2e test yield different results. But it's expected

You can change things in this pr if you want to or close it and start a new one

oleksiikhr:
Thanks for the detailed answer, if not against, I will try to solve it again.

Now encode fullPath strings are compared, and the user now encode his path (changed in tests)

oleksiikhr:
If I understood correctly, then I, as a user, should encode the string in the fallen test.

`<li><router-link to="/é">/é</router-link></li>`
to
`<li><router-link to="/%C3%A9">/é</router-link></li>`

posva:
@Alexeykhr I updated the original issue at https://github.com/vuejs/vue-router/issues/3103

Going to an unencoded URL is not valid, so it doesn't matter if the active link matches or not, but when visiting the encoded valid url, it should match the link

PR #3129, opened 2020-02-21
Documentation: Update guard example to avoid stack overflow

NoelDeMartin:
I am surprised that this hasn't been brought up before, so maybe I am doing something wrong. But I am getting a "too much recursion" error with the sample code in the documentation.

Looking at the code, it seems to me like getting a stack overflow *is* the expected behaviour. May this be a [schrödinbug](http://www.catb.org/jargon/html/S/schroedinbug.html)?

### DIFF
--- a/docs/guide/advanced/navigation-guards.md
+++ b/docs/guide/advanced/navigation-guards.md
@@ -39,7 +39,7 @@ Every guard function receives three arguments:
 ```js
 // BAD
 router.beforeEach((to, from, next) => {
-  if (to.path !== '/login' && !isAuthenticated) next('/login')
+  if (to.name !== 'Login' && !isAuthenticated) next({ name: 'Login' })
   // if the user is not authenticated, `next` is called twice
   next()
 })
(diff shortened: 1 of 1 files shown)


NoelDeMartin:
It doesn't look like the failed tests have anything to do with my change.

docs/guide/advanced/navigation-guards.md
```diff
@@ -39,7 +39,7 @@ Every guard function receives three arguments:
 ```js
 // BAD
 router.beforeEach((to, from, next) => {
-  if (!isAuthenticated) next('/login')
+  if (to.path !== '/login' && !isAuthenticated) next('/login')
```
posva:
```suggestion
  if (to.name !== 'Login' && !isAuthenticated) next({ name: 'Login' })
```
let's use a named route instead for both examples

posva (changes requested):

docs/guide/advanced/navigation-guards.md
NoelDeMartin: If we're doing that, would it be better to use it in the call to next as well for consistency?

`next({ name: 'Login' });`

I'm actually using the name in my code, but I didn't want to mix concepts and the path should be fine because it doesn't contain query params and such.

posva: Yes, that makes sense!

posva:
Thanks!

PR #3135, opened 2020-02-29
fix(types): interface Route, the query property

777genius:
Maybe it can also be null (see the `query` property in the `Location`)


posva:
Thank you but the type is correct: the normalized location query can only contain strings and arrays of strings

PR #3172, opened 2020-04-13
feat(history): Remove event listeners when all apps are destroyed.

### LINKED ISSUE #2341: Remove all beforeEach, afterEach, beforeResolve
### Version
3.0.1

### Reproduction link
[https://jsfiddle.net/se6gLaqn/1/](https://jsfiddle.net/se6gLaqn/1/)

### Steps to reproduce
Create a vue instance, as well as a router instance. Then destroy the vue instance. 

The router instance continues to exist and fire.

### What is expected?
The router should be able to properly destroy itself; either by explicit call, or when the vue instance is destroyed

The router exists and cannot be destroyed

---
I did some googling and see that vue-router cannot currently be destroyed, but I am asking for reconsideration. Unlike the other issue I found (https://github.com/vuejs/vue-router/issues/954) where it appeared the user wanted to destroy and create a new router in the same application, the issue I am having is completely separate. 

I have a massive application that essentially uses pjax methodologies to be able to run several SPAs (vue being one) that create the full site. When we navigate away from one SPA to another, the SPA needs to destroy itself so it's code does not interfere with the other SPAs. However, because the vue router cannot be destroyed; it's attempting to execute logic for routing even when the SPA that created it is destroyed. The issue compounds when we have more than 1 vue application running, as there are now 2 instances of vue-router (or more) running at a time. 

Specifically, we are using the hash version of routing. It looks like the router just attaches an event to the window and listens for popstates, but things get weird when more than one event does something with popstate (an understandably oddball behavior).  I would appreciate reconsideration for allowing vue-router to destroy itself when appropriate. Thanks!

[edited for spelling/grammar]

### LINKED ISSUE #3152: Ability to detach and reattach popstate listeners
### Version
3.1.6

### Reproduction link
https://codesandbox.io/s/nostalgic-mahavira-y0m7q

#### Memory leak

Call `new VueRouter()` multiple times. It is impossible to clean up its popstate listener.

#### Basepath bug

If you have multiple vue applications, each using vue-router, they must all share the same base path, even if you ensure only one of them is active at a time by destroying / remounting when navigating between them. Otherwise they will all attempt to ensure that every route starts with their base path.

Same goes for if you have one vue-router and one react-router. The vue-router will force all urls to start with its base path within the react-owned pages, even if you have destroyed the vue application before navigating to the react application.

### What is expected?
I think the best solution is to call `window.addEventListener('popstate', handler)` when `const intance = new Vue({router})` is called, instead of when `new VueRouter()` is called. And then call `window.removeEventListener('popstate', handler)` when `instance.$destroy()` is called.

Alternatively, it could be cleaned up like so:

```js
// Alternative solution (not my preferred one, but works)

const router = new VueRouter({...})
// later, to clean up event listeners:
router.detach()
// later, to reattach event listeners
router.attach()
```

window.addEventListener is called during `new VueRouter()`, and there is no way to ever remove that event listener.

---
See related https://github.com/single-spa/single-spa/issues/488. I am a maintainer of single-spa, which has dozens of organizations using one or more instances of vue-router. For us, cleaning up the event listener when the application is unmounted is important to being able to have isolated apps on the same page.

internettrans:
This resolves #3152 and resolves #2341

This is my first PR to vue-router and I am not sure if this was the best approach to solving the problem - please let me know! I also need to think of a good way of testing this behavior in either the e2e or unit tests. The unit tests only use the the abstract mode, but the feature only applies to html5/hash modes. And for e2e tests, I need to brainstorm the best way to verify that the hashchange and popstate listeners are being removed.

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -1,6 +1,30 @@
 import Vue from 'vue'
 import VueRouter from 'vue-router'
 
+// track number of popstate listeners
+// This should be replaced with something better
+let numPopstateListeners = 0
+const listenerCountDiv = document.createElement('div')
+listenerCountDiv.textContent = numPopstateListeners + ' popstate listeners'
+document.body.appendChild(listenerCountDiv)
+
+const originalAddEventListener = window.addEventListener
+const originalRemoveEventListener = window.removeEventListener
+window.addEventListener = function (name, handler) {
+  if (name === 'popstate') {
+    listenerCountDiv.textContent =
+      ++numPopstateListeners + ' popstate listeners'
+  }
+  return originalAddEventListener.apply(this, arguments)
+}
+window.removeEventListener = function (name, handler) {
+  if (name === 'popstate') {
+    listenerCountDiv.textContent =
+      --numPopstateListeners + ' popstate listeners'
+  }
+  return originalRemoveEventListener.apply(this, arguments)
+}
+
 // 1. Use plugin.
 // This installs <router-view> and <router-link>,
 // and injects $router and $route to all router-enabled child components
@@ -55,6 +79,7 @@ new Vue({
       <pre id="query-t">{{ $route.query.t }}</pre>
       <pre id="hash">{{ $route.hash }}</pre>
       <router-view class="view"></router-view>
+      <button v-on:click="teardown">Teardown app</button>
     </div>
   `,
 
@@ -66,6 +91,10 @@ new Vue({
       } else {
         this.$router.push('/', increment)
       }
+    },
+    teardown () {
+      this.$destroy()
+      this.$el.innerHTML = ''
     }
   }
 }).$mount('#app')
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -2,9 +2,9 @@ import Vue from 'vue'
 import VueRouter from 'vue-router'
 
 // track number of popstate listeners
-// This should be replaced with something better
 let numPopstateListeners = 0
 const listenerCountDiv = document.createElement('div')
+listenerCountDiv.id = 'popstate-count'
 listenerCountDiv.textContent = numPopstateListeners + ' popstate listeners'
 document.body.appendChild(listenerCountDiv)
 
@@ -79,7 +79,7 @@ new Vue({
       <pre id="query-t">{{ $route.query.t }}</pre>
       <pre id="hash">{{ $route.hash }}</pre>
       <router-view class="view"></router-view>
-      <button v-on:click="teardown">Teardown app</button>
+      <button id="teardown-app" v-on:click="teardown">Teardown app</button>
     </div>
   `,
 
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -51,7 +51,7 @@ const router = new VueRouter({
 // 4. Create and mount root instance.
 // Make sure to inject the router.
 // Route components will be rendered inside <router-view>.
-new Vue({
+const vueInstance = new Vue({
   router,
   data: () => ({ n: 0 }),
   template: `
@@ -79,7 +79,6 @@ new Vue({
       <pre id="query-t">{{ $route.query.t }}</pre>
       <pre id="hash">{{ $route.hash }}</pre>
       <router-view class="view"></router-view>
-      <button id="teardown-app" v-on:click="teardown">Teardown app</button>
     </div>
   `,
 
@@ -91,10 +90,11 @@ new Vue({
       } else {
         this.$router.push('/', increment)
       }
-    },
-    teardown () {
-      this.$destroy()
-      this.$el.innerHTML = ''
     }
   }
 }).$mount('#app')
+
+document.getElementById('unmount').addEventListener('click', () => {
+  vueInstance.$destroy()
+  vueInstance.$el.innerHTML = ''
+})
--- a/examples/multi-app/app.js
+++ b/examples/multi-app/app.js
@@ -1,0 +1,76 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+// track number of popstate listeners
+let numPopstateListeners = 0
+const listenerCountDiv = document.getElementById('popcount')
+listenerCountDiv.textContent = 0
+
+const originalAddEventListener = window.addEventListener
+const originalRemoveEventListener = window.removeEventListener
+window.addEventListener = function (name, handler) {
+  if (name === 'popstate') {
+    listenerCountDiv.textContent =
+      ++numPopstateListeners
+  }
+  return originalAddEventListener.apply(this, arguments)
+}
+window.removeEventListener = function (name, handler) {
+  if (name === 'popstate') {
+    listenerCountDiv.textContent =
+      --numPopstateListeners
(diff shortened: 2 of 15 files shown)


src/history/html5.js
```diff
@@ -8,24 +8,37 @@ import { setupScroll, handleScroll } from '../util/scroll'
...
 export class HTML5History extends History {
+  +initLocation: string
+
   constructor (router: Router, base: ?string) {
     super(router, base)
 
+    this.initLocation = getLocation(this.base)
+  }
+
+  setupListeners () {
```
internettrans, line 19: This is an important new behavior to call out. Previously HTML5History would set up listeners during its constructor, but in this PR it only sets up listeners once setupListeners is called.

My changes to index.js ensure that setupListeners() is called once an app is initialized. I am unsure if the HTML5History class is ever used directly by anything else besides index.js, but if it is then there's a chance this is breaking behavior and we should consider a different implementation.

src/index.js
```diff
@@ -110,18 +116,10 @@ export default class VueRouter {
...
-      }
-      history.transitionTo(
-        history.getCurrentLocation(),
-        setupHashListener,
-        setupHashListener
-      )
+    const setupListeners = () => {
+      history.setupListeners()
     }
+    history.transitionTo(history.getCurrentLocation(), setupListeners, setupListeners)
```
internettrans: There's some new behavior here - previously `transitionTo()` would only get called during app initialization if using html5 or hash history, but now it is being called for abstract mode, too. I did so not because abstract mode needs this, but rather because it simplified the if/else code that existed before and appeared to have no harmful side effects.

posva: For this part, it's better to keep the previous behavior because abstract mode cannot always directly transition: e.g. during SSR whereas History and Hash can retrieve the current location on browsers

posva: I also realized we also need to keep the previous behavior for hash history for the hash event not to fire too early on some browsers

internettrans: >  it's better to keep the previous behavior because abstract mode cannot always directly transition:

👍 updated

> I also realized we also need to keep the previous behavior for hash history for the hash event not to fire too early on some browsers

That behavior still exists in this PR, as far as I can tell. HashHistory does not set up listeners until `setupListeners()` is called, and `setupListeners()` is not called until the first `history.transitionTo()` finishes up (as shown in the line of code above)

internettrans: The thing that's different here is that now that behavior for HashHistory is also applying to Html5History - beforehand the popstate listener was set up before the first transition, but it now is set up only after the first transition.

I think that new behavior is completely fine. My guess as to why HashHistory needed that special behavior is that hashchange events are fired as a result of programmatic navigation (via `location.hash = '#/new'`), whereas the browser **never** fires popstate events  never for programmatic navigation   (via `history.pushState(state, title, '/new')`)

internettrans: Oops I made the changes related to Abstract History weeks ago but forgot to push!! I commented and said "Updated" even though I didn't push. I have now pushed my changes.

internettrans:
This is ready for another review.

posva:
Sorry for the delay on this. I will give it a deeper look and hopefully get it merged soon

examples/basic/app.js
```diff
@@ -1,6 +1,30 @@
 import Vue from 'vue'
 import VueRouter from 'vue-router'
 
+// track number of popstate listeners
```
internettrans, line 4: I'm open to better ways of testing the behavior. This one directly tracks the number of popstate listeners, but we could also try to identify a specific bug that occurs due to multiple popstate events and then change the test to verify that that bug is fixed.

src/history/html5.js
```diff
@@ -8,24 +8,37 @@ import { setupScroll, handleScroll } from '../util/scroll'
...
+    const router = this.router
     const expectScroll = router.options.scrollBehavior
     const supportsScroll = supportsPushState && expectScroll
 
     if (supportsScroll) {
-      setupScroll()
+      this.listeners.push(setupScroll())
     }
 
-    const initLocation = getLocation(this.base)
```
posva, line 21: I think we should still consider the initial location when listeners are attached and avoid storing a property `initLocation` on the instance. Were you thinking about something in particular when you added it?

src/index.js
```diff
@@ -98,6 +98,12 @@ export default class VueRouter {
       // ensure we still have a main app or null if no apps
       // we do not release the router so it can be reused
       if (this.app === app) this.app = this.apps[0] || null
+
+      if (this.apps.length === 0) {
```
posva: we should be able to just check `!this.app`

posva (changes requested):
This is a great PR! Thanks a lot for the thought and work put on this.

I think we could create a new e2e test specifically for multi apps and add a few more cases. Similar to what I have for vue-router-next (https://github.com/vuejs/vue-router-next/blob/master/e2e/multi-app/index.ts) where we can mount and unmount apps on demand. The code is a bit different because it's for Vue 3, but let me know if you need help for it. I can also be available on Discord to chat about it

src/history/html5.js
internettrans: Hmm good question - I was under the impression that the "async guard" and "first popstate event" described on lines 38 and 39 were something that occurred only during initial page load, and not during subsequent navigation events.

But rethinking this, I guess the if statement on line 41 exists as a way to short circuit when we know that `transitionTo()` is going to fire during the `init` function in index.js?

Am happy to take direction on this one, as I'm less familiar than you on this code. Another question is whether we should consider resetting this.current to be START during teardownListeners()?

src/index.js
internettrans: 👍 done

internettrans:
Thanks for the review - I have updated this with your feedback.

src/history/html5.js
```diff
@@ -8,24 +8,37 @@ import { setupScroll, handleScroll } from '../util/scroll'
...
-    window.addEventListener('popstate', e => {
+    const handleRoutingEvent = () => {
+      const expectScroll = router.options.scrollBehavior
+      const supportsScroll = supportsPushState && expectScroll
```
posva: can't we remove these lines since they are right above? They shouldn't change

examples/multi-app/app.js
```diff
@@ -0,0 +1,76 @@
...
+            <li><router-link to="/foo">/foo</router-link></li>
+          </ul>
+          <router-view class="view"></router-view>
+          <button id="unmount${n}" v-on:click="teardown">Teardown app</button>
+        </div>
+      `,
+
+      methods: {
+        teardown () {
+          this.$destroy()
```
posva: It's better to keep this code outside of the instance being destroyed:

- adding unmount buttons next to the mounting ones
- keep an array of the instances created by `new Vue`
- remove the entry from the array when destroying

posva (changes requested):

internettrans:
I updated this with your feedback.

posva (approved):
Thanks a lot for this @joeldenning

shichuanpo:
when function setupListeners execute after "hook:destroyed" event, this cannot work correctly! 
It works if add code as below:
image

PR #3174, opened 2020-04-16
better error message when component is a promise (fix #3167)

### LINKED ISSUE #3167: Better error message Component type is invalid
### Version
3.1.5

### Reproduction link
[https://codesandbox.io/s/vue-routing-example-em6dm?file=/router/index.js](https://codesandbox.io/s/vue-routing-example-em6dm?file=/router/index.js)

Visit the `/hello` page in the codesandbox reproduction and open the console to see the error message

### Steps to reproduce
Setup lazy-loaded routes and make a mistake by forgetting that it must not be an import but a function that returns the import statement
```
const routes = [
  {
    path: '/',
    name: 'Home',
    component: import(/* webpackChunkName: "Home" */ '../views/Home.vue'),
  }
]
```

### What is expected?
An error message that says the component is not valid (because it is a promise) and that the expected type is a function

Vue shows the following error message: 
```
Failed to mount component: template or render function not defined.
```

SimonSiefke:
Closes #3167

### DIFF
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -70,6 +70,12 @@ function addRouteRecord (
         path || name
       )} cannot be a ` + `string id. Use an actual component instead.`
     )
+    assert(
+      route.component.toString() !== '[object Promise]',
+      `route config "component" for path: ${String(
+        path || name
+      )} cannot be a promise. Use a function that returns a promise instead (e.g. \`() => import ('./components/About.vue')\`)`
+    )
   }
 
   const pathToRegexpOptions: PathToRegexpOptions =
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -73,9 +73,7 @@ function addRouteRecord (
     if (route.component && route.component.toString) {
       assert(
         route.component.toString() !== '[object Promise]',
-        `route config "component" for path: ${String(
-          path || name
-        )} cannot be a promise. Use a function that returns a promise instead (e.g. \`() => import ('./components/About.vue')\`)`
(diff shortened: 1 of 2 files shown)


posva (changes requested):
I think this could help a few people, could you align with the message on vue-router-next (https://github.com/vuejs/vue-router-next/blob/master/src/navigationGuards.ts#L173) and add a unit test please?

src/create-route-map.js
```diff
@@ -70,6 +70,12 @@ function addRouteRecord (
         path || name
       )} cannot be a ` + `string id. Use an actual component instead.`
     )
+    if (route.component && route.component.toString) {
+      assert(
+        route.component.toString() !== '[object Promise]',
+        `Component "${String(name)}" in record with path "${path}" is a Promise instead of a function that returns a Promise. Did you write "import('./MyPage.vue')" instead of "() => import('./MyPage.vue')"?`
```
posva, line 76: The name here is the name of the route, not the name of the view (eg default) like at https://github.com/vuejs/vue-router-next/blob/master/src/navigationGuards.ts#L168

src/create-route-map.js
```diff
@@ -70,6 +70,12 @@ function addRouteRecord (
         path || name
       )} cannot be a ` + `string id. Use an actual component instead.`
     )
+    if (route.component && route.component.toString) {
+      assert(
+        route.component.toString() !== '[object Promise]',
```
posva, line 75: Let's check for `then` in case of poly fills like at https://github.com/vuejs/vue-router-next/blob/master/src/navigationGuards.ts#L171

posva (changes requested):
There is still a need to check for `components` option. I think it makes more sense to check for this when trying to fetch the component like it's done at https://github.com/vuejs/vue-router-next/blob/master/src/navigationGuards.ts

PR #3184, opened 2020-04-30
feat(link): add router-link-inactive

### LINKED ISSUE #2648: Allow defining an inactive-class
We already have prop `active-class` that will be appended to the element when the route is active.
But what if that we need to specify class only when the route is not active?

example:
when active => `.btn-primary`
when inactive => `.btn-secondary`

```
<router-link
    :to="{name: 'purchase'}"
    class="btn"
    active-class="btn-primary"
    inactive-class="btn-secondary">
    Purchase
</router-link>
```

mdwheele:
Resolves #2648

Hello!

💁 This PR adds an `inactive-class` prop to `<router-link>`, which will add a class to all inactive links (by default, uses `.router-link-inactive`)

Example:

```vue
<router-link :to="link" class="all links" active-class="active only" inactive-class="inactive only">
  Click me
</router-link>
```

This feature would be _killer_ for anybody using [Tailwind CSS](https://tailwindcss.com).

Please let me know if you would like me to add any additional e2e tests covering this. I don't know if there are any special behaviours around exact-match links or things like that we should cover.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -133,6 +133,13 @@ If you add a `target="_blank"` to your `a` element, you must omit the `@click="n
 
   Configure the active CSS class applied when the link is active. Note the default value can also be configured globally via the `linkActiveClass` router constructor option.
 
+### inactive-class
+
+- type: `string`
+- default: `"router-link-inactive"`
+
+  Configure the CSS class applied when the link is inactive. Note the default value can also be configured globally via the `linkInactiveClass` router constructor option.
+
 ### exact
 
 - type: `boolean`
--- a/examples/inactive-links/app.js
+++ b/examples/inactive-links/app.js
@@ -1,0 +1,30 @@
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+Vue.use(VueRouter)
+
+const Home = { template: '<div><h2>Home</h2></div>' }
+const About = { template: '<div><h2>About</h2></div>' }
+
+const router = new VueRouter({
+  mode: 'history',
+  base: __dirname,
+  routes: [
+    { path: '/', component: Home },
+    { path: '/about', component: About }
+  ]
+})
+
+new Vue({
+  router,
+  template: `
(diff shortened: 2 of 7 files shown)


shaharmor:
Super helpful for everyone using a utility-based css framework like Tailwind.

iamarpitpatidar (commented):
Looka good to me, @posva what do you think?

tobihagemann:
Just stumbled upon this. Looks like this was already solved a long time ago and just needs to be merged?

mdwheele:
@tobihagemann At the time (more than a year ago now) that this was submitted, there was a lot of working going into `vue-router@4` and `@posva` was very busy. Check out the related issue above for more detail, but ultimately, a solution is posted in the documentation for Vue Router 4: https://next.router.vuejs.org/guide/advanced/extending-router-link.html.

This documentation details an `<AppLink>` component that extends `RouterLink` and gives you everything needed to implement an `inactive-class` property. With that said, I still think a flavor of this belongs in `vue-router` because I've now implemented this component 10+ times in different project. It's an absolute requirement if using TailwindCSS (or TailwindUI). 

I'd be happy to dig into this again and help move it forward if the maintainers are interested.

mdwheele:
I'm closing this out; clearing lines of work with no clear path forwards resolution. If there is renewed interest in this feature, I am happy to contribute but reimplementing this in every project is not too bad.

malobre:
For anyone using Tailwind CSS, if your classes are not too numerous you can emulate this feature by using the [important modifier](https://tailwindcss.com/docs/configuration#important-modifier) for active classes, e.g:
```html
<router-link
  :to="{ name: 'root' }"
  active-class="!border-red-500 !text-gray-900"
  class="border-transparent text-gray-500 hover:border-gray-300 hover:text-gray-700 inline-flex items-center px-1 pt-1 border-b-2 text-sm font-medium"
>
  Home
</router-link>
```

PR #3189, opened 2020-05-05
feat(matcher): Adds matchAll to Matcher

mhiguera:
There are cases in which different routes can be matched in runtime (eg: dynamic importing a module within a component). Adding a `matchAll` method helps developers create a fallback strategy for those special cases.

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -10,6 +10,7 @@ import { normalizeLocation } from './util/location'
 
 export type Matcher = {
   match: (raw: RawLocation, current?: Route, redirectedFrom?: Location) => Route;
+  matchAll: (raw: RawLocation, current?: Route, redirectedFrom?: Location) => Array<Route>;
   addRoutes: (routes: Array<RouteConfig>) => void;
 };
 
@@ -23,11 +24,12 @@ export function createMatcher (
     createRouteMap(routes, pathList, pathMap, nameMap)
   }
 
-  function match (
+  function matchAll (
     raw: RawLocation,
     currentRoute?: Route,
-    redirectedFrom?: Location
-  ): Route {
+    redirectedFrom?: Location,
+    matchFirst?: boolean
+  ): Array<Route> {
     const location = normalizeLocation(raw, currentRoute, false, router)
     const { name } = location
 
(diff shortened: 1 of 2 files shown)


posva:
Can you please give more information about the cases?

Without any information, this new API is very unlikely to happen because, in v4, dynamic routing makes the matcher always match the desired route

mhiguera:
Thanks @posva. Sure, I'll try to explain myself better.

For example, suppose you have two dynamic routes (`/:form_id` and `/:content_id`, for simplicity sake) and we have a content published with ID `d41d-8cd9-8f00`. 

The `:form` route could use a dynamic import, something like `import('@/forms/ + this.$route.params.form_id)`, or an API call to check its existence beforehand. 

In that particular case you could waterfall through the defined routes (mimicking Express' `next('route')`, e.g.) with something like:

```javascript
router.beforeResolve((to, from, next) => {
  const defaultComponent = to.matched[0].components.default
  const resolves = defaultComponent.resolves
  if ('undefined' === typeof resolves || resolves(to)) return next()
  const path = to.path
  const allMatches = router.matchAll(path)
  const index = allMatches.findIndex(m => m.matched[0].name == to.name)
  const nextRoute = allMatches[index+1]
  if (nextRoute) router.replace(nextRoute)
})
```
Please notice I'm adding `resolves(route)` to the default component of the routes that need that existence check. 

It's feasible to have forms and contents multiplexed using a different approach but IMHO adding complexity and coupling between components. I find it particularly useful, specially when a project grows bigger, because it shifts responsibility to the order of the routes, thus eliminating the need to substantially modify the behavior of the components.

posva:
If you have multiple routes that have the same path (those two have one single param), you need a way to differentiate them. One way is by using a custom regex:

```
/:form_id(f_\d+)
/:content_id(c_\d+)
```

The router should be able to find the right route from a string representation of the path (URL)

mhiguera:
You are right, that's the way to go when you have predictable formats. If it's multipurpose (or user-generated) slugs, you'll find yourself tethered.

posva:
Okay! Let's close this then as there should be only one valid match when resolving a route.

If you think this is worth pursuing, it should go through the RFC process: https://github.com/vuejs/rfcs to ensure enough information is provided

PR #3199, opened 2020-05-13
fix(scroll): run scrollBehavior on initial load (fix #3196)

### LINKED ISSUE #3196: scrollBehavior does not run on load
### Version
3.1.6

### Reproduction link
[https://codesandbox.io/s/vue-routing-example-pk4q2](https://codesandbox.io/s/vue-routing-example-pk4q2)

### Steps to reproduce
Load page

### What is expected?
scrollBehavior always return `{ x: 0, y: 200 }` so the scroll position should always be 200 pixels from the top

scrollBehavior always return `{ x: 0, y: 200 }` but scroll position y is always 0 when first loaded

Calvin-LL:
fix #3196 by adding handleScroll to the callback of the first call to history.transitionTo


posva:
I will first go through https://github.com/vuejs/vue-router/pull/3172 which might create merge conflicts. This will also require an e2e test

Calvin-LL:
> I will first go through #3172 which might create merge conflicts. This will also require an e2e test

Sounds good I'll go add the e2e tests

Calvin-LL:
> I will first go through #3172 which might create merge conflicts. This will also require an e2e test

e2e tests added

Calvin-LL:
> I will first go through #3172 which might create merge conflicts. This will also require an e2e test

Merge conflicts are resolved. Looks like Netlify fail to deploy because of 8d26c5d, not related to this PR

Mike377:
Is this going to be merged soon? Having the same problem right now.

test/e2e/specs/scroll-behavior.js
```diff
@@ -119,6 +119,44 @@ module.exports = {
...
+          return document.getElementById('anchor').getBoundingClientRect().top < 1
+        },
+        null,
+        'scroll to anchor on load'
+      )
+      .url('http://localhost:8080/scroll-behavior/bar#anchor2')
+      .execute(function () {
+        location.reload(true)
+      })
+      .waitForElementVisible('#app', 1000)
```
posva: do we really need the other tests? At the end they are calling the same scrollHandler

src/index.js
```diff
@@ -117,8 +118,17 @@ export default class VueRouter {
     const history = this.history
 
     if (history instanceof HTML5History || history instanceof HashHistory) {
-      const setupListeners = () => {
+      const handleInitialScroll = (route) => {
+        const expectScroll = this.options.scrollBehavior
+        const supportsScroll = supportsPushState && expectScroll
+
+        if (supportsScroll) {
+          handleScroll(this, route, route, false)
```
posva: I think it would make sense to pass the initial location as well so get the initial location outside of `handleInitialScroll`

```js
const from = history.current
```

and then do

```js
handleScroll(this, route, from, false)
```

posva (changes requested):

test/e2e/specs/scroll-behavior.js
Calvin-LL: make sense, removed redundant tests

src/index.js
Calvin-LL: make sense, changed in new commit

posva (approved):
Thank you!

PR #3204, opened 2020-05-19
fix: Missing renaming

lmichelin:
This PR fixes a missing renaming in #3047 

See this comment for more details: https://github.com/vuejs/vue-router/pull/3047#pullrequestreview-414419587

### DIFF
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -51,7 +51,7 @@ export class AbstractHistory extends History {
         this.updateRoute(route)
       },
       err => {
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks for the quick PR! This is the kind of things that are caught with TypeScript 😓

schmidtk:
> Thanks for the quick PR! This is the kind of things that are caught with TypeScript 😓

Or the Closure Compiler, if you're into that sort of thing. 😅

PR #3210, opened 2020-05-27
feat(scroll): allow behavior in scrollBehavior

### LINKED ISSUE #2069: pass on extra arguments to `scrollTo` to enable smooth scrolling with behaviour
Smooth scrolling to same page anchor tags. 

Right now smooth scrolling can only work by custom implementations, but the new Firefox spec contemplates a scrollBehaviour css property which can be polyfilled in older browsers.

Vue-router should take advantage of this new property

Basically, the object returned from the scrollBehavior function should have some more parameters and change the current x, y positions into named parameters.

This is an example of the future object to be returned.

{ 
    top: 2500, // this is the x value
    left: 0, // this is the y value
    behavior: 'smooth' //
  };

This is getting implemented in most current browsers, couldn't find the spec thought. If anybody finds it please sahre the link.

Anyway, there is this beatyfull polyfill: https://www.npmjs.com/package/smoothscroll-polyfill

which adds this scrollTo({}) behaviour to old browsers.

I haven't dig much into the vue-router code. But it looks to me as if the only change needed is this file:
https://github.com/vuejs/vue-router/blob/dev/src/util/scroll.js

Maybe we could add some conditions on line 126 to return the current implementation or translate the old object:

{ x, y } to

{ top: x, left: y, behaviour: x }

Sorry if markup sucks, I'll do some more work if this gets atention

### LINKED ISSUE #3008: Better error message when the scrollBehavior selector is invalid

posva:
Closes #2069
Closes #3008

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import type Router from '../index'
-import { assert } from './warn'
+import { assert, warn } from './warn'
 import { getStateKey, setStateKey } from './state-key'
 import { extend } from './misc'
 
@@ -82,8 +82,8 @@ export function saveScrollPosition () {
   const key = getStateKey()
   if (key) {
     positionStore[key] = {
-      x: window.pageXOffset,
-      y: window.pageYOffset
+      left: window.pageXOffset,
+      top: window.pageYOffset
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/util/scroll.js
```diff
@@ -107,26 +107,33 @@ function getElementPosition (el: Element, offset: Object): Object {
...
 
-function normalizePosition (obj: Object): Object {
-  return {
-    x: isNumber(obj.x) ? obj.x : window.pageXOffset,
-    y: isNumber(obj.y) ? obj.y : window.pageYOffset
+function normalizePosition (obj: Object, defaultLeft?: number, defaultTop?: number): Object {
+  const left = isNumber(obj.left) ? obj.left : obj.x
+  const top = isNumber(obj.top) ? obj.top : obj.y
+  // defaultLeft = isNumber(defaultLeft) ? defaultLeft : window.pageXOffset
+  // defaultTop = isNumber(defaultTop) ? defaultTop : window.pageYOffset
```
pikax, line 124: remove comment code

pikax (approved):
LGTM

src/util/scroll.js
posva: I need to add it back 👀

posva:
I've had a better idea for this API: https://github.com/vuejs/vue-router/issues/3008#issuecomment-635268418

PR #3211, opened 2020-05-29
fix(errors): NavigationCanceled with async components

lbwa:
Canceled behavior should always throw a same canceled error, instead of `undefined`

Same as these codes:

https://github.com/vuejs/vue-router/blob/c8a4cefcb1a570245040e80bbec20cfe476a1c85/src/history/base.js#L160-L162

Related PR:

- https://github.com/vuejs/vue-router/pull/3047

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -200,7 +200,7 @@ export class History {
       const queue = enterGuards.concat(this.router.resolveHooks)
       runQueue(queue, iterator, () => {
         if (this.pending !== route) {
-          return abort()
+          return abort(createNavigationCancelledError(current, route))
         }
         this.pending = null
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Nice catch, thank you!
Can you remove the changes in the example folder and add a unit test instead like https://github.com/vuejs/vue-router/pull/3047/files#diff-c044368ddb9969465784dfa4ebfa75cd

lbwa:
I will handle it soon

lbwa:
@posva I have reverted redundant modifications and added extra unit test for PR

posva (approved):
Thanks!

PR #3212, opened 2020-05-29
fix: error.stack is possibly null

hareku:
In unsupported `Error.stack` browsers, that throws `TypeError(Unable to get property 'split' of undefined or null reference)`.

### DIFF
--- a/src/history/errors.js
+++ b/src/history/errors.js
@@ -48,9 +48,12 @@ function createRouterError (from, to, type, message) {
   error.to = to
   error.type = type
 
-  const newStack = error.stack.split('\n')
-  newStack.splice(1, 2) // remove 2 last useless calls
-  error.stack = newStack.join('\n')
+  if (typeof error.stack === 'string') {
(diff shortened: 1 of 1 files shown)


src/history/errors.js
```diff
@@ -48,9 +48,12 @@ function createRouterError (from, to, type, message) {
...
 
-  const newStack = error.stack.split('\n')
-  newStack.splice(1, 2) // remove 2 last useless calls
-  error.stack = newStack.join('\n')
+  if (typeof error.stack === 'string') {
+    const newStack = error.stack.split('\n')
+    newStack.splice(1, 2) // remove 2 last useless calls
+    error.stack = newStack.join('\n')
+  }
+
```
posva: Let's completely remove this part, it's better to have the full stack trace when debugging
```suggestion
```

posva (approved):
Thank you!

I completely removed the stack modification instead because I think it will cause more harm than it's trying to prevent

PR #3213, opened 2020-05-29
Catch URI malformed error (fix #2725)

### LINKED ISSUE #2725: Properly encode/decode the percentage character (% -> %25) across browsers and upon initial navigation

aradjdi:
Fix url decoding when we have a percentage character in URL

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -185,7 +185,16 @@ function matchRoute (
 
   for (let i = 1, len = m.length; i < len; ++i) {
     const key = regex.keys[i - 1]
-    const val = typeof m[i] === 'string' ? decodeURIComponent(m[i]) : m[i]
+    let val = m[i]
+    if (typeof m[i] === 'string') {
+      try {
+        val = decodeURIComponent(m[i])
+      } catch (error) {
+        if (error.toString() !== 'URIError: URI malformed') {
+          throw error
+        }
+      }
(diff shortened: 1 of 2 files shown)


posva:
Thank you but this is not the right fix. The problem is the codebase is decoding twice. Example going to `/%252520` will create a path of `/%2520` and if there is a param (`/:text`), the param will be `%20` but the path should be the same as the one in the URL and the param should be `/%2520` (first percentage encoded character is decoded)

claudiolcastro:
Hello @posva,

Just removing the `decodeURIComponent` from `matchRoute` would resolve the issue??

Can you tell me where is the other place that the param is been decoded??

thanks!

PR #3218, opened 2020-06-02
feat(types): RouterConfig for multiple components (#3217)

### LINKED ISSUE #3217: Accurate typings for Routeconfig: component, components, props
The typings have the following problems
- A configuration can have both `component` and `components`
- When multiple named components are used, props should be a `Dictionary` of the regular props typings. Right now it doesn't complain because `Object` is among the regular props typings, however it cannot infer the accurate typing of each of the props, it just uses `any`.

There is no API

javiertury:
Closes #3217

The test demonstrates that with correct typings, typescript is able to infer that route is of type `Route` on its own.
```
             abc: route => route.params,
```

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -94,7 +94,6 @@ interface _RouteConfigBase {
   name?: string
   redirect?: RedirectOption
   alias?: string | string[]
-  children?: RouteConfig[]
   meta?: any
   beforeEnter?: NavigationGuard
   caseSensitive?: boolean
@@ -108,6 +107,7 @@ interface RouteConfigSingleView extends _RouteConfigBase {
 
 interface RouteConfigMultipleViews extends _RouteConfigBase {
   components?: Dictionary<Component>
+  children?: RouteConfig[]
   props?: Dictionary<boolean | Object | RoutePropsFunction>
 }
 
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -89,20 +89,29 @@ export interface PathToRegexpOptions {
   end?: boolean
 }
 
-export interface RouteConfig {
+interface _RouteConfigBase {
   path: string
   name?: string
-  component?: Component
-  components?: Dictionary<Component>
   redirect?: RedirectOption
   alias?: string | string[]
   children?: RouteConfig[]
   meta?: any
   beforeEnter?: NavigationGuard
-  props?: boolean | Object | RoutePropsFunction
   caseSensitive?: boolean
   pathToRegexpOptions?: PathToRegexpOptions
 }
+
+interface RouteConfigSingleView extends _RouteConfigBase {
+  component?: Component
+  props?: boolean | Object | RoutePropsFunction
+}
+
+interface RouteConfigMultipleViews extends _RouteConfigBase {
+  components?: Dictionary<Component>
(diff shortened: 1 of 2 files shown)


types/router.d.ts
```diff
@@ -104,6 +104,14 @@ export interface RouteConfig {
   pathToRegexpOptions?: PathToRegexpOptions
 }
 
+export type RouteConfig = BaseRouteConfig & ({
```
posva: can you instead create 2 interfaces `RouteConfigSingleView` and `RouteConfigMultipleViews` that extend the base config and export RouteConfig as the type union?

types/router.d.ts
```diff
@@ -89,7 +89,7 @@ export interface PathToRegexpOptions {
   end?: boolean
 }
 
-export interface RouteConfig {
+interface BaseRouteConfig {
```
posva: Let's change the name

```suggestion
interface _RouteConfigBase {
```

posva (changes requested):
Thanks for this!

javiertury:
Done

types/router.d.ts
```diff
@@ -89,21 +89,30 @@ export interface PathToRegexpOptions {
...
 
-export interface RouteConfig {
+interface _RouteConfigBase {
   path: string
   name?: string
-  component?: Component
-  components?: Dictionary<Component>
   redirect?: RedirectOption
   alias?: string | string[]
   children?: RouteConfig[]
```
```suggestion
```

types/router.d.ts
```diff
@@ -89,21 +89,30 @@ export interface PathToRegexpOptions {
...
   pathToRegexpOptions?: PathToRegexpOptions
 }
 
+interface RouteConfigSingleView extends _RouteConfigBase {
+  component?: Component
+  props?: boolean | Object | RoutePropsFunction
+}
+
+interface RouteConfigMultipleViews extends _RouteConfigBase {
+  components?: Dictionary<Component>
```
posva, line 109:
```suggestion
  components?: Dictionary<Component>
  children?: RouteConfig[]
```

posva (approved):
Thanks a lot. Let me just move the children next to components as well and I will merge this

andersbc:
The associated commit broke my ts use case

I have a custom interface for my Router configuration, e.g.:

```ts
// file: routes.ts
// ...
import { RouteConfig } from 'vue-router'

interface MyRouteConfig extends RouteConfig {
  meta?: MyMeta; // <- defined elsewhere
}

export const routes: Array<MyRouteConfig> = [
  // ... the usual route configuration, with my added meta-thing
]

```

After the commit, the `RouteConfig` is now a union type, i.e.. `RouteConfig = RouteConfigSingleView | RouteConfigMultipleViews`

However, typescript can't extend from union types, because "An interface can only extend an object type or intersection of object types with statically known members" 

What I really need is to extends the new `RouteConfigMultipleViews` type, but it doesn't get exported. Any chance of exporting it? Thanks :-)

javiertury:
That was a case I missed.

They are trying to fix it in #3234. Personally I think that using a generic `RouteConfig<Meta = any>` [as proposed in this comment](https://github.com/vuejs/vue-router/issues/3234#issuecomment-644683914) is a good solution.

posva:
FYI, adding a generic to the type should go through an RFC because is not that simple of a change

PR #3220, opened 2020-06-05
docs: add navigation failures

posva:
Based on https://github.com/vuejs/rfcs/pull/150

### DIFF
--- a/docs/guide/advanced/navigation-failures.md
+++ b/docs/guide/advanced/navigation-failures.md
@@ -1,0 +1,53 @@
+# Navigation Failures
+
+When using `router-link`, Vue Router internally calls `router.push` to trigger a navigation. Depending on the current location and existing [Navigation Guards](./navigation-guards.md), this navigation might end up in a new page being shown, but there are a couple of situations where we will stay on the same page:
+
+- We are already on the page we are trying to go to
+- A navigation guard aborts the navigation by calling `next(false)`
+- An error is thrown in a navigation by calling `next(new Error())`
+
+When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `osComplete` and rejects instead of invoking `onAbort`.
+
+## Detecting Navigation Failures
+
+_Navigation Failures_ are `Error` instances with a few extra properties. Among them, you can find a `type` property. This will allow you to check the type of the navigation failure:
+
+```js
+import { NavigationFailureTypes } from 'vue-router'
+
+// trying to access an admin-only route
+router.push('/admin').catch(failure => {
+  if (failure) {
+    if (failure.type === NavigationFailureTypes.redirected) {
+      // show a small notification to the user
+      showToast('Login in order to access the admin panel')
+    }
+  }
+})
+```
+
+## `NavigationFailureTypes`
+
+`NavigationFailureTypes` exposes the following properties to differentiate _Navigation Failures_:
+
+- `redirected`: `next(newLocation)` was called inside of a navigation guard to redirect somewhere else.
+- `aborted`: `next(false)` was called inside of a navigation guard to the navigation.
+- `cancelled`: A new navigation completely took place before the current navigation could finish. e.g. `router.push` was called while waiting inside of a navigation guard.
+- `duplicated`: The navigation was prevented because we are already at the target location.
+
+## _Navigation Failures_'s properties
+
+Apart from exposing a `type` property, all navigation failures expose `to` and `from` properties to reflect the current location as well as the target location for the navigation that failed:
+
+```js
+// given we are at `/`
+router.push('/admin').catch(failure => {
+  if (failure) {
+    if (failure.type === NavigationFailureTypes.redirected) {
+      failure.to.path // '/admin'
+      failure.from.path // '/'
+    }
+  }
+```
+
+In all cases, `from` and `to` are normalized route locations.
--- a/docs/guide/advanced/navigation-failures.md
+++ b/docs/guide/advanced/navigation-failures.md
@@ -10,21 +10,25 @@ When using a regular `router-link`, **none of these failures will log an error**
 
 ## Detecting Navigation Failures
 
-_Navigation Failures_ are `Error` instances with a few extra properties. Among them, you can find a `type` property. This will allow you to check the type of the navigation failure:
+_Navigation Failures_ are `Error` instances with a few extra properties. To check if an error comes from the Router, use the `isRouterError` function:
 
 ```js
-import { NavigationFailureType } from 'vue-router'
+import { NavigationFailureType, isRouterError } from 'vue-router'
 
-// trying to access an admin-only route
+// trying to access the admin page
 router.push('/admin').catch(failure => {
   if (failure) {
-    if (failure.type === NavigationFailureType.redirected) {
+    if (isRouterError(failure, NavigationFailureType.redirected)) {
       // show a small notification to the user
       showToast('Login in order to access the admin panel')
     }
   }
 })
 ```
+
+::: tip
+If you omit the second parameter: `isRouterError(failure)`, it will only check if the error comes from the Router.
+:::
 
 ## `NavigationFailureType`
 
@@ -40,10 +44,10 @@ router.push('/admin').catch(failure => {
 Apart from exposing a `type` property, all navigation failures expose `to` and `from` properties to reflect the current location as well as the target location for the navigation that failed:
 
 ```js
-// given we are at `/`
+// trying to access the admin page
 router.push('/admin').catch(failure => {
   if (failure) {
-    if (failure.type === NavigationFailureType.redirected) {
+    if (isRouterError(failure, NavigationFailureType.redirected)) {
       failure.to.path // '/admin'
       failure.from.path // '/'
     }
--- a/docs/guide/advanced/navigation-failures.md
+++ b/docs/guide/advanced/navigation-failures.md
@@ -13,12 +13,12 @@ When using a regular `router-link`, **none of these failures will log an error**
 _Navigation Failures_ are `Error` instances with a few extra properties. Among them, you can find a `type` property. This will allow you to check the type of the navigation failure:
 
 ```js
-import { NavigationFailureTypes } from 'vue-router'
+import { NavigationFailureType } from 'vue-router'
 
 // trying to access an admin-only route
 router.push('/admin').catch(failure => {
   if (failure) {
-    if (failure.type === NavigationFailureTypes.redirected) {
+    if (failure.type === NavigationFailureType.redirected) {
       // show a small notification to the user
       showToast('Login in order to access the admin panel')
     }
@@ -26,9 +26,9 @@ router.push('/admin').catch(failure => {
 })
 ```
 
-## `NavigationFailureTypes`
+## `NavigationFailureType`
 
-`NavigationFailureTypes` exposes the following properties to differentiate _Navigation Failures_:
+`NavigationFailureType` exposes the following properties to differentiate _Navigation Failures_:
 
 - `redirected`: `next(newLocation)` was called inside of a navigation guard to redirect somewhere else.
 - `aborted`: `next(false)` was called inside of a navigation guard to the navigation.
@@ -43,7 +43,7 @@ Apart from exposing a `type` property, all navigation failures expose `to` and `
 // given we are at `/`
 router.push('/admin').catch(failure => {
   if (failure) {
-    if (failure.type === NavigationFailureTypes.redirected) {
+    if (failure.type === NavigationFailureType.redirected) {
       failure.to.path // '/admin'
       failure.from.path // '/'
     }
--- a/docs/guide/advanced/navigation-failures.md
+++ b/docs/guide/advanced/navigation-failures.md
@@ -6,7 +6,7 @@ When using `router-link`, Vue Router internally calls `router.push` to trigger a
 - A navigation guard aborts the navigation by calling `next(false)`
 - An error is thrown in a navigation by calling `next(new Error())`
 
-When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `osComplete` and rejects instead of invoking `onAbort`.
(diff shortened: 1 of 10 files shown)


docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,53 @@
...
+
+- We are already on the page we are trying to go to
+- A navigation guard aborts the navigation by calling `next(false)`
+- An error is thrown in a navigation by calling `next(new Error())`
+
+When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `osComplete` and rejects instead of invoking `onAbort`.
```
bencodezen: You mention a "regular" `router-link`, but is there a special `router-link` that I should be aware of?

bencodezen: > **none of these failures will log an error**

Will these failures still cause an error even if they're not logged?

NataliaTepluhina: Typo 😉 

```suggestion
When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `onComplete` and rejects instead of invoking `onAbort`.
```

NataliaTepluhina (approved):

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,53 @@
+# Navigation Failures
+
+When using `router-link`, Vue Router internally calls `router.push` to trigger a navigation. Depending on the current location and existing [Navigation Guards](./navigation-guards.md), this navigation might end up in a new page being shown, but there are a couple of situations where we will stay on the same page:
+
+- We are already on the page we are trying to go to
+- A navigation guard aborts the navigation by calling `next(false)`
+- An error is thrown in a navigation by calling `next(new Error())`
```
posva:
```suggestion
- A navigation guard throws an error or calls `next(new Error())`
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,53 @@
...
+
+When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `onComplete` and rejects instead of invoking `onAbort`.
+
+## Detecting Navigation Failures
+
+_Navigation Failures_ are `Error` instances with a few extra properties. Among them, you can find a `type` property. This will allow you to check the type of the navigation failure:
+
+```js
+import { NavigationFailureTypes } from 'vue-router'
```
posva: I was thinking we could also expose an `isRouterError` function:

```js
router.push('/admin').catch(failure => {
   if (isRouterError(failure)) // any router error
   if (isRouterError(failure, NavigationFailureTypes.abort)) // only aborted navigations
})
```

trygveaa: As far as I can see, you're not actually exporting this? Importing it just gives me `undefined` with v3.3.4, and it's not present in the typescript types either.

posva: This is not yet implemented in vue router 3

trygveaa: Okay, I think removing the name before implementing this was a bad idea, but I guess you've gathered as much from the comments in #3047.

I think checking the error type as documented in these docs is fine when `NavigationFailureTypes` is exported. However, as far as I can see, you silently ignore the types you don't check. You should rethrow the error if it's any of the other types.

posva: Yeah, I think will add back the `name` for the only error that exposed a `name` before but it won't be documented anywhere.
It's true the examples could show how a non caught error is thrown

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,57 @@
...
+- A navigation guard throws an error or calls `next(new Error())`
+
+When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `onComplete` and rejects instead of invoking `onAbort`.
+
+## Detecting Navigation Failures
+
+_Navigation Failures_ are `Error` instances with a few extra properties. To check if an error comes from the Router, use the `isRouterError` function:
+
+```js
+import { NavigationFailureType, isRouterError } from 'vue-router'
```
posva: Maybe something like `isNavigationFailure` is better naming than `isRouterError` as there aren't any other errors and it's also confusing for it to be named error and failure

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,57 @@
...
+
+- We are already on the page we are trying to go to
+- A navigation guard aborts the navigation by calling `next(false)`
+- A navigation guard throws an error or calls `next(new Error())`
+
+When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. This is part of the _Navigation Failures_ system in Vue Router and it is only available from version 3.2.0 onwards but existed for a long time before: through the two optional callbacks, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `onComplete` and rejects instead of invoking `onAbort`.
```
shentao: Would it make sense to split this into 2-3 paragraphs? 
1) About the error handling when using `router-link`.
2) Using the `onComplete` and `onAbort` callbacks.
"New in 3.1.0" tag
3) `router.push` and `router.replace` return a _Promise_.

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Detecting Navigation Failures
```
shentao, line 17: I assume this is introduced with v3.2.0 acccording to the previous paragraph. What about adding a tag `New in 3.2.0`?

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,57 @@
...
+
+_Navigation Failures_ are `Error` instances with a few extra properties. To check if an error comes from the Router, use the `isRouterError` function:
+
+```js
+import { NavigationFailureType, isRouterError } from 'vue-router'
+
+// trying to access the admin page
+router.push('/admin').catch(failure => {
+  if (failure) {
```
shentao: Is there a case where this would be _falsy_ even though the `catch` was triggered?

shentao (approved):
Looks good! Just some minor comments.

docs/guide/advanced/navigation-failures.md
posva: Since the Navigation failure system is only available from 3.4.0 (the next release), I don't think it makes much sense to add a new in 3.1.0 but instead we could put a New in 3.4.0 at the very top

How would you split into multiple paragraphs the explanation? They are all related sentences

docs/guide/advanced/navigation-failures.md
posva: No, that shouldn't happen, so I think we can remove this!

docs/guide/advanced/navigation-failures.md
posva: I think I should remove the whole reference to versions if I add something at the very top

posva: The thing is people often want to know why things are like that, that's why I added the version things were added at

docs/guide/advanced/navigation-failures.md
shentao: Great!

docs/guide/advanced/navigation-failures.md
shentao: Agree. I would also lean towards `isRoutingError` given that this is an `Error` instance. So either `isNavigationError` or `isRoutingError`.

posva: Oh I was going the other way around, it's an Error instance to have a stack trace but it's technically not an error as it's expected behavior. Maybe Failure isn't a good name either

docs/guide/advanced/navigation-failures.md
shentao: That makes sense. I wonder though if it is a good practice to mix historical context with informative docs. As in, the historical context might be very interesting to read but likely shouldn’t be required to understand what is happening. Maybe a ::tip:: with informations of how it changed across different versions?

posva: That's true, I updated that part, let me know what you think

shentao: Looks great! Thank you!

docs/guide/advanced/navigation-failures.md
shentao: What about `Rejection`?

shentao: It’s much softer than both `Error` and `Failure` and might be quite good at communicating what is actually happening, as in this navigation request has been rejected. Not an error (this is the expected behaviour/feature). This would also allow us to actually have Errors thrown that are actual errors (as in unexpected).

posva: Yeah, it's softer but it's also related to Promise rejection and in v4, they appear when the promise resolve. It would also require changing the `NavigationRejectionType` name, but that could only be done in v4 and add an amendment to the RFC that was merged

shentao: That makes sense! In that case maybe Exception? Running out of ideas here :(

posva: There is also NavigationDropped, Canceled (one of the types), Stopped

shentao: In that case, I think Failure is a good word.

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
+# Navigation Failures
+
+> New in 3.4.0
+
+When using `router-link`, Vue Router internally calls `router.push` to trigger a navigation. Depending on the current location and existing [Navigation Guards](./navigation-guards.md), this navigation might end up in a new page being shown, but there are a couple of situations where we will stay on the same page:
```
bencodezen:
```suggestion
When using `router-link`, Vue Router calls `router.push` to trigger a navigation. While the expected behavior for most links is to navigate a user to a new page, there are a few situations where users will remain on the same page:
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
...
+
+- We are already on the page we are trying to go to
```
bencodezen:
```suggestion
- Users are already on the page that they are trying to navigate to
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
...
+- A navigation guard aborts the navigation by calling `next(false)`
+- A navigation guard throws an error or calls `next(new Error())`
```
bencodezen:
```suggestion
- A [navigation guard](./navigation-guards.md) aborts the navigation by calling `next(false)`
- A [navigation guard](./navigation-guards.md) throws an error or calls `next(new Error())`
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
...
+
+When using a regular `router-link`, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. Let's understand how to differentiate _Navigation Failures_.
```
bencodezen:
```suggestion
When using a `router-link` component, **none of these failures will log an error**. However, if you are using `router.push` or `router.replace`, you might come across an _"Uncaught (in promise) Error"_ message followed by a more specific message in your console. Let's understand how to differentiate _Navigation Failures_.
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
...
+
+::: tip Background story
+ _Navigation Failures_ were exposed on version 3.2.0 but existed for a long time before: through the two optional callbacks of `router.push`, `onComplete` and `onAbort` that can be passed to `router.push`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `onComplete` and rejects instead of invoking `onAbort`.
```
bencodezen:
```suggestion
 In v3.2.0, _Navigation Failures_ were exposed through the two optional callbacks of `router.push`: `onComplete` and `onAbort`. Since version 3.1.0, `router.push` and `router.replace` return a _Promise_ if no `onComplete`/`onAbort` callback is provided. This _Promise_ resolves instead of invoking `onComplete` and rejects instead of invoking `onAbort`.
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
...
+})
+```
+
+::: tip
+If you omit the second parameter: `isNavigationFailure(failure)`, it will only check if the error is a _Navigation Failure_.
+:::
+
+## `NavigationFailureType`
+
+`NavigationFailureType` exposes the following properties to differentiate _Navigation Failures_:
```
bencodezen:
```suggestion
`NavigationFailureType` help developers to differentiate between the various types of _Navigation Failures_. There are four different types:
```

docs/guide/advanced/navigation-failures.md
```diff
@@ -0,0 +1,60 @@
...
+// trying to access the admin page
+router.push('/admin').catch(failure => {
+  if (isNavigationFailure(failure, NavigationFailureType.redirected)) {
+    failure.to.path // '/admin'
+    failure.from.path // '/'
+  }
+})
+```
+
+In all cases, `from` and `to` are normalized route locations.
```
bencodezen:
```suggestion
In all cases, `to` and `from` are normalized route locations.
```

bencodezen (commented):
Added some suggestions for rephrasing in order to improve clarity and maintain some consistency and parallelism throughout the doc.

docs/guide/advanced/navigation-failures.md
posva: You are right, it makes no sense to talk about regular. I was thinking about using a custom router-link

Navigation failures in router-link will not cause an error either, we pass an empty `onAbort` handler to silence them

docs/guide/advanced/navigation-failures.md
posva: Should we say _that is being navigated to_? as there isn't any request going on or is it still clearer that way?

docs/guide/advanced/navigation-failures.md
posva: should just omit it? or maybe say _component_?

bencodezen: Works for me!

docs/guide/advanced/navigation-failures.md
bencodezen: Updated!

posva:
Thanks a lot @NataliaTepluhina @shentao, @trygveaa  and @bencodezen for the feedback!

trygveaa:
@posva: Hm, I tried `import { NavigationFailureType, isNavigationFailure } from 'vue-router'` in typescript, but it says they're not exported. Looking at the types, it seems that they're only static types on the `VueRouter` class. Is that correct?

posva:
Fixed at ecc8e27

Issue #235, opened 2015-11-15
Optional route params


envomer:
Is it possible to have optional route params?

Such as

```
router.map({
    '/foo/:bar?': { // bar is optional
        component: Foo
    },
    '/bar': {
        component: Bar
    }
})
```

yyx990803:
Just add a rule that maps `/foo` to component `Foo`. The question mark is used for query strings and therefore the syntax is confusing.

sodatea:
@yyx990803 
What if I want to have an optional parameter in a named route?

The following code does not work
```javascript
router.map({
    '/threads/:threadId/': {
        name: 'thread',
        component: ThreadView
    },
    '/threads/:threadId/:pageNum': {
        name: 'thread',
        component: ThreadView
    }
})
```

yyx990803:
@sodatea they are not the same route, why give them the same name?

sodatea:
@yyx990803 
Because the two routes *are equivalent* underlyingly, it's natural for them to share the same name.

An *optional* parameter means that the template markup needn't to care about what value to pass for the parameter. But if you have to define another route for that purpose, the template markup would end up taking care of the router name instead, which brings no benefit at all.

nullptr128:
I agree, stumbled upon this today as well. It would be great to have an optional parametr that will be just undefined if not set.

boovad:
Yeah.. I also needed and still need an option for an optional param. +1

scottbedard:
Optional params with the the ability to set default values would be really helpful

simplesmiler:
If the parameter is not required, then how about turning it into a query parameter instead?
`/threads/1/?page=2`

Saymon-biz:
There is many different situations where a parameter must be optional
e.g. not authenticated user tried to get access to the following url:
/administration/users
and we need to redirect him to login page with optional parameter "redirect-after-login" (e.g. /login/%20administration%20users) to redirect him to /administration/users after authentication.
But this route also should work without parameter, e.g. if user clicked directly by the link to the login page.
Right now I can solve this issue the following way:
```javascript
router.map({
  '/login': {
    name: 'login',
    component: Login
  },
  '/login/:redirect-after-login': {
    name: 'login-with-redirect',
    component: Login
  }
})

```
Seems I can't solve this without duplicating route definition.
An optional parameter looks better, isn't it?

simplesmiler:
@Saymon-biz this perfectly fits to `/login?redirect=<encoded_url>`.

[...]

PR #3229, opened 2020-06-12
Add information on optional URL parameters

### LINKED ISSUE #2525: Optional dynamic routing parameters in documentation
It aims to solve the lack of documentation on optional parameters when performing dynamic route matching. At the moment, the documentation for dynamic route matching does NOT show that you have the ability to make your params optional with the '?' character. Example:
```
    {
      path: '/blog:id?', // optional parameter. this feature is not shown in the documentation
      name: 'blog',
      component: Blog,
    },
```

https://router.vuejs.org/guide/essentials/dynamic-matching.html#reacting-to-params-changes is updated to show this feature

tnorthcutt:
Closes https://github.com/vuejs/vue-router/issues/2525

Related: https://github.com/vuejs/vue-router/issues/235

### DIFF
--- a/docs/guide/essentials/dynamic-matching.md
+++ b/docs/guide/essentials/dynamic-matching.md
@@ -38,6 +38,8 @@ You can have multiple dynamic segments in the same route, and they will map to c
 
 In addition to `$route.params`, the `$route` object also exposes other useful information such as `$route.query` (if there is a query in the URL), `$route.hash`, etc. You can check out the full details in the [API Reference](../../api/#the-route-object).
 
+A param can be made optional by appending a `?`, e.g. `/user/:username?`. See [Advanced Matching Patterns](#advanced-matching-patterns) for more information on matching patterns.
+
 ## Reacting to Params Changes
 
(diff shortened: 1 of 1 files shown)


posva:
Thank you for this but I think the documentation is already pointing to the right places at https://router.vuejs.org/guide/essentials/dynamic-matching.html#advanced-matching-patterns

This section of the documentation will be improved in v4 where the path to regex mechanism is internal

tnorthcutt:
I understand what you're saying. Respectfully, I think you're missing out on an opportunity to help a lot of people with next to know effort or cost.

It seems extremely likely to me that optional URL parameters are an extremely common use-case among users of vue-router. However, currently the fact that they are possible is buried at the bottom of the page, and is only mentioned in passing as "optional dynamic segments"; to learn how to use them one has to click through to documentation on a dependency and find the examples there.

Introducing the possibility far earlier on the page, right after dynamic segments are introduced, feels much more contextually appropriate, and much more likely to be communicating the right information at the time that a developer might be wondering "can I make these segments optional?".

I really appreciate all of your work on this project, and of course it's your decision to close this. I just wanted to share my reasoning, in the hope that you'll reconsider 😄

tnorthcutt:
Also, for what it's worth, as a new contributor it kind of stinks to come across what feels like some missing documentation, then find that there's an already open issue for it (https://github.com/vuejs/vue-router/issues/2525), and think to myself "oh great, I can help out and add this to the docs!", then make a PR and have it closed even though it clearly solves the existing open issue ¯\_(ツ)_/¯

PR #3243, opened 2020-06-23
Update base.js，don't need this abort.

ononee:
Don't need this abort.


posva:
That is intended. Please at least provide some context of what you are trying to fix when opening a Pull Request. Ideally a test case

PR #3246, opened 2020-06-26
Fix: TypeError: Cannot read property ‘path’ of undefined on Google webcache pages

docmattman:
This fixes an issue where the current route (`this.$route`) is undefined.  I haven't seen it in development or production, but I noticed that when the page is being loaded from Google webcache pages, the current route is undefined and causes an infinite loop, thus preventing the page from loading properly.  I have a post with findings here:

https://forum.vuejs.org/t/typeerror-cannot-read-property-path-of-undefined/98458


posva:
Thanks for the PR but the problem would be `this.$route` to be undefined. It shouldn't be. Can you provide a reproduction? It's even weirder than it happens with Google cache. If you find more information about where the `$route` is undefined, that would also help.

docmattman:
@posva, would it be helpful if I send an example of a Google webcache page where the problem occurs?  If so, is there a private place I can post it?  I'd prefer not to supply that in a public place.

posva:
as long as it's not a whole application and It's preferably a development build, you can send it over our discord chat (http://chat.vuejs.org)

docmattman:
@posva , the problem there is that this issue doesn't appear in development or on a production build.  It's unique in the fact that it only occurs on Google cached pages.  I could try to send some code, but it doesn't seem reproduce-able until Google crawls the page and caches it.  So I'm not sure how to proceed here.  Any suggestions for what you think would be most helpful to submit in this case?

posva:
You pretty much described how to reproduce it. So I would go with a boiled down application and try triggering the cache to showcase the error. Since the build doesn't matter, it would be better if the code isn't minified to debug better

docmattman:
@posva , I sent details to you via direct message on discord.  Let me know if you want/need it somewhere else instead.  Basically I found a way to recreate it without waiting for Google to crawl the site and cache a page.

posva:
Thanks for the efforts on this but I'm closing it because, as discussed, the `base` must be correct for the application to work. A warning wouldn't even help because it would disappear in production.

PR #3248, opened 2020-06-27
fix(types): add missing `options` property type

lbennett-stacki:
See [VueRouter.prototype.options](https://github.com/vuejs/vue-router/blob/dev/src/index.js#L27)


posva (approved):
Thanks!

PR #3251, opened 2020-07-01
feat(errors): capture errors thrown in redirect callback in onError

### LINKED ISSUE #3201: onError does not capture errors thrown in redirect callback
### Version
3.1.6

### Reproduction link
[https://codepen.io/trainiac/pen/rNOZvRO?editors=1112](https://codepen.io/trainiac/pen/rNOZvRO?editors=1112)

### Steps to reproduce
1. create route config like

```ts  
{
    path: '/oldpath/:part',
    redirect: (to: Route) => {
      if(to.ooopsmistake.part) {
        return `/newpath/${to.params.part}`
      }
      return '/newpath/'
    },
  }
```

### What is expected?
When router.push('/oldpath/test'). an error is throw becuase there is no oopsmistake property on the to route.  This should be caught in router.onError.

It's only caught in router.push('/oldpath/test').catch(err is here)

raulcabello:
Resolves  #3201 

This PR allows exceptions to be caught in router.onError.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -77,46 +77,55 @@ export class History {
     onComplete?: Function,
     onAbort?: Function
   ) {
-    const route = this.router.match(location, this.current)
-    this.confirmTransition(
-      route,
-      () => {
-        const prev = this.current
-        this.updateRoute(route)
-        onComplete && onComplete(route)
-        this.ensureURL()
-        this.router.afterHooks.forEach(hook => {
-          hook && hook(route, prev)
-        })
-
-        // fire ready cbs once
-        if (!this.ready) {
-          this.ready = true
-          this.readyCbs.forEach(cb => {
-            cb(route)
+    try {
+      const route = this.router.match(location, this.current)
+      this.confirmTransition(
+        route,
+        () => {
+          const prev = this.current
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Thanks but doing a try catch around the whole navigation is too much. Shouldn't try catching the router.match be enough?

raulcabello:
Hi @posva ! Yes you are right, try catching router.match should be enough. I just thought it might be a good idea to also catch errors in the confirmTransition. 
I will make the changes later, thanks for the feedback :)

src/history/base.js
```diff
@@ -77,46 +77,58 @@ export class History {
...
+      route = this.router.match(location, this.current)
+    } catch (e) {
+      this.errorCbs.forEach(cb => {
+        cb(e)
+      })
+      // Exception should still be thrown
+      // https://github.com/vuejs/vue-router/issues/3201
+      throw e
+    }
+    if (route) {
```
posva: since we throw again if we fail, I don't think we need the if

raulcabello: yes, I just removed it in a new commit

posva (approved):
Thanks!

PR #3252, opened 2020-07-07
fix: Adjust router.go to fire navigation guards

### LINKED ISSUE #3250: Abstract history mode doesn't fire `afterEach` hooks when using `router.go()`
### Version
3.3.4

### Reproduction link
[https://codepen.io/sirlancelot/pen/wvMPeZj?editors=1011](https://codepen.io/sirlancelot/pen/wvMPeZj?editors=1011)

### Steps to reproduce
1. Open Reproduction (Ensure CodePen's Console is open)
1. In the Output Window, click the "One" link
1. Observe that the `afterEach` hook fires ("Navigate / -> /one" is emitted in the Console)
1. In the Output Window, click the "Back" link

### What is expected?
* The `afterEach` hooks should fire when `router.go()` is used.
* "Navigate /one -> /" should be emitted in the Console.

`afterEach` hooks are not firing. Nothing is emitted in the Console.

sirlancelot:
Adjust router.go to fire navigation guards when using abstract routing mode.

Close #3250

### DIFF
--- a/src/history/abstract.js
+++ b/src/history/abstract.js
@@ -44,11 +44,10 @@ export class AbstractHistory extends History {
       return
     }
     const route = this.stack[targetIndex]
-    this.confirmTransition(
-      route,
+    this.transitionTo(
+      route.fullPath,
       () => {
         this.index = targetIndex
-        this.updateRoute(route)
       },
       err => {
         if (isRouterError(err, NavigationFailureType.duplicated)) {


docs/guide/advanced/navigation-guards.md
```diff
@@ -167,6 +167,7 @@ beforeRouteLeave (to, from, next) {
 7. Call `beforeRouteEnter` in activated components.
 8. Call global `beforeResolve` guards.
 9. Navigation confirmed.
+1. Call global `onReady` hooks (Only during initial navigation).
```
sirlancelot: Exact numbering is not required in markdown. Normally to make a numbered list I would just make everything `1. ` in order to avoid the delta showing all subsequent items in the list as changed. If the team is okay with that, I'll update the numbering accordingly.

posva (changes requested):
hmm, I don't think it's a good idea calling it inside the runQueue function. I think we should be consistent with how it was before https://github.com/vuejs/vue-router/pull/2292 and call it where `updateRoute` was called but after `ensureURL`

No need to update the documentation for this one.

We would also need a test. This one should work as a unit test

sirlancelot:
Thanks for the feedback @posva. I'll work on the requested the changes and add a unit test tonight (PDT).

posva:
@sirlancelot Hello! Let me know if you still want to work on this or if you don't mind me taking it

sirlancelot:
Hi @posva, apologies for the radio silence. I've actually discovered that the real issue has to do with abstract mode's `router.go()` logic. Other routing modes basically call out to `transitionTo` rather than `confirmTransition`. I'll update the pull request accordingly. I haven't made a unit test for it yet as I couldn't really find a good spot where this kind of behavior would be tested.

posva:
No worries at all! As you a unit test, since this is only about abstract mode, you can create a new file that instantiates the router in abstract mode, setup an after each, trigger a navigation and check it works.

posva:
Fixed in https://github.com/vuejs/vue-router/commit/4da7021a1900fc0f36f362dc41e8132597c3b162

PR #3253, opened 2020-07-09
added export on the interfaces (RouteConfigSingleView and RouteConfig…

alexsiilvaa:
…MultipleViews)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -13,5 +13,7 @@ export {
   Location,
   Route,
   NavigationGuard,
-  NavigationGuardNext
+  NavigationGuardNext,
+  RouteConfigSingleView,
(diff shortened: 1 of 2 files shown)


posva:
Hello, what is the use case for this?

alexsiilvaa:
Hello, I need the exposed interfaces to be able to extend them.

Example, before the last release I managed to:
```
export interface IRoute extends RouteConfig {
  // code
}
```

I suggest exposing the interfaces to be able to do:
```
export interface IRoute extends RouteConfigSingleView {
// code
}

```
With change I will no longer have the message:

> "An interface can only extend an object type or intersection of object types with statically known members"

posva:
What do you extend them for?

PR #3268, opened 2020-07-28
fix(create-mathcer): remove duplicated decodeURIComponent

### LINKED ISSUE #2725: Properly encode/decode the percentage character (% -> %25) across browsers and upon initial navigation

### LINKED ISSUE #3138: Double decoding of url params
### Version
3.1.6

### Reproduction link
[https://lyaskovec.github.io/tests/](https://lyaskovec.github.io/tests/)

### Steps to reproduce
There are two cases where you can see issue with double decoding of URL:
case 1:
- Go to the example page
- Click on link  /%25
- Click on link /home
- Try to use browsers navigation buttons

case2:
- Go to the example page
- Click on link  /%25
- reload the browser page

### What is expected?
It should show a correct page

I got "URIError: URI malformed" error

claudiolcastro:
So, checking this issue https://github.com/vuejs/vue-router/issues/2725
I found that removing the `decodeURIComponent` from the `matchRoute` kinda resolve the problem, because the param already comes decoded.

I don't know if this is the proper way to fix this issue, but I will give some examples below that worked for me:

**URL**: _/route-matching/params/%252520-working/bar_  

PS*: Check the `foo` param

**Without the fix**  :
FAIL-route-matching_params_%252520-working_bar

**After the change:**
WORK-route-matching_params_%252520-working_bar

This example was also broking:
**URL**: /route-matching/params/100%25-working/bar"

FAIL-route-matching_params_100%25-working_bar

Closes #2725 
Closes #3138

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -185,10 +185,9 @@ function matchRoute (
 
   for (let i = 1, len = m.length; i < len; ++i) {
     const key = regex.keys[i - 1]
-    const val = typeof m[i] === 'string' ? decodeURIComponent(m[i]) : m[i]
     if (key) {
       // Fix #1994: using * with props: true generates a param named 0
-      params[key.name || 'pathMatch'] = val
+      params[key.name || 'pathMatch'] = m[i]
     }
   }
 


posva (changes requested):
Thanks for this! Let's add a few e2e tests to check this behavior in hash-mode.js and basic.js. One for initial navigation and a different one for a router-link navigation

You can check existing tests in those files as well

mestevezdeister:
Any new about this issue? Does it definitely fix the problem? I've tried applying the commit in my repository and it works for me. Is it going to be merge in the 3.x.x release? I've checked v3.4.3 and does not include the fix, the decoding problem still occurs.

Thank you very much!

posva:
It's missing tests if anybody is up to contribute!

mestevezdeister:
I don't know how to; but if I can and you can give me some guidance or refer me to some documentation about how to contribute in testing I would be glad to contribute.

mestevezdeister:
I've made the same modification and add the mentioned e2e tests in #3323

PR #3287, opened 2020-08-08
Make sure global onError called on abortation. Example provided. (repo)

afwn90cj93201nixr2e1re:


afwn90cj93201nixr2e1re:
@posva
you really still don't undestand the issue on calling next(false)? It's not global hookable, it's impossible due this condtiion.
That's why im asking for implementing GLOBAL VUE ROUTER onAbort event

The first line gonna call onError, the second one not gonna call it, we can't hook anything  else, there's no global way for hooking **Before Router Enter** stoping on **FALSE**
```js
beforeRouteEnter(to, from, next){
   if(1){
          next(new Error('This gonna call onError global route Hook'); // This gonna call onError global route Hook
     return;
    } 
   if(1){
       next(false); // but that's not, this call only onAbort local route link.
       return;
    }
}
```
```js
// IT'S CALLED ONLY ON new Error instances, Not called on next(false)
router.onError((err)=>{
    console.log('Something returned ERROR inside beforeRouterEnter hook',err);
})
// GLOBAL NOT LOCAL, STILL NOT IMPLEMENTED
router.onAbort((err)=>{
    console.log('Something returned false inside beforeRouterEnter hook',err);
})
```

_Originally posted by @afwn90cj93201nixr2e1re in https://github.com/vuejs/vue-router/pull/3287#issuecomment-670969086_

afwn90cj93201nixr2e1re:
This PR shouldn't be merged, coz onError should be called only on **Errors** not on false too...
Reopen issue and make label **contribution welcomed** for implementing onAbort event.

This PR is only for you, coz you still don't understand the issue, and just reading few first lines without taking look to a code.

afwn90cj93201nixr2e1re:
изображение

posva:
Something like `onAbort` would need to go through the RFC process because Navigation Failures changed in v4: `afterEach` will trigger after each navigation. Porting back that behavior would be a breaking change and adding a new feature to remove it right away would be problematic. You can however wrap `router.push` to add any custom handling

afwn90cj93201nixr2e1re:
Thanks for answer, im **very happy**, now u get what im talking about...
I can't use router.push, coz i have 1000 components and i have links everywhere, i can't do it, just can't, you should understand that this is impossible. It's too heavy and shitcoding style.

https://codesandbox.io/s/vue-routing-example-pfhhb That's an example for you.

> onAbort would need to go through the RFC process because Navigation Failures changed in v4

I don't think so onError wasn't changed only Navigation Failures was refactored, it's don't have any realtion with global hooks, global hooks wasn't changed, we are just expanding it, i checked out both version and v4 and 2.8 everywhere it's can be added, onAbort should be called only on false returning, so, there's 1 line code chaning and onAbort defenitions in both version in 2.8 and in 4.x

Just place onAbort nearby onError.

Waiting for response.

afwn90cj93201nixr2e1re:
@posva 


Also you can look up to screenshot:
https://github.com/vuejs/vue-router/pull/3287#issuecomment-670969303

afwn90cj93201nixr2e1re:
Also can you create new Issue with explanation that we need onAbort? Coz you gonna lost this idea again, as in 2019 october...

afwn90cj93201nixr2e1re:
Also can you create new Issue with explanation that we need onAbort? Coz you gonna lost this idea again, as in 2019 october...


Also i sent this pr especially for 2.8 branch version, not for 4.

Adding new onAbort not really gonna hurt new logic from v4...
Just an extra call.
Tbh i really dont understand why v4 calling afterEach on beforeRouterEnter's next(false) calling too.

@posva

afwn90cj93201nixr2e1re:
https://github.com/vuejs/vue-router/issues/3027

posva:
3.x is different from 2.x 
There are no updates on 2.x anymore

We expect people to communicate in a more respectful way in Vue to keep things welcoming a for everybody

PR #3298, opened 2020-08-17
feat(history): Reset history.current when all apps are destroyed

### LINKED ISSUE #3294: Reset history.current on destroy app
### Version
3.3.4

### Reproduction link
[https://jsfiddle.net/NightNei/5bv7co0u/10/](https://jsfiddle.net/NightNei/5bv7co0u/10/)

### Steps to reproduce
**1)** Open "Reproduction link" -> OK - beforeResolve handler is run
**2)** Click "foo" btn -> OK - beforeResolve handler is run
**3)** Click "home" btn -> OK - beforeResolve handler is run
**4)** Click "destroy" btn -> OK - app destroyed
**5)** Click "start" btn -> FAIL - app started, but beforeResolve handler isn't run

### What is expected?
`beforeResolve` is always run on start app

`beforeResolve` isn't run on restarting app

---
I am a maintainer of "Isomorphic Layout Composer" - Micro-frontends library( https://github.com/namecheap/ilc ).  
And we have faced the problem of remounting Vue apps - when we render the same app second time, it will ignore running of `beforeHooks`.  

Since on restart app `history.current.matched` contains last instance of route component so this condition: https://github.com/vuejs/vue-router/blob/dfc289202703319cf7beb38d03c9258c806c4d62/src/history/base.js#L159 returns `true` and `router.beforeHooks` are ignored.
On Destroy app we should reset `history.current` to default value(`START`).

nightnei:
This resolves #3294

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -252,11 +252,17 @@ export class History {
     // Default implementation is empty
   }
 
-  teardownListeners () {
+  teardown () {
+    // clean up event listeners
+    // https://github.com/vuejs/vue-router/issues/2341
     this.listeners.forEach(cleanupListener => {
       cleanupListener()
     })
     this.listeners = []
+
+    // reset current history route
+    // https://github.com/vuejs/vue-router/issues/3294
+    this.current = START
   }
 }
 
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -252,17 +252,25 @@ export class History {
     // Default implementation is empty
   }
 
-  teardown () {
+  teardownListeners () {
     // clean up event listeners
     // https://github.com/vuejs/vue-router/issues/2341
     this.listeners.forEach(cleanupListener => {
       cleanupListener()
     })
     this.listeners = []
+  }
 
+  resetHistoryState () {
     // reset current history route
     // https://github.com/vuejs/vue-router/issues/3294
     this.current = START
+    this.pending = null
+  }
+
+  teardown () {
+    this.teardownListeners()
+    this.resetHistoryState()
   }
 }
 
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -252,25 +252,18 @@ export class History {
     // Default implementation is empty
   }
 
-  teardownListeners () {
+  teardown () {
     // clean up event listeners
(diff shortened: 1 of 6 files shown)


nightnei:
@posva @joeldenning look at the PR, please

src/history/base.js
```diff
@@ -252,11 +252,17 @@ export class History {
     // Default implementation is empty
   }
 
-  teardownListeners () {
+  teardown () {
```
internettrans, line 255: If we rename this method, it might make sense to rename the `setupListeners` method, too. https://github.com/vuejs/vue-router/blob/dfc289202703319cf7beb38d03c9258c806c4d62/src/history/base.js#L251

src/history/base.js
```diff
@@ -252,11 +252,17 @@ export class History {
...
+    // clean up event listeners
+    // https://github.com/vuejs/vue-router/issues/2341
     this.listeners.forEach(cleanupListener => {
       cleanupListener()
     })
     this.listeners = []
+
+    // reset current history route
+    // https://github.com/vuejs/vue-router/issues/3294
+    this.current = START
```
internettrans, line 265: Should we set `this.pending = null` here, too?

internettrans (commented):
Looks good, left a few comments.

src/history/base.js
nightnei: probably you are right, good point, thank you

src/history/base.js
```diff
@@ -253,11 +253,25 @@ export class History {
...
   }
+
+  resetHistoryState () {
+    // reset current history route
+    // https://github.com/vuejs/vue-router/issues/3294
+    this.current = START
+    this.pending = null
+  }
+
+  teardown () {
```
posva: I think we can just put everything inside here as I don't think it makes sense to call `teardownListeners` or `resetHistoryState` separately

posva (changes requested):

src/history/base.js
nightnei: @posva done

posva (approved):
Thank you!

nightnei:
> Thank you!

Thank you for your effort of supporting and developing vue :)

Issue #2458, opened 2018-10-30
next(false) doesn't render component when using browser navigation


JosNun:
### Version
3.0.1

### Reproduction link
[https://codesandbox.io/s/rjwjwqlzlm](https://codesandbox.io/s/rjwjwqlzlm)

### Steps to reproduce
- Use browser navigation to navigate to a page where the `beforeEnter()` method calls `next(false)`.
- Eg. When viewing [rjwjwqlzlm.codesandbox.io/](https://rjwjwqlzlm.codesandbox.io/) in a browser, paste [rjwjwqlzlm.codesandbox.io/](https://rjwjwqlzlm.codesandbox.io/) in the omnibox.

### What is expected?
Like it works when using the router-link navigation, one would expect that changing the URL from `somedomain.com/` to `somedomain.com/admin` (or other "protected" route) would return the user to the previous page.

### What is actually happening?
Instead of working properly, the URL is changed to what it is supposed to be, but no component is rendered at the initial route. 

---
A workaround is using `next(from.fullPath)` instead of using `next(false)`

posva:
The thing is, by entering a new URL, you are in a _"new context"_ with no access to previous URL. The workaround would not work if changing from `/foo` to `/admin`.
You can however check the `matched` array to see if it matched any route to properly redirect the user

BenceSzalai:
Possible workarounds, that work even when navigating from a path other than `/`:

Workaround 1:
Use hash mode instead of history for [navigation](https://router.vuejs.org/guide/essentials/history-mode.html#html5-history-mode).

Workaround 2:
You can continue using History mode, by implementing this:
Store the last successfully accessed route in sessionStorage using the [`afterEach`](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-after-hooks) feature, and read that from sessionStorage and use in the `next()` calls, instead of `false`.

PR #3313, opened 2020-09-06
docs: Added section to Caveats using history mode

BenceSzalai:
Calling `next(false)` may cause issue (blank page) when the user navigates to a different URL using the address bar, and it should be mentioned. Also added workaround with example.

Related issue: https://github.com/vuejs/vue-router/issues/2458

### DIFF
--- a/docs/guide/essentials/history-mode.md
+++ b/docs/guide/essentials/history-mode.md
@@ -124,7 +124,9 @@ Add this to your `firebase.json`:
 }
 ```
 
-## Caveat
+## Caveats
+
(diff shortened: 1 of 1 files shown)


posva:
Thank you for the PR but while I appreciate the effort put in it, I find this more confusing than helping and there also a few things are not true. The workaround proposed might be what one wants but not always

BenceSzalai:
Nevermind. At least it is written somewhere and one may find it when googling an issue.
Thanks for looking into it!

May I ask you please to let me know what is not true?

PR #3319, opened 2020-09-11
Allow objects to be passed in as params in Location

Tubbebubbe:
Objects can be passed on as parameters to `this.$router.push(Location)`. This PR prevents linter from emitting errors.


posva:
Thank you but this is not supported. Only params declared in path are supported (strings)

PR #3351, opened 2020-10-21
feat(scroll): dd behavior support on scrollBehavior

### LINKED ISSUE #2069: pass on extra arguments to `scrollTo` to enable smooth scrolling with behaviour

gaetansenn:
Add the possibility to pass behaviour type to window.scrollTo if supported by browser.

Close #2069

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -165,7 +165,8 @@ function scrollToPosition (shouldScroll, position) {
       window.scrollTo({
         left: position.x,
         top: position.y,
-        behavior: shouldScroll.behaviour || 'auto'
+        // $flow-disable-line
+        behavior: shouldScroll.behavior
       })
     } else {
       window.scrollTo(position.x, position.y)
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -160,6 +160,15 @@ function scrollToPosition (shouldScroll, position) {
   }
 
   if (position) {
-    window.scrollTo(position.x, position.y)
+    // $flow-disable-line
+    if ('scrollBehavior' in document.documentElement.style) {
+      window.scrollTo({
+        left: position.x,
+        top: position.y,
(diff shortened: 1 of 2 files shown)


src/util/scroll.js
```diff
@@ -160,6 +160,15 @@ function scrollToPosition (shouldScroll, position) {
   }
 
   if (position) {
-    window.scrollTo(position.x, position.y)
+    // $flow-disable-line
+    if ('scrollBehavior' in document.documentElement.style) {
+      window.scrollTo({
+        left: position.x,
+        top: position.y,
+        behavior: shouldScroll.behaviour || 'auto'
```
posva: we can pass undefined, so we should be able to just do:

```suggestion
        behavior: shouldScroll.behaviour
```

PR #3352, opened 2020-10-21
docs: fix Webpack code splitting link

samzeng:


posva (changes requested):
can you point to the right anchor as well, please?
I also noticed these are docs for webpack 5 but comments seems to be the same

samzeng:
> can you point to the right anchor as well, please?

The link here don't need to anchor.

You can check the same paragraph description in [Vue.JS document(Second paragraph)](https://cn.vuejs.org/v2/guide/components-dynamic-async.html#%E5%BC%82%E6%AD%A5%E7%BB%84%E4%BB%B6).

> 一个推荐的做法是将异步组件和 webpack 的 [code-splitting 功能一起配合使用](https://webpack.js.org/guides/code-splitting/)：

PR #3355, opened 2020-10-22
Update view.js

wprashed:


posva:
@safanbd Please don't open more PR with irrelevant changes, it's considered spam and if you keep doing it we will have to block you

PR #3358, opened 2020-10-25
docs: navigation-failures.md fix examples

Alex-Sokolov:
Because everywhere on the page used `router` for navigation, which need to be imported as `router`


posva:
Thank you but the code is correct. `VueRouter` is the class, `router` is the instance

PR #3370, opened 2020-11-12
feat(history): allow full base url

stephahart:
I'm working on a project that includes a <base> tag in the HTML so that all static assets are loaded through a CDN on a different domain. When you add a <base> tag, relative paths pushed to the history API use the href of the <base> instead of window.location.host.

In order to not direct the user to the CDN, I'd like to make a change to allow vue-router to accept full URLs. This only requires a minor change to two lines:
1. To ensure that a leading slash is not added before the protocol of a full URL in `normalizeBase`.
2. To keep double slashes after the protocol in `cleanPath`.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -274,7 +274,7 @@ function normalizeBase (base: ?string): string {
     }
   }
   // make sure there's the starting slash
-  if (base.charAt(0) !== '/') {
+  if (base.charAt(0) !== '/' && !base.startsWith('http')) {
     base = '/' + base
   }
   // remove trailing slash
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -70,5 +70,11 @@ export function parsePath (path: string): {
 }
 
 export function cleanPath (path: string): string {
-  return path.replace(/\/\//g, '/')
+  let protocol = ''
+  let p = path
+  if (p.startsWith('https://') || p.startsWith('http://')) {
+    protocol = p.startsWith('https://') ? 'https://' : 'http://'
+    p = p.replace(protocol, '')
+  }
+  return protocol + p.replace(/\/\//g, '/')
 }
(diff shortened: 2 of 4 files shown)


posva:
Thanks for the PR with tests but I don't think this is a good idea:

- Base history is not supposed to be directly instantiated for tests
- This assumes http only protocols

Do you have a boiled downn repro of the problem I can take a look at?

stephahart:
Hey @posva, thanks for the reply. To reproduce this problem, run any Vue app that uses vue-router. After the first page loads, open up dev tools and edit the head of the html to add a base element:
```
<html>
  <head>
    <base href="https://www.google.com">
    ...
  </head>
  ...
</html>
```
It doesn't matter what that href is as long as it's different than the current host. Then, click on any vue-router link, and you'll see that the domain of the link is what's set in that base element.

This could certainly be expected behavior for some if not most apps, so we would like to be able to provide a full URL to the router as a workaround.

posva:
The base tag is supposed to be present when the page renders, the router picks it up

stephahart:
Yep, the router picks it up as expected. Regardless of when the base element is added, though, I would like be able to continue to route on the current origin while our assets are loaded through the href defined in the base element. Right now this isn't possible because vue-router doesn't allow the origin in the 'base' configuration option.

posva:
I'm not sure I understand... I will need a proper reproduction with reproduction steps and expected vs actual behavior: https://new-issue.vuejs.org/?repo=vuejs/vue-router

stephahart:
Thanks, @posva. I created an issue here: https://github.com/vuejs/vue-router/issues/3372. Please let me know if creating a sample repo would make the issue clearer.

posva:
I understand now what you are trying to do but that is not possible (in HTML) as commented at https://github.com/vuejs/vue-router/issues/3372#issuecomment-727024198

PR #3384, opened 2020-11-24
Make sure scrollRestoration is writable.

afoncatti:
I'm receiving this error
TypeError · Cannot set property scrollRestoration of [object History] which has only a getter

with additional info:
node_modules/vue-router/dist/vue-router.esm.js:1694:39 setupScroll	
function setupScroll () {
  // Prevent browser scroll behavior on History popstate
  if ('scrollRestoration' in window.history) {
    window.history.scrollRestoration = 'manual';
  }

I propose to add this try-catch to verify if scrollRestoration is really writable before trying to set it.


afoncatti:
lint error

PR #3385, opened 2020-11-24
types: add RouteMeta interface to enable module augmentation

### LINKED ISSUE #3183: Update Route meta type to a user-extensible empty interfaces
Currently a route's meta field is listed as "any". This makes it impossible for users to define the structure of their own metadata. As an example, I wish to use the "meta" property for middleware. I've created a Middleware type and would like an interface like the following:

```
declare module "vue-router" {
    interface Route {
        meta: {
            middleware?: Middleware[]
        }
    }
}
```

If I try to define it this way, I get an error saying that all declarations of "meta" must have identical modifiers, and that "meta" was already defined in router.d.ts... However if I DON'T define it this way then I lose the ability to statically type check this property and catch developers who accidentally pass an incorrect type (e.g. passing a string containing the middleware's name, instead of the middleware object itself)

Now there's admittedly a limitation of this. Currently, it's possible to pass a non-object for the meta of a route. If you want to stick just a number or a string in there, it's completely valid. Making this a user-extensible interface would eliminate that use case (and therefore be a breaking API change) -- but I find it highly suspect that anyone is using the meta field in that way in a production app of any meaningful size. Without key-value pairs, storing only a single string into the "meta" field doesn't feel very useful.

```
export interface Route {
  path: string
  name?: string | null
  hash: string
  params: Dictionary<string>
  fullPath: string
  matched: RouteRecord[]
  redirectedFrom?: string
  meta?: {} // <-- replaced "any" with "{}"
}
```

HoldYourWaffle:
Close #3183

I added a `RouteMeta` interface to the type declarations to allow for stronger type safety using [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).

For example:
```ts
declare module 'vue-router/types/router' {
    interface RouteMeta {
        x: string;
    }
}
```
Would define `myRoute.meta.x` to be a `string`. Non-existent properties will show an error.

Unfortunately this is a (minor) breaking change. As far as I know this can't be avoided without sacrificing the added type-safety.
Users can manually revert to the original "`any`" behavior using an augmentation like this:
```ts
declare module 'vue-router/types/router' {
    interface RouteMeta {
        [key: string]: any;
    }
}
```
I can't make this the default however, because it can't be _overwritten_ using module augmentation.
In other words: the "`any`"-index type would stay "active", thus defeating the purpose of added type-safety.

I hope this suggestion can be considered nonetheless.

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -126,7 +126,7 @@ interface _RouteConfigBase {
   children?: RouteConfig[]
   redirect?: RedirectOption
   alias?: string | string[]
-  meta?: any
+  meta?: RouteMeta
   beforeEnter?: NavigationGuard
   caseSensitive?: boolean
   pathToRegexpOptions?: PathToRegexpOptions
@@ -153,7 +153,7 @@ export interface RouteRecord {
   parent?: RouteRecord
   redirect?: RedirectOption
   matchAs?: string
-  meta: any
+  meta: RouteMeta
   beforeEnter?: (
     route: Route,
     redirect: (location: RawLocation) => void,
@@ -205,5 +205,8 @@ export interface Route {
   fullPath: string
   matched: RouteRecord[]
   redirectedFrom?: string
-  meta?: any
+  meta?: RouteMeta
+}
+
+export interface RouteMeta {
 }
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -208,5 +208,4 @@ export interface Route {
   meta?: RouteMeta
 }
 
-export interface RouteMeta {
-}
+export interface RouteMeta extends Record<string | number | symbol, any> {}
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,6 +5,7 @@ export default VueRouter
 
 export {
   RouterMode,
+  RouteMeta,
   RawLocation,
   RedirectOption,
   RouterOptions,
--- a/types/test/index.ts
+++ b/types/test/index.ts
@@ -18,7 +18,9 @@ const Abc = { template: '<div>abc</div>' }
 const Async = () => Promise.resolve({ template: '<div>async</div>' })
 
 let err: any
-if (VueRouter.isNavigationFailure(err, VueRouter.NavigationFailureType.aborted)) {
+if (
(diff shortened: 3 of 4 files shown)


posva (changes requested):
Thanks, can you make like the existing one at https://github.com/vuejs/vue-router-next/pull/407 and add a test?

HoldYourWaffle:
@posva Although better than using plain `any`, I don't think having `RouteMeta` extend `Record<..., any>` is the best way here.
Like I mentioned in my initial comment, this kind of defeats the purpose of the added type safety.

It's better than nothing of course, since properties that have been declared will have their appropriate types.
However, invalid (undeclared) properties will not cause an error, meaning that things like typo's will go unnoticed.

Therefore I'd propose leaving this "Record extension" up to the user. Then anyone who wants to is free to have _their_ `RouteMeta` extend it, but it won't be _forced_ on all the others.

posva:
For Vue Router 3, that would be a breaking change, so we will have to keep `any`, but it can maybe be changed to unknown in v4 (vue-router-next)

HoldYourWaffle:
I suppose that's the only way then. I'll update this as soon as I get back to my workspace.
Should I create a PR to vue-router-next as well for changing `any` to `unknown`?

posva:
Sure! Let's see if existing tests pass and if the change is reasonable

types/router.d.ts
```diff
@@ -177,5 +177,8 @@ export interface Route {
   fullPath: string
   matched: RouteRecord[]
   redirectedFrom?: string
-  meta?: any
+  meta?: RouteMeta
+}
+
+export interface RouteMeta {
 }
```
```suggestion
export interface RouteMeta extends Record<string | number | symbol, any> {}
```

qwekelly:
@posva hello, The problem still exists on vue-router@4.0.4，can you solve it ? thanks!
and vue-router@4.0.4 vue-router.d.ts still exists ：
`export declare interface RouteMeta extends Record<string | number | symbol, unknown> {
}`

posva (changes requested):
The last thing that needs to be done here is to export `RouteMeta` from `index.d.ts` and add a test at https://github.com/vuejs/vue-router/tree/dev/types/test 🙂

PR #3390, opened 2020-11-26
fix(type): use any for Location.params

ylc395:
When passing props to route components, params are props, and props can be "any" type. For example:

in route config:
```js
const routes = [{
    name: 'a',
    path: '/a',
    props: true,
    component: A, // component A has a `aPropObject` prop 
}];
```

in component:
```js
this.$router.push({
    name: 'xxxx',
    params: {
        aPropObject: {}, // an object or anything else
    },
});
```


posva:
Thank you but that usage is not supported. The type is correct

PR #3433, opened 2020-12-28
fix(util/query): prevent overriding prototype propery names using querystring

anio:
Allowing prototype property names such as toString, \_\_proto__, ... to get override by querystrings with same names, leads to unknown behaviors.

Steps to reproduce the issue:

1. start the `dev` server.
2. go to the URL: http://localhost:8080/route-params/items/1/logs?toString

image

### DIFF
--- a/src/util/query.js
+++ b/src/util/query.js
@@ -58,10 +58,16 @@ function parseQuery (query: string): Dictionary<string> {
     return res
   }
 
+  const illegalNames = Object.getOwnPropertyNames(Object.getPrototypeOf(res))
+
   query.split('&').forEach(param => {
     const parts = param.replace(/\+/g, ' ').split('=')
     const key = decode(parts.shift())
     const val = parts.length > 0 ? decode(parts.join('=')) : null
+
+    if (illegalNames.includes(key)) {
+      return
+    }
 
     if (res[key] === undefined) {
       res[key] = val


posva:
given that `query` is transformed into a plain object, it's normal for it to disallow these keys. You can still override parsQuery to return a Map to allow any key.

As a developer, you would avoid using these keys and as a user, you don't manually invent keys on a query. If anything (developers and people who have learned it from developers), you change the values directly on the URL (e.g. a search). So what is the breaking use case you found?

anio:
> given that `query` is transformed into a plain object, it's normal for it to disallow these keys.

But what happens when someone tries to have methods these?

For example it deconstructs the result object when we send `__proto__`.
[http://localhost:8080/route-params/items/1/logs/info?_\_proto__&constructor](http://localhost:8080/route-params/items/1/logs/info?__proto__&constructor)

image

As you see `0` become an Object.

posva:
> As a developer, you would avoid using these keys and as a user, you don't manually invent keys on a query

My point is no user is going to write `__proto__` on the url unless they are a developer just checking things

Did you find any security concerns?

anio:
> My point is no user is going to write `__proto__` on the url unless they are a developer just checking things

Sure. But, does that mean it should not be fixed?

> Did you find any security concerns?

At this moment there is no security concern.

src/util/query.js
```diff
@@ -58,11 +58,17 @@ function parseQuery (query: string): Dictionary<string> {
     return res
   }
 
+  const illegalNames = Object.getOwnPropertyNames(Object.getPrototypeOf(res))
+
   query.split('&').forEach(param => {
     const parts = param.replace(/\+/g, ' ').split('=')
     const key = decode(parts.shift())
```
posva, line 65: let's do an `if (Object.prototype.hasOwnProperty(key)) return` here instead

posva (changes requested):

src/util/query.js
wxwzl: I do not appreciate this change. if it be fixed, toString will be a function rather than 2  in the url “https://vesaas.com/?toString=2" that is very strange.

wxwzl: I thought this need should be solved by developers when they new Router with RouteConfig which contains "parseQuery / stringifyQuery". as a common router, this should not be filtered;

posva: After more thoughts, I don't think we need to skip properties in the native parsing of queries. Users should be able to still provide their own implementation

PR #3455, opened 2021-01-24
feat: optimize method to determine whether the two path are the same …

### LINKED ISSUE #3447: Why is it designed as below? it will cause history.replaceState method to be called
### Version
3.4.9

### Reproduction link
[https://vesaas.com/?a=1&b=2&a=2](https://vesaas.com/?a=1&b=2&a=2)

### Steps to reproduce
input https://vesaas.com/?a=1&b=2&a=2 into webBroswer . then it becomes to https://vesaas.com/?a=1&a=2&b=2

### What is expected?
do not trigger history.replaceState called 

history.replaceState was called.

---

src/history/html5.js
`ensureURL (push?: boolean) {
    if (getLocation(this.base) !== this.current.fullPath) {
      const current = cleanPath(this.base   this.current.fullPath)
      push ? pushState(current) : replaceState(current)
    }
  }`

src/util/query.js
`function parseQuery (query: string): Dictionary<string> {
  const res = {}

  query = query.trim().replace(/^(\?|#|&)/, '')

  if (!query) {
    return res
  }

    const parts = param.replace(/\ /g, ' ').split('=')
    const val = parts.length > 0
      ? decode(parts.join('='))
      : null

    if (res[key] === undefined) {
      res[key] = val
    } else if (Array.isArray(res[key])) {
      res[key].push(val)
    } else {
      res[key] = [res[key], val]
    }
  })

  return res
}`

wxwzl:
Fix https://github.com/vuejs/vue-router/issues/3447 , optimize method to determine whether the two path are the same in ensureUrl methods

### DIFF
--- a/src/history/hash.js
+++ b/src/history/hash.js
@@ -3,6 +3,7 @@
 import type Router from '../index'
 import { History } from './base'
 import { cleanPath } from '../util/path'
+import { isSameRoute, createPlainRoute } from '../util/route'
 import { getLocation } from './html5'
 import { setupScroll, handleScroll } from '../util/scroll'
 import { pushState, replaceState, supportsPushState } from '../util/push-state'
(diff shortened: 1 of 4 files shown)


wxwzl:
make a new more completely fix

PR #3458, opened 2021-01-26
fix(warn): only warn if  "tag" or "event"  is used

### LINKED ISSUE #3457: v-router warning about using v-slot
### Version
3.5.0

### Reproduction link
[https://github.com/vuejs/vue-router/commit/d2cb951#diff-d69779a774d18aab6696dcf281454982a329104177c1a5f885089da9fefefd94R134-R148](https://github.com/vuejs/vue-router/commit/d2cb951#diff-d69779a774d18aab6696dcf281454982a329104177c1a5f885089da9fefefd94R134-R148)

### Steps to reproduce
See the conditions for firing the warning:

https://github.com/vuejs/vue-router/commit/d2cb951#diff-d69779a774d18aab6696dcf281454982a329104177c1a5f885089da9fefefd94R134-R148

### What is expected?
You should only get the warning about using v-slot if you are actually using `tag` or `event` props

Since `tag` and `event` have non-falsey default values, the warning will always fire with any use of router-link.

---
Perhaps you should just compare if the values for `tag` or `event` are non-default and fire the warning, even though this doesn't cover the edge case that `tag` or `event` are actually being passed in with a value that matches the default.

jonaskuske:
closes #3457
Only show deprecation warning if the props "tag" or "event" are actually passed by the consumer (key exists in `$options.propsData`, figured this out by experimenting 😅), not if they have any value at all (they always have a value since a default is specified in the props definition)

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -132,14 +132,14 @@ export default {
     }
 
     if (process.env.NODE_ENV !== 'production') {
-      if (this.tag && !warnedTagProp) {
+      if ('tag' in this.$options.propsData && !warnedTagProp) {
         warn(
           false,
           `<router-link>'s tag prop is deprecated and has been removed in Vue Router 4. Use the v-slot API to remove this warning: https://next.router.vuejs.org/guide/migration/#removal-of-event-and-tag-props-in-router-link.`
         )
         warnedTagProp = true
       }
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #3460, opened 2021-01-27
docs: add type to the code of some md files

FAKER-A:
Add type to the code of some md files

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -293,7 +293,7 @@ Since it's just a component, it works with `<transition>` and `<keep-alive>`. Wh
   type PositionDescriptor =
     { x: number, y: number } |
     { selector: string } |
-    ?{}
+    void
 
   type scrollBehaviorHandler = (
     to: Route,
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -289,7 +289,7 @@ Since it's just a component, it works with `<transition>` and `<keep-alive>`. Wh
 
   Signature:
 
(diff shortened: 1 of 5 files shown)


posva (changes requested):
Can you remove all the changes to `docs-gitbook`? those files should be removed

FAKER-A:
> Can you remove all the changes to `docs-gitbook`? those files should be removed

Ok Thanks Reply

docs/api/README.md
```diff
@@ -289,7 +289,7 @@ Since it's just a component, it works with `<transition>` and `<keep-alive>`. Wh
 
   Signature:
 
-  ```
+  ```ts
   type PositionDescriptor =
     { x: number, y: number } |
```
posva, line 294:
```suggestion
    { x?: number, y?: number } |
```

posva (changes requested):
The types were originally in flow, can you adapt them to TS by changing the `?{}` into `void` as well?

Thanks!

docs/api/README.md
FAKER-A: You mean this？
```ts
type PositionDescriptor =
    { x: number, y: number } |
    { selector: string } |
    void
```

posva: yeah, that should do!

FAKER-A: ok

posva (approved):
Thanks!

PR #3470, opened 2021-02-02
docs(link): add `custom` prop to samples

ashleahhill:
The `tag` prop of `<router-link>` has been removed in v4.0.0+

https://github.com/vuejs/rfcs/blob/master/active-rfcs/0021-router-link-scoped-slot.md

The `custom` prop is functional, however it is not mentioned anywhere in the docs.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -23,6 +23,7 @@ sidebar: auto
 ```html
 <router-link
   to="/about"
+  custom
   v-slot="{ href, route, navigate, isActive, isExactActive }"
 >
   <NavLink :active="isActive" :href="href" @click="navigate"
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -113,17 +115,36 @@ If you add a `target="_blank"` to your `a` element, you must omit the `@click="n
   <router-link :to="{ path: 'relative/path'}" append></router-link>
   ```
 
-### tag
+### tag (3.1.0 < 4.0.0)
```
posva: can you remove all changes after this one (including it). The docs of v4 are at https://next.router.vuejs.org/

posva (changes requested):

docs/api/README.md
ashleahhill: Well, I feel silly now. I didn't find that link anywhere yesterday :-P

posva (approved):
Thanks!

PR #3477, opened 2021-02-15
feat: optimize method to determine whether the two path are the same in ensureUrl methods

### LINKED ISSUE #3447: Why is it designed as below? it will cause history.replaceState method to be called

wxwzl:
Fix https://github.com/vuejs/vue-router/issues/3447 , optimize method to determine whether the two path are the same in ensureUrl methods

### DIFF
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -2,6 +2,7 @@
 
 import type VueRouter from '../index'
 import { stringifyQuery } from './query'
+import { normalizeLocation } from './location'
 
 const trailingSlashRE = /\/?$/
 
@@ -148,4 +149,9 @@ export function handleRouteEntered (route: Route) {
       }
     }
   }
+}
+/** just create a route without any added process */
+export function createPlainRoute (url: string): Route {
+  const location = normalizeLocation(url)
+  return createRoute(null, location)
 }
--- a/src/util/route.js
+++ b/src/util/route.js
@@ -97,8 +97,8 @@ export function isSameRoute (a: Route, b: ?Route): boolean {
 function isObjectEqual (a = {}, b = {}): boolean {
   // handle null value #1566
   if (!a || !b) return a === b
-  const aKeys = Object.keys(a).sort()
-  const bKeys = Object.keys(b).sort()
+  const aKeys = Object.keys(a)
+  const bKeys = Object.keys(b)
   if (aKeys.length !== bKeys.length) {
     return false
   }
--- a/examples/index.html
+++ b/examples/index.html
@@ -18,6 +18,8 @@
     <li><a href="route-props">Route Props</a></li>
     <li><a href="route-alias">Route Alias</a></li>
     <li><a href="route-params">Route Params</a></li>
+    <li><a href="route-query-hash">Route Query Hash</a></li>
+    <li><a href="route-query-history">Route Query History</a></li>
     <li><a href="router-errors">Router errors</a></li>
     <li><a href="transitions">Transitions</a></li>
     <li><a href="data-fetching">Data Fetching</a></li>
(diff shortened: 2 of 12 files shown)


src/util/route.js
```diff
@@ -96,8 +97,8 @@ export function isSameRoute (a: Route, b: ?Route): boolean {
 function isObjectEqual (a = {}, b = {}): boolean {
   // handle null value #1566
   if (!a || !b) return a === b
-  const aKeys = Object.keys(a).sort()
-  const bKeys = Object.keys(b).sort()
+  const aKeys = Object.keys(a)
+  const bKeys = Object.keys(b)
```
posva, line 101: Why should this be removed?

src/util/route.js
```diff
@@ -149,3 +150,8 @@ export function handleRouteEntered (route: Route) {
     }
   }
 }
+/** just create a route without any added process */
+export function createPlainRoute (url: string): Route {
```
posva, line 154: Ideally, this function (and its tests) should be removed since we just need to call `normalizeLocation()` on the url and compare it with `this.current`

posva (changes requested):
Can you add e2e tests for both hash and html5 history as well?

src/util/route.js
wxwzl: if that not be removed, they are the same object as below:
```
var a = {a:1,b:2}; var b = {b:2,a:1};

```

Object.keys method will keep the order of attributes in an object as the attributes  were  added.

if that not be removed,  "https://vesaas.com/?a=1&b=2" and  "https://vesaas.com/?b=2&a=1" will be thought the same route.
```
  ensureURL (push?: boolean) {
    const location = getHash()
    const plainRoute = createPlainRoute(location)
    if (!isSameRoute(plainRoute, this.current)) {
      const current = this.current.fullPath
      push ? pushHash(current) : replaceHash(current)
    }
  }
``` 
 the "ensureURL " method will not go into 'if ' block in this scene.

src/util/route.js
wxwzl: but how to compare with this current ,this.current is a Route object.I thought if it turn to  a  Route object , the logic can be keep smooth and complete.

wxwzl:
> Can you add e2e tests for both hash and html5 history as well?

I try my best to do it.

mmeester:
upvote 👍

mmeester (approved):

danielgroen (approved):

mmeester:
I've just tried the fix by @wxwzl but as far as I can determine this isn't fixing the issue that query params are sorted on init.

PR #3513, opened 2021-04-06
fix: change path with no slash in hash mode (fix #3512)

### LINKED ISSUE #3512: Change route path with no slash in hash mode，the corresponding component is not rendered
### Version
3.5.1

### Reproduction link
[https://codesandbox.io/s/vue-router-issue-example-u8xv5](https://codesandbox.io/s/vue-router-issue-example-u8xv5)

### Steps to reproduce
1. Create hash mode router，assuming that you currently have index path and login path，like this `http://localhost:8080/#/index` and `http://localhost:8080/#/login`

2. When i modify the index path to the login path in the browser address bar，like this `http://localhost:8080/#/index` 
 =>  `http://localhost:8080/#login`，**note：no slash after the `#`**

3. The corresponding component is not rendered

### What is expected?
Render the corresponding component

The corresponding component is not rendered

ZacharyL2:
fix: change path with no slash in hash mode

[#3512](https://github.com/vuejs/vue-router/issues/3512)

### DIFF
--- a/src/history/hash.js
+++ b/src/history/hash.js
@@ -14,7 +14,10 @@ export class HashHistory extends History {
     if (fallback && checkFallback(this.base)) {
       return
     }
-    ensureSlash()
+    const hash = getHash()
+    if (!hasSlash(hash)) {
+      replaceHash('/' + hash)
+    }
   }
 
   // this is delayed until the app mounts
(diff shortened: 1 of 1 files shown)


posva:
see https://github.com/vuejs/vue-router/issues/3512#issuecomment-814197511

PR #3519, opened 2021-04-12
fix: replace existing route in addRoute show warning msg

### LINKED ISSUE #3517: addRoute() with name couldn't replace the existing rule with the same name.
### Version
3.5.1

### Reproduction link
[https://github.com/klwfwdk/addroute-issue](https://github.com/klwfwdk/addroute-issue)

### Steps to reproduce
1.click to call router.addRoute() to add a  new router named 'About' witch has the same name attribute with a route rule already existing.
2.click router-link 'About'

### What is expected?
the route overwrite into new one. The browser should show 'This is an about page New'

Still appears as an unupdated component. The browser showed 'This is an about page'

qiuzuhui:
fix: addRoute is not work when there is already route with same name. replaced existing route in addRoute and show warning msg

### DIFF
--- a/src/create-route-map.js
+++ b/src/create-route-map.js
@@ -144,10 +144,8 @@ function addRouteRecord (
     })
   }
 
-  if (!pathMap[record.path]) {
-    pathList.push(record.path)
-    pathMap[record.path] = record
-  }
+  pathList.push(record.path)
+  pathMap[record.path] = record
 
   if (route.alias !== undefined) {
     const aliases = Array.isArray(route.alias) ? route.alias : [route.alias]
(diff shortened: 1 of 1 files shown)


hendrikheil:
Would love to see this merged. We currently have to use a fork to get the documented behavior, which is not ideal

Tested this and can confirm it works as expected :+1:

9-lives:
没有人 merge 吗？

src/create-route-map.js
```diff
@@ -146,8 +146,8 @@ function addRouteRecord (
 
   if (!pathMap[record.path]) {
     pathList.push(record.path)
-    pathMap[record.path] = record
   }
+  pathMap[record.path] = record
```
rwaltenberg, line 150: This is undocumented behavior. The documentation states that only if the route has a name and the name matches with an existing route, it should be overwritten.

rwaltenberg (changes requested):

772778995:
Niubitli！！

Cheaterman:
VueJS please fix

000Sonic:
+1

panruza:
+1

PR #3531, opened 2021-04-22
docs: api update example with router.START_LOCATION

Alex-Sokolov:

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -356,7 +356,7 @@ Since it's just a component, it works with `<transition>` and `<keep-alive>`. Wh
   })
 
   router.beforeEach((to, from) => {
-    if (from === START_LOCATION) {
+    if (from === router.START_LOCATION) {
       // initial navigation
     }
   })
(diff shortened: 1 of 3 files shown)


docs/api/README.md
```diff
@@ -356,7 +356,7 @@ Since it's just a component, it works with `<transition>` and `<keep-alive>`. Wh
   })
 
   router.beforeEach((to, from) => {
-    if (from === START_LOCATION) {
+    if (from === router.START_LOCATION) {
```
posva: It should be static property, so
```suggestion
    if (from === Router.START_LOCATION) {
```

Can you update zh and ru files as well in the docs folder?

posva (changes requested):

docs/api/README.md
Alex-Sokolov: Maybe also change `Router` to `VueRouter`? Because now we have `router` and `Router` which are confusing

Alex-Sokolov:
@posva Thanks for review! Maybe also change `Router` to `VueRouter`? Because now we have `router` and `Router` which are confusing

posva:
You are right, that should avoid some mistakes

Alex-Sokolov:
@posva changed too

posva (approved):
Thanks!

PR #3545, opened 2021-05-06
feat: 🎸 added `commit` to package.json

Blakelist7:
Now you can use `npm run commit` to add commit messages if you are
unfamiliar with commit rules


posva (approved):
Thanks

Blakelist7:
please let me know if there are some more issues that i can work on

PR #3554, opened 2021-05-19
fix(types): added missing router.match

MaxNvk:
Fixed the missing definition of a `match` method in `VueRouter` class


posva (approved):
Thanks!

PR #3556, opened 2021-05-20
fix(history): stricter check of base in HTML5

### LINKED ISSUE #3555: With routerBase being "/a/", landing on "/app" is incorrectly redirected to "/a/pp"
### Version
3.5.1

### Reproduction link
[https://jsfiddle.net/eyedean/4qcd9kgy/21/](https://jsfiddle.net/eyedean/4qcd9kgy/21/)

### Steps to reproduce
It's hard to reproduce in the jsfiddle and such.  That's what I mentioned in the `TODO` section of the link above.

**Actual reproduction**
* Let's say I have a Vue App with HTML5 History Router and RouterBase being `/a/`.
* If I land on `www.mydomain.com/bpp/`, VueRouter properly shows the App's 404 (catch on `/*/` in routes) page.
* However if I land on `www.mydomain.com/app`, then Vue Router redirects it to `www.mydomain.com/a/pp` which is not expected.  (Assume `/a/pp` is an existing page, and not the intended target of `/app`)

### What is expected?
show 404 on `/app`, just like `/bpp`.

redirects it to `/a/pp`, which can be a different existing page.

---
I tried to debug this on my own.  Here is what I found.

When I set VueRouterBase to `/a/` (with trailing slash, as advised in the documentation), `router.options.base` is properly `/a/`, but the `history.base` is [going through `normalizeBase`](https://github.com/vuejs/vue-router/blob/677f3c1f714fb61cc495345e535409b1cbb90429/src/history/base.js#L51) and becomes `/a`.  That's why `/app` gets sliced (via https://github.com/vuejs/vue-router/blob/677f3c1f714fb61cc495345e535409b1cbb90429/src/history/html5.js#L90-L92 ) and becomes `path: "/pp"`.

I also tried getting around this by setting routerbase to `/a//` (double trailing slash).  It stopped the `/app` => `/a/pp` redirect, successfully, but then all my links were ugly! (e.g. `<router-link to="/login">` would produce `www.mydomain.com/a//login` with double slash in between.)

eyedean:
Fixed a bug in getLocation of HTML5 History that would slice the `path`, even if it's just a string-prefix of the `base`. (e.g. for `base=/a/` and `path=/app` it would turn it to `/a/pp`.)

With the fix, I replaced the string-prefix check (via `path.indexOf(base) === 0`) into, an OR of these two conditions:
* Either `path === base` (case insensitive, as it was)
* Or, the `path` starts with `base` as a directory (i.e. `base` plus at least a `/` after that), so `path.indexOf(base+"/") === 0`

I tested it locally in my own repo that had this problem, and the fix is effective.  I also ran the `npm test` and it passed all the tests.  I didn't find any example unit test is already testing custom `base` to add to it.

Thanks.

Fix #3555


src/history/html5.js
```diff
@@ -87,7 +87,9 @@ export class HTML5History extends History {
 
 export function getLocation (base: string): string {
   let path = window.location.pathname
-  if (base && path.toLowerCase().indexOf(base.toLowerCase()) === 0) {
```
posva, line 90: Let's create two variables `pathLowerCase` and `baseLowerCase`

posva (changes requested):
Thanks!

src/history/html5.js
eyedean: @posva done

src/history/html5.js
```diff
@@ -87,7 +87,11 @@ export class HTML5History extends History {
...
+  const pathLowerCase = path.toLowerCase()
+  const baseLowerCase = base.toLowerCase()
+  // base="/a" shouldn't turn path="/app" into "/a/pp" https://github.com/vuejs/vue-router/issues/3555
```
posva:
```suggestion
  // base="/a" shouldn't turn path="/app" into "/a/pp"
  // https://github.com/vuejs/vue-router/issues/3555
  // so we ensure the trailing slash in the base
```

posva (approved):
Thanks!

PR #3558, opened 2021-05-29
docs: aria-current-value type `"true"` and `"false"`

yamanoku:
Since the type of https://router.vuejs.org/api/#aria-current-value is the same as the value that can be set for [`aria-current`](https://www.w3.org/TR/wai-aria-1.2/#aria-current) I thought that `"true"` and `"false"` should also be specified in the API documentation.

It would also be useful to make people aware that `aria-current-value="false"` can be overridden if `aria-current="page"` is unintentionally set.


posva (approved):
Thanks!

PR #3569, opened 2021-06-16
refactor(addRoute): create route only when parent.alias exists

alphawq:

### DIFF
--- a/src/create-matcher.js
+++ b/src/create-matcher.js
@@ -32,7 +32,7 @@ export function createMatcher (
     createRouteMap([route || parentOrRoute], pathList, pathMap, nameMap, parent)
 
     // add aliases of parent
-    if (parent) {
+    if (parent && parent.alias) {
       createRouteMap(
         // $flow-disable-line route is defined if parent is
         parent.alias.map(alias => ({ path: alias, children: [route] })),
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Can you provide a reproduction or a failing test (before the change)?

alphawq:
> Can you provide a reproduction or a failing test (before the change)?

It's not exactly a bug, but if parent.alias does not exist, there is no need to execute createRouteMap

src/create-matcher.js
```diff
@@ -32,7 +32,7 @@ export function createMatcher (
     createRouteMap([route || parentOrRoute], pathList, pathMap, nameMap, parent)
 
     // add aliases of parent
-    if (parent) {
+    if (parent && parent.alias) {
```
posva: So you meant

```suggestion
    if (parent && parent.alias.length) {
```

? Otherwise it would always execute anyway

alphawq: oh~ yes,  you are right! Thanks for your suggestion.

alphawq: oh~ yes,  you are right! Thanks for your suggestion.

posva (approved):

PR #3572, opened 2021-06-20
docs(ja): translate navigation-failures.md to japanese.

fu9da:


kazupon (approved):
Thank you for your translation!
Your translation is perfect! 💯

PR #3576, opened 2021-06-22
Update history-mode.md

weue:


posva:
Thank you but this is unnecessary

PR #3592, opened 2021-07-12
fix(scrollBehavior): trigger scroll behavior if same route with hash

### LINKED ISSUE #1668: Navigation to the same route fails (anchor, hash)
### Version
2.4.2

### Reproduction link
https://jsfiddle.net/ParachuteKadse/pL0sehq5/show/

### Steps to reproduce
1. Click on Foo
2. Click Foo-3
3. Scroll up
4. Repeat step 2

### What is expected?
Scroll down to Foo-3, like the first time.

Screen stays at the same position and doesn't scroll to Foo-3.

### Special notes
You can see the correct URL in the status-bar of your browser 
(Tested on Chrome)

ryanklarhoelter:
close #1668

This will make clicks on already active anchor links work.

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -20,6 +20,7 @@ import {
   isNavigationFailure,
   NavigationFailureType
 } from '../util/errors'
+import { handleScroll } from '../util/scroll'
 
 export class History {
   router: Router
@@ -161,6 +162,7 @@ export class History {
       route.matched[lastRouteIndex] === current.matched[lastCurrentIndex]
     ) {
       this.ensureURL()
+      handleScroll(this.router, current, route, false)
       return abort(createNavigationDuplicatedError(current, route))
     }
 
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -162,7 +162,9 @@ export class History {
       route.matched[lastRouteIndex] === current.matched[lastCurrentIndex]
     ) {
       this.ensureURL()
-      handleScroll(this.router, current, route, false)
+      if (route.hash) {
+        handleScroll(this.router, current, route, false)
+      }
       return abort(createNavigationDuplicatedError(current, route))
     }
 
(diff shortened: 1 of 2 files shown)


ryanklarhoelter:
Removing the whole "isSameRoute" condition would also do the trick (as @gspain found out: https://github.com/vuejs/vue-router/issues/1668#issuecomment-381252106). But I'm concerned that this would have some unwanted side effects that I'm not aware of.

ryanklarhoelter:
Because this would have an effect on all the current implementations I could also think of an option for the scroll behavior that developers could set to true—and only then the scroll behavior for same routes would be triggered.

posva (changes requested):
Thanks for this! Can you add an e2e test and make sure `handleScroll` is only called if there is a hash?

ryanklarhoelter:
Hi @posva! I guess we misunderstood a little. I've added a test now that checks if clicking the same anchor twice works (scrolls to the anchor location again). But: I haven't added a test that checks if "handleScroll()" is called for URLs with hashes only. I think it's useful to call "handleScroll()" whenever a link is clicked that doesn't change the URL. That will make multiple clicks on the same anchor work out of the box—plus it gives the developer the chance to implement an individual behavior (by customizing the scroll behavior) for clicking a link to the current site again (when there is no hash). They could decide for example to scroll to the top of the page or to reload the entire page (as it happens for simple HTML pages). What do you think?

posva:
It has to be hash only because otherwise this would break existing `scrollBehavior()` for users: right now clicking the link we are on does nothing but, in most cases (scroll to top), with the new changes, that would scroll to the top instead.

In Vue Router 4, scrollBehavior already works like you suggested.

ryanklarhoelter:
Done! 👍

posva (approved):
Thanks!

HenrijsS:
Judgement day is here. It's finally fixed!!!

paulolieuthier:
When can we expect this to be released?

PR #3608, opened 2021-07-30
Remove tautology

ScopeyNZ:
"trivial" and "easy" mean the same thing in this context, so it might as well just be simplified to "easy" to make it easier for everyone to understand.


posva (changes requested):
If we change this, we should make it closer to next docs: https://next.router.vuejs.org/guide/advanced/lazy-loading.html#lazy-loading-routes. The correct wording would be saying it supports it out of the box.

ScopeyNZ:
Thanks for the review @posva. I'm not planning on doing any major docs refactor, just took a few seconds to remove this extra word. I don't think it's worth conflating "reword the section" with "fix minor grammatical error".

Feel free to close the PR otherwise :)

posva:
Ok, thanks!

PR #3618, opened 2021-08-16
docs: note about router version

devzakir:


posva:
Thank you, I merged the two tips together to avoid having too much at the top of the docs.

PR #3626, opened 2021-09-05
 feat: add constructor hint

Hazlank:
Same with vue and vuex,add hint to tell user use new  `operator`

### DIFF
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -146,7 +146,9 @@ export class History {
             cb(err)
           })
         } else {
-          warn(false, 'uncaught error during route navigation:')
+          if (process.env.NODE_ENV !== 'production') {
+            warn(false, 'uncaught error during route navigation:')
+          }
           console.error(err)
         }
       }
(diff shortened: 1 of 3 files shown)


posva (changes requested):
Thanks! Can you make the dev warn, dev only, capitalize router and move the instance of check to the first param of the warn?

Hazlank:
@posva ```warn``` don't need to judge dev。In other places, it has been judged outside，except  ```base.js```

src/history/base.js
```diff
@@ -146,8 +146,10 @@ export class History {
             cb(err)
           })
         } else {
-          warn(false, 'uncaught error during route navigation:')
-          console.error(err)
+          if (process.env.NODE_ENV !== 'production') {
+            warn(false, 'uncaught error during route navigation:')
+            console.error(err)
```
posva: let's move the console.error outside of the if statement as it should be kept in production

posva (changes requested):

Hazlank:
@posva my bad,I changed it :joy:

posva (approved):

PR #3652, opened 2021-10-25
fix: clean more than two consecutive slashes

### LINKED ISSUE #2593: replaceState Error when path starts with // and scrollBehavior is set
### Version
3.0.2

### Reproduction link
[https://github.com/yegorLitvinov/vue-router-bug](https://github.com/yegorLitvinov/vue-router-bug)

### Steps to reproduce
Set scrollBehavior property in router config. Put two forward slashes into beginning of the url (e.g. http://localhost:8081//about)

### What is expected?
No errors in console

In firefox: SecurityError: The operation is insecure.
In Chrome: Uncaught DOMException: Failed to execute 'replaceState' on 'History': A history state object with URL 'http://about/' cannot be created in a document with origin 'http://localhost:8081' and URL 'http://localhost:8081//about'

danielroe:
**Issue**:
Currently a path beginning with three slashes (`///`) will in fact potentially redirect to another domain. This is because the `cleanPath` utility only replaces two slashes, leaving two behind.

**Reproduction**:
https://stackblitz.com/edit/vue-router-3-redirect

This PR simply replaces consecutive slashes with a single slash.

### DIFF
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -70,5 +70,5 @@ export function parsePath (path: string): {
 }
 
 export function cleanPath (path: string): string {
-  return path.replace(/\/\//g, '/')
+  return path.replace(/\/+/g, '/')
 }


posva (approved):
Technically this is not the correct fix as `cleanPath` is used in a variety of places, so it would change routes that have multiple slashes but at the same time, it's already turning paths like `//thing` into `/thing` so this simple fix shouldn't have any undesired side effects.

PR #3657, opened 2021-10-28
fix: 🐛 treat isActive to case-insensitive

### LINKED ISSUE #3656: case-insensitive on isActive property
### Version
3.5.3

### Reproduction link
[localhost/active-links/abouT](http://localhost:8080/active-links/abouT)

### Steps to reproduce
- Run `yarn run dev` in the root path of the project
- Go to http://localhost:8080/active-links/abouT

### What is expected?
The "/about" will be treated as active.

The "/about" is not treated as active.

---
Since Vue Router already supports case-insensitive matching for routes themselves. I think it should be the same for the active match to be case-insensitive too.

Kyle-Ye:
✅ Closes: #3656

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -72,10 +72,11 @@ export default {
       ? createRoute(null, normalizeLocation(route.redirectedFrom), null, router)
       : route
 
+    const ignoreCase = current.matched[0].regex.ignoreCase
     classes[exactActiveClass] = isSameRoute(current, compareTarget, this.exactPath)
     classes[activeClass] = this.exact || this.exactPath
       ? classes[exactActiveClass]
-      : isIncludedRoute(current, compareTarget)
+      : isIncludedRoute(current, compareTarget, ignoreCase)
 
     const ariaCurrentValue = classes[exactActiveClass] ? this.ariaCurrentValue : null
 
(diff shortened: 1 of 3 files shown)


Kyle-Ye:
Maybe we need to add some tests for this new behavior, any suggestion where to write the new test?

posva (changes requested):
This is not the correct patch as there is an option to make the path matching case sensitive and this should respect it.
This information should be available on the route records as `regex.ignoreCase` or something similar.
Can you also add a few unit tests for isIncludeRoute?

Kyle-Ye:
> This is not the correct patch as there is an option to make the path matching case sensitive and this should respect it.
> This information should be available on the route records as `regex.ignoreCase` or something similar.
> Can you also add a few unit tests for isIncludeRoute?


Add the test and use regex.ignoreCase property. Any suggestion on how to safely get ignoreCase properly?

src/components/link.js
```diff
@@ -72,10 +72,11 @@ export default {
       ? createRoute(null, normalizeLocation(route.redirectedFrom), null, router)
       : route
 
+    const ignoreCase = current.matched[0].regex.ignoreCase
```
Kyle-Ye, line 75: Any suggestion on how to safely get ignoreCase properly? @posva

Kyle-Ye:
Hello @posva. Any suggestion on this PR?

PR #3659, opened 2021-11-04
feat: export RouterLink RouterView in esm

xiaoxiangmoe:
* In order to align with vue-router v4, export `RouterLink`, `RouterView`, `START_LOCATION` and `isNavigationFailure` in esm. 
* add typing for `RouterLink` and `RouterView`


posva:
Thanks, unfortunately this would be a breaking change. It was proposed before but we had to wait til v4.

xiaoxiangmoe:
@posva it will not break commonjs users and  esm users. I create two separate entries: `index.js` and `index.esm.js`

PR #3688, opened 2021-12-28
fix(install): adjust this.$options.router judgment

jectychen:
The judgment "isDef(this.$options.router)" can be more rigorous.
Since in some cases, the router can be null, which will cause the execute error.
And We have faced this situation in  [@testing-library/vue/render.js](https://github.com/testing-library/vue-testing-library/blob/main/src/render.js) 
BTW, this change have been seen in the vue-router-next.

### DIFF
--- a/src/install.js
+++ b/src/install.js
@@ -20,7 +20,7 @@ export function install (Vue) {
 
   Vue.mixin({
     beforeCreate () {
-      if (isDef(this.$options.router)) {
+      if (this.$options.router) {
         this._routerRoot = this
         this._router = this.$options.router
         this._router.init(this)


posva:
Thanks! can you provide a simple reproduction of when this would fail? Or more context maybe? Why would the statement need to check for more than false values?

posva:

gaaamii:
I met same problem with testing-library/vue@5, then I could avoid this issue with below.

```javascript
render(MyComponent, { router: undefined }) // pass router option as undefined explicitly
```

I think this PR does not have to be merged because this problem occured in testing-library/vue@5 which is for EOL version of Vue, Vue@2.

But I comment for someone who uses same version as me. I would be happy if it is helpful to someone 😺

PR #3702, opened 2022-01-21
Docs: Add version labels to languages

### LINKED ISSUE #1532: Versions in docs
Every language in the docs show version 2.0 and I think it should be the last updated version. Right now if a user wants to learn about a new feature (let's say, implemented in v2.6) there is no way of knowing if the docs in the desired language are up to date

```bash

    2.0 - English (Up to v2.6)
    2.0 - French (Up to v2.5)
    2.0 - Japanese (Up to v2.4)
    2.0 - 中文 (Up to v2.3)
    2.0 - German (Up to v2.2)
    2.0 - Русский (Up to v2.1)
    2.0 - 한국어(Korean) (Up to v2.0)
    2.0 - Español (Up to v2.1)
    0.7 Docs
 
or

    2.6 - English
    2.5 - French
    2.4 - Japanese
    2.3 - 中文
    2.2 - German
    2.1 - Русский
    2.0 - 한국어(Korean)
    2.1 - Español
    0.7 Docs

```

burbelohugo:
Fixes #1532.

Adds version labels to the languages list so users can see which version of the documentation certain language versions are updated to.

Demo:

Screen Shot 2022-01-21 at 5 23 16 PM

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -85,7 +85,7 @@ module.exports = ctx => ({
     smoothScroll: true,
     locales: {
       '/': {
-        label: 'English',
+        label: '2.6 - English',
         selectText: 'Languages',
         editLinkText: 'Edit this page on GitHub',
         nav: [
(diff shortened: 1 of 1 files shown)


posva:
Thanks but these are not Vue Router versions 😅  They seem to be Vue versions

PR #3707, opened 2022-01-30
WIP: Optimized matcher

iurisilvio:
This is a draft implementation with some new ideas.

* Static paths don't need regex match for most cases.
* Match don't have to be always linear, it is possible to match only with static paths.

Regex match is 10x slower than simple object access. I still have to run some dynamic matches to not break ordering compatibility.

I still want to integrate it better with current code, I did this implementation to validate the issue and run some benchmarks.

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -77,6 +77,7 @@ declare type RouteRecord = {
   beforeEnter: ?NavigationGuard;
   meta: any;
   props: boolean | Object | Function | Dictionary<boolean | Object | Function>;
+  pathToRegexpOptions: PathToRegexpOptions;
 }
 
 declare type Location = {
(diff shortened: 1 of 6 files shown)


iurisilvio:
I did a small benchmark script to prove the concept, in my own machine.

https://gist.github.com/iurisilvio/a916ce2de16336ec97aab8db5525f7f9

For 100 static routes, current vue-router takes 14ms average to resolve and my version takes less than 1ms.

Even to resolve the first result, it is 30% slower because it evaluates one regex.

iurisilvio:
For dynamic paths, it will evaluate only part of the routes, based on first level path. If you have 100 rules but only 5 starting with the same first level path, it will evaluate at most 5 rules.

PR #3710, opened 2022-01-30
Update scroll-behavior.md

milad-alizadeh:
It seems like the documentation is out of date. Scroll behaviour simply does not work with `x` and `y`. When `savedPostion` is returned, it returns `{ left: value, top: value}`. I updated my code to that and it works perfectly. Other people also seem to have this issue here. https://stackoverflow.com/questions/54535838/scroll-behaviour-vuejs-not-working-properly
Also, typescript keeps shouting about the wrong type


milad-alizadeh:
My bad I was looking at the wrong docs

PR #3712, opened 2022-02-05
test: Set NODE_ENV before error handling tests.

iurisilvio:
Error handling tests fail depending on other tests ordering, because `process.env.NODE_ENV` can be `production` and this test file don't set `NODE_ENV`.

Fix #3712.


posva (approved):
Thanks!

Issue #3709, opened 2022-01-30
Matcher is linear and uses too many regex


iurisilvio:
### What problem does this feature solve?
The matcher implementation always use path-to-regex lib. It is too much for static paths, where it is possible to just compare strings.

Also, the implementatioin is O(N), if you have 100 routes and want to match the last one (that is static), you run 100 regex tests.
Regex is around ~10x slower than a simple string match and with a lookup table it is O(1).

https://github.com/vuejs/vue-router/blob/5d7afbe25f6b212bd4adb8344da72f6227a5c6d2/src/create-matcher.js#L85-L89

### What does the proposed API look like?
The API don't change, it is just optimization.

I did a draft PR #3707 to show the idea.

posva:
This is really nice but it's not worth the increased complexity and library size.

v4 (https://github.com/vuejs/router) uses a custom implementation so maybe you want to get a look in that one but the final size should be kept very similar.

The complexity O(n) is necessary to cater to some existing use cases and there are plans to provide a way to use a custom matcher with different strategies that would allow lighter and faster routers but less features

iurisilvio:
I understand the problem with increased complexity and library size.

I wrote the PR just as a proof of concept, but I think it can be simpler and smaller integrating to create-route-map. Do you consider the change if I do that?

The O(n) complexity is necessary in some cases, but not for static paths.

My nuxt app has ~100 routes and 70% are static, so I have a huge overhead and I think it is a common use case.

iurisilvio:
I did a simple static matcher implementation in #3713, matching with static routes before full route matching. I tried to do a minor change to improve matcher for static paths.

Another easy optimization would be to not test against static paths in current match loop, just marking some routes as static.

Would you consider this approach?

Also, I'll check the v4 implementation to optimize them if it is possible.

PR #3713, opened 2022-02-05
Static matcher

iurisilvio:
Improves #3709.

### DIFF
--- a/test/unit/specs/create-map.spec.js
+++ b/test/unit/specs/create-map.spec.js
@@ -61,9 +61,7 @@ describe('Creating Route Map', function () {
       '',
       '/foo',
       '/bar/',
-      '/bar',
       '/bar-redirect/',
-      '/bar-redirect',
       '*'
     ])
   })
--- a/test/unit/specs/create-map.spec.js
+++ b/test/unit/specs/create-map.spec.js
@@ -19,6 +19,7 @@ const routes = [
     children: [
       {
         path: '',
+        pathToRegexpOptions: { strict: true },
         component: Baz,
         name: 'bar.baz'
       }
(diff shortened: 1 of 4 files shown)


iurisilvio:
The failing test is #3714.

test/unit/specs/create-map.spec.js
```diff
@@ -61,9 +61,7 @@ describe('Creating Route Map', function () {
       '',
       '/foo',
       '/bar/',
-      '/bar',
```
posva, line 64: These should still work. Were they removed by accident?

iurisilvio: I removed because the previous rule always match before, the strict flag is false by default.

I was not convinced about this behavior, so I wrote a test to validate it.

https://github.com/buserbrasil/vue-router/commit/c247c520d121b4847d0e9d88db5368885e5bd931

I think it is an implementation detail and don't change any behavior.

It is possible to maintain them in the list, but I have to think how to do it without increasing the size.

iurisilvio: My change remove static rules that will never match because you have another rule matching before. This way I can trust if it exists in path map, it should be used instead of running through the matching loop.

posva: yeah, these need to be added back, allowing a trailing slash when strict is false is necessary. If you think this is too much, that's fine, it's better to drop these changes on v3

iurisilvio: I changed the test to assert that the parent route is not removed from the list when child route is strict. It don't change matching behavior.

`/bar` was not removed because `/bar/` is strict, but `/bar-redirect` will never be reached, so it was removed.

I'm ok if it is not merged and I'll work on v3 soon to avoid O(n) matches.

PR #3733, opened 2022-03-02
feat: export RouterLink RouterView

xiaoxiangmoe:

### DIFF
--- a/types/Route.d.ts
+++ b/types/Route.d.ts
@@ -1,4 +1,4 @@
-import Vue, { ComponentOptions, PluginFunction, AsyncComponent } from 'vue'
+import Vue, { AsyncComponent, ComponentOptions } from 'vue'
 
 type Component = ComponentOptions<Vue> | typeof Vue | AsyncComponent
 type Dictionary<T> = { [key: string]: T }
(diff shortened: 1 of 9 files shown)


types/router.d.ts
```diff
@@ -1,4 +1,135 @@
-import Vue, { ComponentOptions, PluginFunction, AsyncComponent } from 'vue'
+import Vue, { VNode, ComponentOptions, PluginFunction, AsyncComponent } from 'vue' 
+
+export interface RouterLinkProps {
```
posva: can you move the type declarations to `RouterLink.d.ts` and `RouterView.d.ts`?

posva (changes requested):

types/router.d.ts
xiaoxiangmoe: done

posva (approved):

types/VueRouter.d.ts
```diff
@@ -0,0 +1,90 @@
+import Vue, { AsyncComponent, ComponentOptions, PluginFunction } from 'vue'
```
posva, line 1: can you put this back where it was?

types/Route.d.ts
```diff
@@ -1,4 +1,4 @@
-import Vue, { ComponentOptions, PluginFunction, AsyncComponent } from 'vue'
```
posva, line 1: can you keep the original filename?

posva (changes requested):

types/VueRouter.d.ts
xiaoxiangmoe: I put it in a separate file because I want to unravel the circular dependency.

VueRouter depends on RouterView and RouterLink
RouterLink depends on NavigationFailure and Route

But `VueRouter` and `NavigationFailure` and `Route` are defined in`types/router.d.ts`.

Should we allow circular dependency here?

PR #3746, opened 2022-05-10
Update history-mode.md

SpringHgui:


posva:
Thanks, but we only add configs, we cannot add a link to a package we do not control for security reasons.

posva:
Thanks, but we only add configs, we cannot add a link to a package we do not control for security reasons.

PR #3762, opened 2022-07-15
feat(composition-api): add composition-api

EveChee:
add useRoute & useRouter


posva:
Thanks, let's keep working on this on https://github.com/vuejs/vue-router/pull/3763 as it's closer the intended direction

PR #3770, opened 2022-08-20
feat: add vue 2.7 support 

xiaoxiangmoe:


types/router.d.ts
```diff
@@ -1,9 +1,14 @@
-import Vue, { ComponentOptions, PluginFunction, AsyncComponent } from 'vue'
+import Vue, { ComponentOptions, PluginFunction, AsyncComponent, VNode } from 'vue'
 
-type Component = ComponentOptions<Vue> | typeof Vue | AsyncComponent
+type Component = ComponentOptions<Vue> | typeof Vue | AsyncComponent | {}
```
posva, line 3: let's not add `{}`, it's a special type in TS

types/router.d.ts
```diff
@@ -1,9 +1,14 @@
...
 type Dictionary<T> = { [key: string]: T }
 type ErrorHandler = (err: Error) => void
 
+/**
+ * * `"hash"`: uses the URL hash for routing. Works in all Vue-supported browsers, including those that do not support HTML5 History API.
+ * * `"history"`: requires HTML5 History API and server config. See HTML5 History Mode.
+ * * `"abstract"`: works in all JavaScript environments, e.g. server-side with Node.js. **The router will automatically be forced into this mode if no browser API is present.**
```
posva, line 10:
```suggestion
 * - `"hash"`: uses the URL hash for routing. Works in all Vue-supported browsers, including those that do not support HTML5 History API.
 * - `"history"`: requires HTML5 History API and server config. See HTML5 History Mode.
 * - `"abstract"`: works in all JavaScript environments, e.g. server-side with Node.js. **The router will automatically be forced into this mode if no browser API is present.**
```

types/router.d.ts
```diff
@@ -17,42 +22,181 @@ export type NavigationGuard<V extends Vue = Vue> = (
...
+   * Original options object passed to create the Router
+   */
   options: RouterOptions
+  /**
+   * Configure the router mode.
+   * 
+   * default: `"hash"` (in browser) | `"abstract"` (in Node.js)
+   * 
+   * available values: `"hash" | "history" | "abstract"`
+   * * `"hash"`: uses the URL hash for routing. Works in all Vue-supported browsers, including those that do not support HTML5 History API.
```
posva, line 42: let's remove this list, it's redundant with the docs above

types/router.d.ts
```diff
@@ -61,50 +205,122 @@ export declare class VueRouter {
...
+ */
 export interface RouterOptions {
   routes?: RouteConfig[]
+  /**
+   * Configure the router mode.
+   * 
+   * default: `"hash"` (in browser) | `"abstract"` (in Node.js)
+   * 
+   * available values: `"hash" | "history" | "abstract"`
+   * * `"hash"`: uses the URL hash for routing. Works in all Vue-supported browsers, including those that do not support HTML5 History API.
```
posva, line 296: let's not duplicate the list docs

posva (changes requested):

posva:
I couldn't push so I merged this manually.

src/composables/index.js
```diff
@@ -112,3 +112,9 @@ function throwNoCurrentInstance (method) {
     `[vue-router]: Missing current instance. ${method}() must be called inside <script setup> or setup().`
   )
 }
+
+export { default as default } from '../index'
+export { default as RouterLink } from '../components/link'
+export { default as RouterView } from '../components/view'
+export { NavigationFailureType, isNavigationFailure } from '../util/errors'
+export { START as START_LOCATION } from '../util/route'
```
posva, line 120: I removed these exports after merging as they exist in vue-router

PR #3792, opened 2022-10-07
docs: removed incorrect description for addRoute method

martell2010:


posva:
Thanks but this is correct

Issue #3819, opened 2022-12-30
`router.resolve` did not handle `<base>` correctly


ichaoX:
### Version
3.6.5

### Reproduction link
[5s3iuf.csb.app/root/main.html](https://5s3iuf.csb.app/root/main.html#/foo)

### Steps to reproduce
Click on Link A or open Link B in a new tab.

And check the generated href attribute in the inspector.

### What is expected?
In hash mode, the following results are logical, I would prefer the last one, because the `HashHistory` is not related to the `baseURI` and the availability of the `<base>` source is unknown.

```javascript
// "https://example.com/root/rel/path#/bar"
new URL("#/bar",document.baseURI).href

// "https://example.com/root/main.html#/bar"
new URL("#/bar",location).href
// If location.href and document.baseURI are same-origin or both paths are the same, 
// it is possible to slice the generated URL to shorten it.
```

### What is actually happening?
The generated href attributes are all `"/./rel/path#/foo/#/bar"`.

Both the relative path and the hash of Link A are not handled correctly.

Link B has the same issue when opened in a new tab.

posva:
A base should likely not include a hash in it. Note that it changes the browser behavior of `<a>` and other things (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base). For instance, when you are opening in a new tab, that's the browser, vue-router simply lets that event happen

ichaoX:
> A base should likely not include a hash in it. [citation needed]

I have tested `document.baseURI` and it will contain the hash value.

Another main problem is that `router.resolve` treats it as an absolute path, while `<base>` is a relative path.

It's less reliable here:

https://github.com/vuejs/vue-router/blob/492bf0eaf1db617314cdc5c84e584c213dd2fb81/src/history/base.js#L270-L288

The hash mode should not join the '/' in the middle, the server may return 404 Not Found.

https://github.com/vuejs/vue-router/blob/492bf0eaf1db617314cdc5c84e584c213dd2fb81/src/router.js#L280-L283

To solve the problem of inconsistent opening in new tabs in hash mode, it can either do nothing special for `<base>`, or use `new URL("#/bar",location).href` to generate the URL.

ichaoX:
Hello, do I need to open a new issue?

Issue #2865, opened 2019-08-01
Make automatically added trailing slash in base optional


tychenjiajun:
### What problem does this feature solve?
We use the following base config
```js
base: /page/index.html
```

what we want in href generated by <router-link>

```html
<a href="/page/index.html#/foo"/></a>
```

what we have now
```html
<a href="/page/index.html/#/foo"/></a>
```

### What does the proposed API look like?
```js
base: {
    url: '/page/index.html'
    trailingSlash: false
}
```

posva:
What is the use case for that?

tychenjiajun:
@posva Our backend only allow us to use index.html to be the only entry to the SPA.

posva:
Initial navigation should still work as the router will still recognize `index.html#/users/:id`. But, the base should be `/page/` without the index.html, that way the server can probably also answer correctly

tychenjiajun:
@posva No, the server response 404 on `/page/` and fail to answer the CSS and js files if request URL is `/page/index.html/#/`. I was told to only use `/page/index.html` to be the only entry.

`<router-link>` works well when clicked. For example, `<router-link to="/foo">` can navigate us from `/page/index.html#/` to  `/page/index.html#/foo`, but its rendered href is `<a href="/page/index.html/#/foo">`, which makes us unable to right click it and open in a new tab.

And here is an example: https://codesandbox.io/s/vue-template-v7qmf

posva:
I forgot about this but with Hash mode, there should be no `base` specified. It's picked up automatically from the URL

PR #3823, opened 2023-01-11
fix(history): Resolve the href in `<base>` correctly (#3819)

ichaoX:
* Resolve the href in `<base>` correctly.

* In hash mode, the hash in base is automatically removed and the base trailing slash is distinguished.

* By default the result of `router.resolve().href` is the same as the actual switched URL.

Relates to: #3819, #2865

### DIFF
--- a/examples/hash-mode/app.js
+++ b/examples/hash-mode/app.js
@@ -40,7 +40,7 @@ const Query = { template: '<div>query: "{{ $route.params.q }}"</div>' }
 // 3. Create the router
 const router = new VueRouter({
   mode: 'hash',
-  base: __dirname,
+  base: require('path').join(__dirname, '/'),
   routes: [
     { path: '/', component: Home }, // all paths are defined without the hash.
     { path: '/foo', component: Foo },
--- a/src/history/base.js
+++ b/src/history/base.js
@@ -272,19 +272,27 @@ function normalizeBase (base: ?string): string {
     if (inBrowser) {
       // respect <base> tag
       const baseEl = document.querySelector('base')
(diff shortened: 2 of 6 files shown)


posva:
Thanks for the PR. Can you provide a failing test or boiled-down repro? Note a base should not contain a hash

ichaoX:
An example of reproduction can be found in #3819, and you can view the source code of the Reproduction link.

Some browser extensions or proxies may modify `<base>`, and as a standard feature of html, it is necessary to support it correctly.

This PR should implement support for almost all legal baseURIs.
The `<base>` tag is then respected after ensuring the prerequisites of availability and consistency .
Support `base.href` containing relative URI or hash, even cross-origin cases.

posva:
That base has a hash though. As I said, it cannot have a hash

ichaoX:
That reproduction link contains several mistakes at the same time:

1. incorrect relative path
2. incorrect hash
3. there is an extra slash between base and hash

Additional reproduction links:

https://5s3iuf.csb.app/root/path.html

https://5s3iuf.csb.app/root/sameorigin.html

ichaoX:
@posva 
Do you need any other repro?

examples/hash-mode/app.js
```diff
@@ -40,7 +40,7 @@ const Query = { template: '<div>query: "{{ $route.params.q }}"</div>' }
 // 3. Create the router
 const router = new VueRouter({
   mode: 'hash',
-  base: __dirname,
+  base: require('path').join(__dirname, '/#hash'),
```
posva: The problem is this is not an expected scenario: a base cannot have a hash

posva (changes requested):

PR #3875, opened 2023-11-16
fix: onBeforeRouteLeave not remove (fix #3826)

### LINKED ISSUE #3826: "onBeforeRouteLeave" guard is not removed after switching to another view when using keep-alive and composables api
### Version
3.6.5

### Reproduction link
[github.com](https://github.com/zengguirong/vue27-router-composition-bug)

### Steps to reproduce
1. Clone the Github repo
2. Run 'npm install'
3. Run 'npm run dev' and open the broswer
4. Click the first link
5. Click the second link
6. Click the first link one more time

### What is expected?
The window alert "onBeforeRouteLeave" only once after step 5

1. The window alert "onBeforeRouteLeave" after step 5
2. The window alert "onBeforeRouteLeave" after step 6

---
This bug only appears when using keep-alive and composables api.

wynn719:
The problem seems to be that `onBeforeRouteLeave` lacks handling of `onActivated` and `onDeactivated` hooks in the keep-alive component.

I referred to vue router 4 to fix the processing of `onBeforeRouteLeave`.

https://github.com/vuejs/router/blob/main/packages/router/src/navigationGuards.ts

Fix #3826

### DIFF
--- a/src/composables/guards.js
+++ b/src/composables/guards.js
@@ -35,9 +35,9 @@ export function onBeforeRouteLeave (guard) {
 }
 
 function registerGuard (router, guard, fn, depth) {
-  return router.beforeEach((to, from, next) => {
-    return fn(to, from, depth) ? guard(to, from, next) : next()
-  })
+  return router.beforeEach((to, from, next) =>
+    fn(to, from, depth) ? guard(to, from, next) : next()
+  )
 }
 
 const noop = () => {}
@@ -70,7 +70,7 @@ function useFilteredGuard (guard, fn) {
     })
     onDeactivated(() => {
       removeGuard()
-      removeGuard = null // reset removeGuard
+      removeGuard = null
     })
 
     return removeGuard
--- a/src/composables/guards.js
+++ b/src/composables/guards.js
@@ -34,6 +34,12 @@ export function onBeforeRouteLeave (guard) {
   return useFilteredGuard(guard, isLeaveNavigation)
 }
 
+function registerGuard (router, guard, fn, depth) {
+  return router.beforeEach((to, from, next) => {
+    return fn(to, from, depth) ? guard(to, from, next) : next()
+  })
+}
+
 const noop = () => {}
 function useFilteredGuard (guard, fn) {
   const instance = getCurrentInstance()
@@ -56,17 +62,11 @@ function useFilteredGuard (guard, fn) {
       : null
 
   if (depth != null) {
-    const registerGuard = () => {
-      return router.beforeEach((to, from, next) => {
-        return fn(to, from, depth) ? guard(to, from, next) : next()
-      })
-    }
-
-    let removeGuard = registerGuard()
+    let removeGuard = registerGuard(router, guard, fn, depth)
     onUnmounted(removeGuard)
 
     onActivated(() => {
-      removeGuard = removeGuard || registerGuard()
+      removeGuard = removeGuard || registerGuard(router, guard, fn, depth)
     })
     onDeactivated(() => {
       removeGuard()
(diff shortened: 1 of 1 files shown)


src/composables/guards.js
```diff
@@ -56,11 +56,23 @@ function useFilteredGuard (guard, fn) {
       : null
 
   if (depth != null) {
-    const removeGuard = router.beforeEach((to, from, next) => {
-      return fn(to, from, depth) ? guard(to, from, next) : next()
-    })
+    const registerGuard = () => {
```
posva: Can you refactor this into an external function that takes what it needs? I think router, guard and fn?

src/composables/guards.js
```diff
@@ -56,11 +56,23 @@ function useFilteredGuard (guard, fn) {
...
+    let removeGuard = registerGuard()
     onUnmounted(removeGuard)
+
+    onActivated(() => {
+      removeGuard = removeGuard || registerGuard()
+    })
+    onDeactivated(() => {
+      removeGuard()
+      removeGuard = null // reset removeGuard
+    })
```
posva: Couldn't this just be
```suggestion
    onDeactivated(removeGuard)
```

posva (changes requested):
Thanks!

src/composables/guards.js
wynn719: Do you mean this?

```javascript
function registerGuard(router, guard, fn, depth) {
  return router.beforeEach((to, from, next) => {
    return fn(to, from, depth) ? guard(to, from, next) : next()
  })
}

function useFilteredGuard (guard, fn) {
  ...
  let removeGuard = registerGuard(router, guard, fn, depth)
  ...
}
```

src/composables/guards.js
wynn719: I need to reset `removeGuard` to `null` to ensure that `onActivated` will re-execute `registerGuard`, and `removeGuard` will not disappear after being called.

posva: Ah right, I thought that could be changed too but then I thought `onActivated()` could be called while a guard is active, creating multiple active guards too.

src/composables/guards.js
wynn719: > Can you refactor this into an external function that takes what it needs? I think router, guard and fn?

After changing the code, I resolved this conversation. If you have other suggestions, please feel free to let me know. This is my first time trying to solve the problem of an open source project. There may be some things that are not done well enough. 😁

src/composables/guards.js
```diff
@@ -34,6 +34,12 @@ export function onBeforeRouteLeave (guard) {
   return useFilteredGuard(guard, isLeaveNavigation)
 }
 
+function registerGuard (router, guard, fn, depth) {
+  return router.beforeEach((to, from, next) => {
+    return fn(to, from, depth) ? guard(to, from, next) : next()
+  })
+}
```
posva:
```suggestion
  return router.beforeEach((to, from, next) =>
    fn(to, from, depth) ? guard(to, from, next) : next()
  )
}
```

src/composables/guards.js
```diff
@@ -56,11 +62,17 @@ function useFilteredGuard (guard, fn) {
...
-      return fn(to, from, depth) ? guard(to, from, next) : next()
+    let removeGuard = registerGuard(router, guard, fn, depth)
+    onUnmounted(removeGuard)
+
+    onActivated(() => {
+      removeGuard = removeGuard || registerGuard(router, guard, fn, depth)
+    })
+    onDeactivated(() => {
+      removeGuard()
+      removeGuard = null // reset removeGuard
```
posva:
```suggestion
      removeGuard = null
```

posva (approved):
Thanks!

PR #3901, opened 2024-08-25
correct typedef of Route

### LINKED ISSUE #3566: The type definition of Route seems to be partially wrong
### Version
3.5.1

### Reproduction link
[https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L203](https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L203)

### Steps to reproduce
None, just the source code seems to be wrong.

### What is expected?
The type definition of Route#query should be

```
query: Dictionary<string | null | (string | null)[]>
```

The current implementation cannot express this type of query

```
/path?foo#bar
```

as it doesn't have any value and it is not converted to an array.

https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L203

The type definition of Route#query is

```
```

aryanbhosale:
Fixes #3566 

updated typedef of interface Route from `query: Dictionary<string | (string | null)[]>` to `query: Dictionary<string | null | (string | null)[]>`

### DIFF
--- a/types/router.d.ts
+++ b/types/router.d.ts
@@ -411,7 +411,7 @@ export interface Route {
   path: string
   name?: string | null
   hash: string
-  query: Dictionary<string | (string | null)[]>
+  query: Dictionary<string | null | (string | null)[]>
   params: Dictionary<string>
   fullPath: string
   matched: RouteRecord[]
--- a/types/test/route-query.spec.ts
+++ b/types/test/route-query.spec.ts
@@ -1,0 +1,103 @@
+import VueRouter from '../index'
+import { Route } from '../index'
+
+const component = { template: '<div>test</div>' }
+
+const router = new VueRouter({
+  routes: [
+    { path: '/:id', component }
+  ]
+})
+
+describe('Route query types', () => {
+  it('should handle string query parameter', () => {
+    const route: Route = {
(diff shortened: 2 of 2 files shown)


posva (changes requested):
Can you add a test?

aryanbhosale:
> Can you add a test?

Which files exactly? im kinda new to this so just the files that need to be covered would be great

posva:
You can add it here: https://github.com/vuejs/vue-router/tree/dev/types/test

aryanbhosale:
> You can add it here: https://github.com/vuejs/vue-router/tree/dev/types/test

Thank you! Will have a look

aryanbhosale:
> You can add it here: https://github.com/vuejs/vue-router/tree/dev/types/test

is my latest commit what you were looking for in the test?

posva:
Thanks! That should be the test. I will take a look on my next pass over vue router 3.

yash-dayama:
I have updated the code after reading all the comments can you please check this @posva

posva:
This is the old non maintained version of Vue router. [I only do paid reviews and release for companies](https://cal.com/posva/consultancy-vue-2?layout=mobile&month=2025-08)

PR #3931, opened 2025-07-31
Fix/route query type 3566

### LINKED ISSUE #3566: The type definition of Route seems to be partially wrong

yash-dayama:
Fixes #3566

Updated typedef of interface Route from:
to:
query: Dictionary<string | null | (string | null)[]>

Changes made:
1. Updated TypeScript type definition to properly handle null values in query parameters
2. Updated Flow type definitions to match the TypeScript changes
3. Added dedicated type tests in types/test/query-types.test.ts that verify:
   - String values (?foo=bar)
   - Null values (?foo)
   - Array values (?foo=bar&foo=baz)
   - Mixed arrays with both strings and nulls

This change ensures proper typing for query parameters that don't have any value, while maintaining backward compatibility.

### DIFF
--- a/flow/declarations.js
+++ b/flow/declarations.js
@@ -1,3 +1,5 @@
+/* @flow */
+
 declare var document: Document;
 
 declare class RouteRegExp extends RegExp {
@@ -18,6 +20,8 @@ declare module 'path-to-regexp' {
 }
 
 declare type Dictionary<T> = { [key: string]: T }
+declare type QueryValue = string | null | Array<string | null>
+declare type QueryDictionary = { [key: string]: QueryValue }
 
 declare type NavigationGuard = (
   to: Route,
@@ -84,7 +88,7 @@ declare type Location = {
   name?: string;
   path?: string;
   hash?: string;
-  query?: Dictionary<string | null | Array<string | null>>;
+  query?: QueryDictionary;
   params?: Dictionary<string>;
   append?: boolean;
   replace?: boolean;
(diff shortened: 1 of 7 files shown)


yash-dayama:
After 3 retries here is what I have done @posva can you please have a look

```
Fixed the Route query type definition issue with the following changes:

1. Type Definitions:
   - Updated Route interface query type to properly handle null values:
     ```ts
     query: Dictionary<string | null | (string | null)[]>
     ```
   - Added QueryValue and QueryDictionary types in Flow for better type organization
   - Added dedicated type tests in types/test/query-types.test.ts

2. Runtime Changes:
   - Improved isObjectEqual to properly handle array comparisons and string conversions
   - Updated resolveQuery and stringifyQuery to use QueryDictionary type
   - Fixed query type handling in normalizeLocation

These changes ensure proper typing for all query parameter scenarios:
- String values (?foo=bar)
- Null values (?foo)
- Array values (?foo=bar&foo=baz)
- Mixed arrays with both strings and nulls

All tests are passing and Flow type checking is successful. This maintains backward compatibility while fixing the type definition issues reported in #3566.
```

posva:
Thanks but same as in the other PR 😅

